Issue #51, opened 2018-01-03
Mutations over Websocket


wakemaster39:
According to the Apollo web socket transport implementation, a websocket is suppose to be able to handle Queries and Mutation in addition to Subscriptions. I was playing around with attempting to send a mutation over the websocket channel but the response back from this is always a GQL_ERROR package with the payload of both the data and the errors being null, which makes it very difficult to track down where the error actually occurs.

pekkah:
At the moment queries and mutations over websockets are not implemented. Do you want to make a PR for it?

wakemaster39:
@pekkah I was trying to look at the codebase and figured out where to tie it in and haven't quiet figured out how would be best. I have managed to learn most of how the websocket package functions, but how a document/subscription executor functions is still a black box for me. 

Do you have any suggestions on where might be a good place to start tying that feature set in and I can take a look from there? My initial thought was it should be handled by the GraphQL subscription executor and it should just immediately unsubscribe after result is send.

PR #52, opened 2018-01-04
Websocket Handle other method from subscriptions

wakemaster39:
This is a quick implementation of how to handle allow the web socket protocol to handle not only subscriptions but queries and mutations as well. This is related to issue #51.

The main issue that I have with the implementation is that I need to parse the document to figure out what the operation type is. Any suggestions on how to work around this I would look into changing it. 

If this implementation methodology is acceptable, I will look at cleaning it up and adding in the necessary tests.

### DIFF
--- a/src/WebSockets/Messages/OperationMessage.cs
+++ b/src/WebSockets/Messages/OperationMessage.cs
@@ -12,6 +12,6 @@ namespace GraphQL.Server.Transports.WebSockets.Messages
         public string Id { get; set; }
 
         [JsonProperty("payload")]
-        public JObject Payload { get; set; }
+        public dynamic Payload { get; set; }
     }
 }
--- a/src/WebSockets/SubscriptionDeterminator.cs
+++ b/src/WebSockets/SubscriptionDeterminator.cs
@@ -1,19 +1,11 @@
-using System;
-using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 using GraphQL.Execution;
-using GraphQL.Instrumentation;
 using GraphQL.Language.AST;
-using GraphQL.Types;
-using GraphQL.Validation;
-using GraphQL.Validation.Complexity;
-using GraphQLParser.AST;
-using OperationType = GraphQL.Language.AST.OperationType;
+using GraphQL.Server.Transports.WebSockets.Abstractions;
 
 namespace GraphQL.Server.Transports.WebSockets
 {
-    public class SubscriptionDeterminator
+    public class SubscriptionDeterminator : ISubscriptionDeterminator
     {
         private readonly IDocumentBuilder _documentBuilder;
 
@@ -28,28 +20,15 @@ namespace GraphQL.Server.Transports.WebSockets
 
         public bool IsSubscription(ExecutionOptions config)
         {
-
-            config.Schema.FieldNameConverter = config.FieldNameConverter;
-
             ValidateOptions(config);
 
-            if (!config.Schema.Initialized)
-            {
-                config.Schema.Initialize();
-            }
+            config.Document = config.Document ?? _documentBuilder.Build(config.Query);
 
-            var document = config.Document ?? _documentBuilder.Build(config.Query);
-
-            return GetOperation(config.OperationName, document).OperationType == OperationType.Subscription;
+            return GetOperation(config.OperationName, config.Document).OperationType == OperationType.Subscription;
         }
 
         private void ValidateOptions(ExecutionOptions options)
         {
-            if (options.Schema == null)
-            {
-                throw new ExecutionError("A schema is required.");
-            }
-
             if (string.IsNullOrWhiteSpace(options.Query))
             {
                 throw new ExecutionError("A query is required.");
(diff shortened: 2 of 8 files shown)


joemcbride:
Thanks for the PR!  A few comments.

It is possible to build the `Document` and add that to the `ExecutionOptions`.  If you do that then the `DocumentExecutor` won't try to build it again.  So I would suggest to build the `Document`, check the operation, and forward the `Document` along.  As it is setup currently, the `Document` will get built twice.

Perhaps that `GetOperation` method can be moved to the `Document` class so it could be reused here.

Also, it seems like the `Payload` property on `OperationMessage` will need to be changed to something else (or some other pattern), so that we're not writing JSON to a string and then parsing it just to write it again.

src/WebSockets/SubscriptionDeterminator.cs
```diff
@@ -0,0 +1,68 @@
...
+            config.Schema.FieldNameConverter = config.FieldNameConverter;
+
+            ValidateOptions(config);
+
+            if (!config.Schema.Initialized)
+            {
+                config.Schema.Initialize();
+            }
+
+            var document = config.Document ?? _documentBuilder.Build(config.Query);
```
joemcbride: Just a note here, you don't need the `Schema` nor the `ExecutionOptions` to build the `Document`.  All you need is the `Query`.

wakemaster39:
@joemcbride I refactored the code and got the tests passing since the approach was ok. I'll put in a PR on the main codebase for the GetOperation refactoring but I left that as is for now. 

For refactoring the Payload, I used the dynamic approach that the GraphQL.Client library is using. I will also attempt to clean up my WebSocket Client and pass it along to there so there is a complete working implementation.

pekkah:
Sorry for the delay. I'll try reviewing this today.

pekkah:
After looking at this I think the best place would be to do the forking of execution is inside the subscription executer.

pekkah:
Or not. The reference JS implementation has these as separate so I suppose it would be best to keep them separate here as well.

src/WebSockets/Messages/OperationMessage.cs
```diff
@@ -12,6 +12,6 @@ public class OperationMessage
         public string Id { get; set; }
 
         [JsonProperty("payload")]
-        public JObject Payload { get; set; }
+        public dynamic Payload { get; set; }
```
pekkah: Why dynamic?

src/WebSockets/SubscriptionProtocolHandler.cs
```diff
@@ -79,11 +81,47 @@ public Task HandleConnectionClosed(OperationMessageContext context)
...
+                Inputs = query.GetInputs(),
+                Query = query.Query,
+                ExposeExceptions = options?.ExposeExceptions ?? false,
+                ValidationRules = options?.ValidationRules,
+                UserContext = options?.BuildUserContext?.Invoke(context)
+            };
+
+            var isSubscription = _determinator.IsSubscription(exOptions);
+
+            if (isSubscription)
```
pekkah: I would probably implement this so that even when executing mutation or query I would still handle it as an subscription and just wrap the result into `ReplaySubject` and complete it after the first (only) value. This would keep the pipeline clean and allow other future enhancements without burden of forked logic.

src/WebSockets/Messages/OperationMessage.cs
wakemaster39: I chose dynamic because it really doesn't matter what the payload is, it should be serialised and sent on. It could be saved as an object a well I suppose.

src/WebSockets/SubscriptionProtocolHandler.cs
wakemaster39: I disagree that this should be handled as a subscription with only one value. I feel that the name `SubscriptionProtocolHandler.cs` is actually wrong and should be called something like `WebSocketProtocolHandler.cs`.

My reasoning here is that currently this file is handling messages which happen to be coming from a websocket. The messages just happen to be a subscription, mutation or a query. Someone could in theory go and implement a `HttpLongPoolingHandler.cs` which would also need to handle subscriptions, mutations and queries.

At a framework level, subscriptions were different enough to require their own handler independent from queries/mutations. So I feel that this protocol handler should call the appropriate handler and return responses appropriately. 

I am happy to look at how to modify this code, but unless the need for a separate handler for subscriptions is removed I feel something along these lines is appropriate.

pekkah: Maybe I've been been too deep in these lately but WebSocket protocol to me is the actual protocol that handles the transfer between the browser and server and does not have anything to do with GraphQL. 

I chose the name SubscriptionProtocolHandler because ApolloSubscriptionProtocolHandler was getting bit too long. That's what this is. It handles subscriptions according to how Apollo handles them. Being compatible with Apollo client.

wakemaster39: The reasoning for my agrument actually comes from that project as well. Following the protocol information provided here: https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md . 

Please correct me if I am miss understanding this in any way, but this protocol seems to define a mechanism of sending graphql requests over a websocket channel and doesn't specifically require the payload to be a subscription. When I read the Message flow section, specifically the Connected Phase, I tried to emulate that in the code above. 

I think that the subscription-transport-ws has also grown past the subscription requirement is really is a transport-ws library now.

PR #53, opened 2018-01-05
Deserialize directly from stream

joemcbride:
* Use provided .ToInputs() extension for JObject

### DIFF
--- a/src/AspNetCore/Common/GraphQLQuery.cs
+++ b/src/AspNetCore/Common/GraphQLQuery.cs
@@ -22,12 +22,7 @@ namespace GraphQL.Server.Transports.AspNetCore.Common
 
         public static Inputs GetInputs(JObject variables)
         {
-            if (variables == null)
-                return null;
-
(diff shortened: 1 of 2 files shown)


src/AspNetCore/Common/GraphQLQuery.cs
```diff
@@ -22,12 +22,7 @@ public Inputs GetInputs()
 
         public static Inputs GetInputs(JObject variables)
         {
-            if (variables == null)
-                return null;
-
-            var inputs = variables.GetValue() as Dictionary<string, object>;
-
-            return new Inputs(inputs);
+            return variables.ToInputs();
```
joemcbride: Should this perhaps be removed entirely?

pekkah: Yes. Just using one would be less confusing.

pekkah: Lets do a another PR for this.

Issue #56, opened 2018-01-22
Question: When working with Scoped services


samyhrer:
This is not an issue, just a general question I hope someone can provide some insight for.

We are currently trying to create a Graphql endpoint with an Entity Framework based service layer. Furthermore we try to utilize Dependency Injection in order to inject the dbContext in our Mutation-, Query- and Subscription-implementations. Using Dependency Injection has although proven difficult as the injected dbContext is Scoped by nature. This seems to conflict with how this library require(?) that everything is registered as a Singleton. That said, I believe I've solved it for the HTTP-middleware by creating a custom HTTP-Middleware where the Schema is resolved in the Invoke-method instead of the constructor. The WebSocket-Middleware is although a bit more complicated, and we've not found any elegant solutions to this. 

So any advice on using Subscriptions together with a Schema that relies on scoped dependencies will be much appreciated.

ntaranov:
Hi, I was migrating from using a controller action to middleware and encountered the same issue, here are my 2 cents. 

Controllers are created on per request basis while middleware instances are created as singletons by default. 

I'm using standard .net core ID and problem seems to be as follows. 
1. You inject something referencing EF DB context somewhere fields' resolver functions. 
For me it was DbQuery which referenced scoped DB context. 
2. Middleware singleton is created, all the types needed for this are resolved and all objects needed for the creation of the sigleton are instantiated. 
3. Middleware references schema and sets it in constructor forever. 
4. Schema references Query and Mutation, they are resolved once and forever when schema is created. 
5. As fields are added in Graph types constructor, this effectively means constructor injection for fields. 
6. As a result no matter how you register your dependecies, DB context is going to be resolved and instantiated only once, when the middleware singleton is created. 
7. As a result, all requests share 1 instance of DB context on server, and as soon as 2 HTTP requests arrive with little time in between, and exception is thrown. 

Implementing you own middleware, using ISchema method injection in Invoke is going to result in Shema being resolved on every request. This will result in a bit worse performance, but the main problem is that the solution might not guarantee DB context concurrency safety. 

I appears that fields [are opinionated](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Execution/DocumentExecuter.cs#L328) about if their resolve functions are executed in a single thread or not.

[...]

PR #61, opened 2018-02-02
added IDocumentExecutionListener to IDocumentExecuter from IOC

klowdo:
Because Dateloaders in grapgl-dotnet i we need to add IDocumentExecutionListener to the execution listners to handle the scope of the request.

Then it is just to add IDocumentExecutionListener implementations to the service collection

### DIFF
--- a/src/AspNetCore/GraphQLHttpMiddleware.cs
+++ b/src/AspNetCore/GraphQLHttpMiddleware.cs
@@ -10,6 +10,7 @@ using GraphQL.Server.Transports.AspNetCore.Common;
 using GraphQL.Types;
 using GraphQL.Validation;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
 using Newtonsoft.Json;
 
@@ -22,25 +23,22 @@ namespace GraphQL.Server.Transports.AspNetCore
         private readonly IDocumentExecuter _executer;
         private readonly IDocumentWriter _writer;
         private readonly TSchema _schema;
-        private readonly IEnumerable<IDocumentExecutionListener> _documentListeners;
 
         public GraphQLHttpMiddleware(
             RequestDelegate next,
             IOptions<GraphQLHttpOptions> options,
             IDocumentExecuter executer,
             IDocumentWriter writer,
-            TSchema schema,
-            IEnumerable<IDocumentExecutionListener> documentListeners)
+            TSchema schema)
         {
             _next = next;
             _options = options.Value;
             _executer = executer;
             _writer = writer;
             _schema = schema;
-            _documentListeners = documentListeners;
         }
 
-        public async Task Invoke(HttpContext context)
+        public async Task Invoke(HttpContext context, IEnumerable<IDocumentExecutionListener> documentExecutionListeners)
         {
             if (!IsGraphQlRequest(context))
             {
@@ -48,7 +46,7 @@ namespace GraphQL.Server.Transports.AspNetCore
                 return;
             }
 
-            await ExecuteAsync(context, _schema);
+            await ExecuteAsync(context, _schema, documentExecutionListeners);
         }
 
         private bool IsGraphQlRequest(HttpContext context)
@@ -56,7 +54,7 @@ namespace GraphQL.Server.Transports.AspNetCore
             return context.Request.Path.StartsWithSegments(_options.Path);
         }
 
-        private async Task ExecuteAsync(HttpContext context, ISchema schema)
+        private async Task ExecuteAsync(HttpContext context, ISchema schema, IEnumerable<IDocumentExecutionListener> documentExecutionListeners)
         {
             var request = Deserialize<GraphQLQuery>(context.Request.Body);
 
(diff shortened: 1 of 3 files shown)


src/AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -20,19 +22,22 @@ public class GraphQLHttpMiddleware<TSchema> where TSchema : ISchema
...
+        private readonly IEnumerable<IDocumentExecutionListener> _documentListners;
 
         public GraphQLHttpMiddleware(
             RequestDelegate next,
             IOptions<GraphQLHttpOptions> options,
             IDocumentExecuter executer,
             IDocumentWriter writer,
-            TSchema schema)
+            TSchema schema,
+            IEnumerable<IDocumentExecutionListener> documentListners)
```
pekkah: Listeners

src/AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -64,6 +69,9 @@ private bool IsGraphQlRequest(HttpContext context)
                 _.UserContext = _options.BuildUserContext?.Invoke(context);
                 _.ExposeExceptions = _options.ExposeExceptions;
                 _.ValidationRules = _options.ValidationRules.Concat(DocumentValidator.CoreRules()).ToList();
+                _documentListners
+                    .ToList()
```
pekkah: Why create another list

klowdo: When you take toList() it instaciates all the registred types and makes a list of it.  
It may be same as just enumerating it. 
do you sugest this approach instead?
```
foreach(var listner in _documentListners){
    _.Listeners.Add(listner)
}

```

src/AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -64,6 +69,9 @@ private bool IsGraphQlRequest(HttpContext context)
                 _.UserContext = _options.BuildUserContext?.Invoke(context);
                 _.ExposeExceptions = _options.ExposeExceptions;
                 _.ValidationRules = _options.ValidationRules.Concat(DocumentValidator.CoreRules()).ToList();
+                _documentListeners
+                    .ToList()
```
pekkah: This will allocate a new list. Less allocations is better.

klowdo: okey! I fix!

src/AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -20,19 +22,22 @@ public class GraphQLHttpMiddleware<TSchema> where TSchema : ISchema
...
+        private readonly IEnumerable<IDocumentExecutionListener> _documentListeners;
 
         public GraphQLHttpMiddleware(
             RequestDelegate next,
             IOptions<GraphQLHttpOptions> options,
             IDocumentExecuter executer,
             IDocumentWriter writer,
-            TSchema schema)
+            TSchema schema,
+            IEnumerable<IDocumentExecutionListener> documentListeners)
```
sahb1239: `IEnumerable<IDocumentExecutionListener> documentListeners)` should be moved to the method `Invoke` since the middleware is only instantiated one time at app startup.

For example DataLoaderDocumentListener is registrered as Transient in the [DataLoader tests](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL.DataLoader.Tests/QueryTestBase.cs#L42). In this pull request the actual lifetime would instead be singleton.

See [Per-request dependencies](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware/?tabs=aspnetcore2x)

klowdo: Okey i will look into this. thanks!

klowdo: @pekkah @sahb1239 i have a heard time injecting it correct way to SubscriptionProtocolHandler.
i'm thinking of a factory or just inject the IServiceProvider to request service at execution.

sahb1239: I think IServiceProvider would be sufficient

sahb1239: Issue #56 is releated

pekkah:
Is this ready`?

klowdo:
I think so? I cloud try som more with Dataloaders if you want

jphenow:
This looks great - I need listeners supported to use this lib (which I'd love to, it's otherwise awesome).

After looking back at this code and considering how #64 was added, I think I personally prefer that direction as I find it nice to not _have_ to conform to some sort of IOC. Furthermore, it deviates from the resolver interface from GraphQL. While this specific library is aimed at aspnet core it seems like we're oddly leaning into a middle-ground for resolving dependencies that's maybe nice to avoid entirely.

Anyways, I'm probably thinking too hard - if I can have listeners, I'll be overjoyed regardless of implementation at this point.

pekkah:
As  3.0 supports configuring execution options using Microsoft.Extensions.Options this is now supported.

PR #64, opened 2018-03-06
Add Socket and HTTP listener options

jphenow:
Thoughts? Happy to alter and add tests, etc.


jphenow:
To clarify a little,  I'm moving from my hand-written Middleware (which is very similar to this) which uses listeners to initialize DataLoaderContexts

joemcbride:
I think this is essentially a duplicate of #61

jphenow:
Ha! So it is. I didn't look close enough then. My bad! I'll close this. Thanks for glancing

jphenow:
This is what I get for third-beer-in coding

PR #69, opened 2018-03-09
Target .NET Standard 2.0

jinhong-:
I changed the target to .NET Standard 2.0 for AspNetCore and WebSockets project


pekkah:
3.0 targets netstandard2.0

Issue #65, opened 2018-03-07
Make GraphQLHttpMiddleware only apply to POST requests.


BenjaBobs:
Get requests cause a null-ref exception when using GraphQL Http middleware.
I think adding a check for the request method would be enough to fix it.

https://github.com/graphql-dotnet/server/blob/master/src/AspNetCore/GraphQLHttpMiddleware.cs#L50-L53

Good job on the lib btw. Just started using it, very easy and intuitive.

PR #84, opened 2018-03-21
Fixed HttpMiddleware catching non-POST requests (#65)

BenjaBobs:


src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -49,13 +48,13 @@ public class GraphQLHttpMiddleware<TSchema> where TSchema : ISchema
 
         private bool IsGraphQlRequest(HttpContext context)
         {
-            return context.Request.Path.StartsWithSegments(_options.Path);
+            return context.Request.Method.Equals("POST") && context.Request.Path.StartsWithSegments(_options.Path);
```
pekkah: Would this be better https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.httpmethods.ispost?view=aspnetcore-2.0 ?

BenjaBobs: Good call!

Issue #94, opened 2018-03-24
Implement HTTP middleware to official recommendation


benmccallum:
Currently looks like it just detects the right path (e.g. /graphql by default) and then assumes everything is in the request body. It should handle requests as recommended here: http://graphql.org/learn/serving-over-http/

I've done this in a my own implementation with a proper Controller, but will do a PR for this project shortly.

benmccallum:
Currently working on a PR.

benmccallum:
Almost finished the [change](https://github.com/benmccallum/server/commit/f391663f59db5210dd2770a3326325863ab6c5ff). I'm just having a problem with reading the context.Request.Body as it fails straight after it's read. Got a StackOverflow [question](https://stackoverflow.com/questions/49490350/exception-after-reading-context-request-body-in-asp-net-core-middleware) open on it as I'm not 100% sure what's happening.

benmccallum:
Completed as part of PR https://github.com/graphql-dotnet/server/pull/100

PR #97, opened 2018-03-26
Implement HTTP middleware to official recommendation #94

benmccallum:
PR for issue logged at: https://github.com/graphql-dotnet/server/issues/94

### DIFF
--- a/src/AspNetCore/GraphQLHttpMiddleware.cs
+++ b/src/AspNetCore/GraphQLHttpMiddleware.cs
@@ -57,16 +57,13 @@ namespace GraphQL.Server.Transports.AspNetCore
         private async Task ExecuteAsync(HttpContext context, ISchema schema)
         {
             // Handle requests as per recommendation at http://graphql.org/learn/serving-over-http/
-            GraphQLQuery request = null;
+            var request = new GraphQLQuery();
             if (context.Request.Method.Equals("GET", StringComparison.OrdinalIgnoreCase))
             {
-                var q = context.Request.Query;
-                request = new GraphQLQuery()
-                {
-                    Query = q.TryGetValue("query", out StringValues queryValues) ? queryValues[0] : null,
-                    Variables = q.TryGetValue("variables", out StringValues variablesValues) ? JObject.Parse(variablesValues[0]) : null,
-                    OperationName = q.TryGetValue("operationName", out StringValues operationNameValues) ? operationNameValues[0] : null
-                };
+                var qs = context.Request.Query;
+                request.Query = qs.TryGetValue("query", out StringValues queryValues) ? queryValues[0] : null;
+                request.Variables = qs.TryGetValue("variables", out StringValues variablesValues) ? JObject.Parse(variablesValues[0]) : null;
+                request.OperationName = qs.TryGetValue("operationName", out StringValues operationNameValues) ? operationNameValues[0] : null;
             }
             else if (context.Request.Method.Equals("POST", StringComparison.OrdinalIgnoreCase))
             {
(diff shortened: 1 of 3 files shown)


src/AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -54,7 +56,27 @@ private bool IsGraphQlRequest(HttpContext context)
 
         private async Task ExecuteAsync(HttpContext context, ISchema schema)
         {
-            var request = Deserialize<GraphQLQuery>(context.Request.Body);
+            // Handle requests as per recommendation at http://graphql.org/learn/serving-over-http/
+            var request = new GraphQLQuery();
+            if (context.Request.Method.Equals("GET", StringComparison.OrdinalIgnoreCase))
```
pekkah: https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.httpmethods?view=aspnetcore-2.0

src/AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -54,7 +56,27 @@ private bool IsGraphQlRequest(HttpContext context)
...
+            {
+                var qs = context.Request.Query;
+                request.Query = qs.TryGetValue("query", out StringValues queryValues) ? queryValues[0] : null;
+                request.Variables = qs.TryGetValue("variables", out StringValues variablesValues) ? JObject.Parse(variablesValues[0]) : null;
+                request.OperationName = qs.TryGetValue("operationName", out StringValues operationNameValues) ? operationNameValues[0] : null;
+            }
+            else if (context.Request.Method.Equals("POST", StringComparison.OrdinalIgnoreCase))
```
pekkah: See above

src/AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -54,7 +56,27 @@ private bool IsGraphQlRequest(HttpContext context)
...
+            {
+                switch (context.Request.ContentType)
```
pekkah: Does this handle encodings as well?

benmccallum: Not exactly sure what you mean by encodings. Could you give an example?

benmccallum:
Hey @pekkah , keen to get this through. See my question above.

pekkah:
If the content type includes the charset=UTF-8 will this still work? I haven't checked if the ASP.NET Core parses the contenttype with charset or not..

benmccallum:
Oh I see what you mean, I'll have a look now.

benmccallum:
You're right. The charset and whatever else is passed on the end of the ContentType header does come through and will break my comparison. Good pickup. Code reviews FTW.

Did a bit of [digging](https://github.com/aspnet/Mvc/blob/760c8f38678118734399c58c2dac981ea6e47046/test/WebSites/BasicWebSite/Formatters/CustomFormatter.cs) and found `System.Net.Http.Headers.MediaTypeHeaderValue` which that can normalize it.

Also added a missing scenario, handling the POST request with ?query={...} supplied. In this case it should be handled as the GET request is.

Few concerns I have now are:

1. POST with broken `ContentType: application/json; charset=myNonExistantCharset;` will skip the parsing, go to the DocumentExecutor with empty GraphQLRequest and come back with something like "query must be supplied". Should we short-circuit and return a BadRequest with relevant message?
2. POST with empty ContentType or anything we don't handle, same thing? Short-circuit with a BadRequest?

pekkah:
For 1. and 2. Bad request sounds ok.

benmccallum:
@pekkah, done mate. Over to you.

pekkah:
Some of those nested if,switches could use some cleanup but that can be a chore before stable release.

pekkah:
Just noticed that you created this against the master branch. Could you change it to develop as that's where the next version is brewing.

benmccallum:
Sure, no problem.

PR #101, opened 2018-03-29
Updating docs and standardising GraphiQL's api url

benmccallum:


benmccallum:
Second commit is a "breaking change" I guess, but needed to be done for put-of-the-box consistency

PR #112, opened 2018-05-16
Update README.md

benmccallum:
ExecutionOptions is no longer generic in latest version of graphql-dotnet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ public void ConfigureServices(IServiceCollection services)
     services.AddGraphQLHttp();
     
     // setup execution options for ChatSchema
-    services.Configure<ExecutionOptions<ChatSchema>>(options =>
+    services.Configure<ExecutionOptions>(options =>
             {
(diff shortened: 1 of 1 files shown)


pekkah:
I'm sure it is https://github.com/graphql-dotnet/server/blob/develop/src/Transports.Subscriptions.WebSockets/ExecutionOptions.cs

benmccallum:
This isn't of type `WebSockets.ExecutionOptions<T>` though, it's `GraphQL.ExecutionOptions`. https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Execution/ExecutionOptions.cs

benmccallum:
?

joemcbride:
@benmccallum Looks like the syntax + structure overall was changed, so this change no longer applies.

benmccallum:
Yep, was on mobile so couldn't take a good look myself. An explanation is always appreciated.

Issue #111, opened 2018-05-14
The remote party closed the WebSocket connection without completing the close handshake. Unexpected end of request content.


BenjaBobs:
I'm getting this exception on `v3.1.0-alpha0020`

```
System.AggregateException:
   at GraphQL.Server.Transports.Subscriptions.Abstractions.SubscriptionServer+<OnConnect>d__18.MoveNext (GraphQL.Server.Transports.Subscriptions.Abstractions, Version=3.1.0.0, Culture=neutral, PublicKeyToken=null)
   at GraphQL.Server.Transports.WebSockets.WebSocketConnection+<Connect>d__9.MoveNext (GraphQL.Server.Transports.WebSockets, Version=3.1.0.0, Culture=neutral, PublicKeyToken=null)
   at GraphQL.Server.Transports.WebSockets.GraphQLWebSocketsMiddleware`1+<ExecuteAsync>d__9.MoveNext (GraphQL.Server.Transports.WebSockets, Version=3.1.0.0, Culture=neutral, PublicKeyToken=null)
   at GraphQL.Server.Transports.WebSockets.GraphQLWebSocketsMiddleware`1+<Invoke>d__7.MoveNext (GraphQL.Server.Transports.WebSockets, Version=3.1.0.0, Culture=neutral, PublicKeyToken=null)
   at Microsoft.AspNetCore.Builder.RouterMiddleware+<Invoke>d__4.MoveNext (Microsoft.AspNetCore.Routing, Version=2.0.2.0, Culture=neutral, PublicKeyToken=adb9793829ddae60)
   at Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware+<Invoke>d__7.MoveNext (Microsoft.AspNetCore.Diagnostics, Version=2.0.2.0, Culture=neutral, PublicKeyToken=adb9793829ddae60)
   at Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware+<Invoke>d__7.MoveNext (Microsoft.AspNetCore.Diagnostics, Version=2.0.2.0, Culture=neutral, PublicKeyToken=adb9793829ddae60)
   at Microsoft.AspNetCore.Cors.Infrastructure.CorsMiddleware+<Invoke>d__7.MoveNext (Microsoft.AspNetCore.Cors, Version=2.0.2.0, Culture=neutral, PublicKeyToken=adb9793829ddae60)
[...]

Issue #110, opened 2018-05-14
Websocket error -4077 ECONNRESET connection reset by peer.


BenjaBobs:
I'm getting this exception using `v3.1.0-alpha0020`.

```
One or more errors occurred. 
(One or more errors occurred. (The remote party closed the WebSocket connection without completing the close handshake.)) 
One or more errors occurred. (The remote party closed the WebSocket connection without completing the close handshake.) 
The remote party closed the WebSocket connection without completing the close handshake. 
Error -4077 ECONNRESET connection reset by peer Error -4077 ECONNRESET connection reset by peer.

System.AggregateException:
   at GraphQL.Server.Transports.Subscriptions.Abstractions.SubscriptionServer+<OnConnect>d__18.MoveNext (GraphQL.Server.Transports.Subscriptions.Abstractions, Version=3.1.0.0, Culture=neutral, PublicKeyToken=null)
   at GraphQL.Server.Transports.WebSockets.WebSocketConnection+<Connect>d__9.MoveNext (GraphQL.Server.Transports.WebSockets, Version=3.1.0.0, Culture=neutral, PublicKeyToken=null)
   at GraphQL.Server.Transports.WebSockets.GraphQLWebSocketsMiddleware`1+<ExecuteAsync>d__9.MoveNext (GraphQL.Server.Transports.WebSockets, Version=3.1.0.0, Culture=neutral, PublicKeyToken=null)
   at GraphQL.Server.Transports.WebSockets.GraphQLWebSocketsMiddleware`1+<Invoke>d__7.MoveNext (GraphQL.Server.Transports.WebSockets, Version=3.1.0.0, Culture=neutral, PublicKeyToken=null)
   ...
Inner exception System.AggregateException handled at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw:
System.Net.WebSockets.WebSocketException:
   at System.Net.WebSockets.ManagedWebSocket+<ReceiveAsyncPrivate>d__61.MoveNext (Microsoft.AspNetCore.WebSockets, Version=2.0.2.0, Culture=neutral, PublicKeyToken=adb9793829ddae60)
   at GraphQL.Server.Transports.WebSockets.WebSocketReaderPipeline+<ReadMessageAsync>d__11.MoveNext (GraphQL.Server.Transports.WebSockets, Version=3.1.0.0, Culture=neutral, PublicKeyToken=null)
[...]

PR #114, opened 2018-05-16
Fix infinite loop when client disconnects without notifying server.

BenjaBobs:
This should fix both #110 and #111


samyhrer:
Will this fix be made available as part of an new alpha-release on nuget?

BenjaBobs:
I am currently using alpha0022 and it is included in that release.

samyhrer:
@BenjaBobs Thanks. Didn't catch that alpha0022 is available on MyGet and not NuGet :)

PR #115, opened 2018-05-17
Use MVC JsonSerializerSettings & Indent JSON in Development

RehanSaeed:
The `JsonSerializerSettings` is available in MVC via `IOptions<T>` that is very easy to configure via `.AddMvcCore(...).AddJsonOptions(...)`. So perhaps we should use it.

The second part of this PR is to indent the JSON when running in development mode for better readability.


RehanSaeed:
This PR also helps with https://github.com/graphql-dotnet/graphql-dotnet/issues/662 in that MVC sets `JsonSerializerSettings.DateFormatString` to the correct ISO8601 value with the offset by default.

RehanSaeed:
Moved to https://github.com/graphql-dotnet/server/pull/121

PR #123, opened 2018-06-12
Add Listeners to GraphQLHttpOptions

freakingawesome:
This allows Dataloader to be used with this library


johnrutherford:
I was planning on adding the DataLoader support, but haven't had the time to do it yet.

I had a different approach in mind. I think it would be best to use DI to get all registered `IDocumentExecutionListener`s. Something like this:

``` csharp
var listeners = context.RequestServices.GetService<IEnumerable<IDocumentExecutionListener>>();
```

And the necessary components would be registered inside `Startup.ConfigureServices()`. And there would ideally be an extension method to do that in one line.

joemcbride:
Agree with @johnrutherford, probably best to have those injection through DI.

fubar-coder:
The advantage of this PR is that you can have different listeners for different graphql endpoints.

freakingawesome:
I'm closing this in favor of the more thorough approach in #127.

PR #127, opened 2018-06-28
Simplify configuration of ExecutionOptions

freakingawesome:
This removes the duplication of fields between `ExecutionOptions` and `GraphQLHttpOptions`, and offers direct customization of `ExecutionOptions` fields by way of a delegate function.

Here is an example of Configuration. Note that we no longer need a custom `BuildUserContext` function on `GraphQLHttpOptions` because `ConfigureAsync` also passes the current `HttpContext`.

```c#
app.UseGraphQLHttp<ISchema>(new GraphQLHttpOptions
{
    ConfigureAsync = async (opts, ctx) =>
    {
        opts.UserContext = new MyUserContext
        {
            User = ctx.User
        };
        opts.EnableMetrics = true;
        opts.ExposeExceptions = env.IsDevelopment();
        await Task.CompletedTask;
    },
});
```

Validators can be injected via the `IServiceCollection`:

```c#
services.AddSingleton(DocumentValidator.CoreRules().Concat(new[] {
    new MyCustomValidationRule(),
}));
```

Likewise, `IDocumentExecutionListener` instances can be injected via `IServiceCollection`, allowing, for example, DataLoader injection:

```c#
services.AddSingleton<IDocumentExecutionListener, DataLoaderDocumentListener>()
```

### DIFF
--- a/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
+++ b/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
@@ -1,8 +1,10 @@
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Net;
 using System.Net.Http.Headers;
 using System.Threading.Tasks;
+using GraphQL.Execution;
 using GraphQL.Http;
 using GraphQL.Server.Transports.AspNetCore.Common;
 using GraphQL.Types;
@@ -26,6 +28,12 @@ namespace GraphQL.Server.Transports.AspNetCore
         private readonly IDocumentExecuter _executer;
         private readonly IDocumentWriter _writer;
         private readonly TSchema _schema;
+
+        public virtual ExecutionOptions CreateDefaultExecutionOptions() => new ExecutionOptions
+        {
+            EnableMetrics = true,
+            SetFieldMiddleware = true,
+        };
 
         public GraphQLHttpMiddleware(
             RequestDelegate next,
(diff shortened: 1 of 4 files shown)


pekkah:
Could you also update the readme

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -89,30 +97,38 @@ private bool IsGraphQLRequest(HttpContext context)
...
             var userContextBuilder = context.RequestServices.GetService<IUserContextBuilder>();
-            if (userContextBuilder != null)
+            if (userContextBuilder != null) {
+                opts.UserContext = await userContextBuilder.BuildUserContext(context);
+            }
+
+            if (opts.ValidationRules == null)
             {
-                userContext = await userContextBuilder.BuildUserContext(context);
+                opts.ValidationRules = context.RequestServices.GetService<IEnumerable<IValidationRule>>();
```
joemcbride: This is a change in behavior in that people will now be required to add all of the Core Rules to the DI Container.  Was that intentional?

freakingawesome: Fixed in 93b56cf

freakingawesome:
I just pushed up another commit which abstracts away some of the boilerplate for creating an `ExecutionOptions` instance. It now requires an `IExecutionOptionsFactory` on the middleware.

```c#
services.AddSingleton<IExecutionOptionsFactory, ExecutionOptionsFactory>();
```

This most recent push also addresses @joemcbride's concern about validation rules backwards-compatibility.

If there is quorum on these changes, I'll update the README with relevant information.

pekkah:
It's alpha so I have no problems with breaking things if it improves the project in longer run.

johnrutherford:
I just opened pull request #128 with a different approach to address the same key issues.

pekkah:
I think we should go with #128. It unifies the configuration for both http and websockets which should make it easier to get started.

PR #128, opened 2018-06-29
DataLoader support and refactor

johnrutherford:
I'm proposing some extensive changes here so I'm open to discussion.

This is what I set out to accomplish:
* Unify configuration and execution logic for all transports. I created a new Core project for common stuff. 
* All options configuration is done in ConfigureServices to more closely follow the pattern that MS has established.
* Execution listeners and validation rules are pulled in through DI to make it simple for them to be added.
* `IGraphQLBuilder` is used to register GraphQL-related services in a fluent way. 
* Middleware classes get dependencies with the expected lifetime.
* Middleware is added to the pipeline with a `PathString` so it is not created/executed at all if the path does not match. 

The only potential problem I see with this redesign is that it's not possible to have different listeners/validation rules for different endpoints or different schemas. I'm not sure if that may be a requirement for some users.


johnrutherford:
I just pushed a fix for the failing tests. There's still more I could do to clean up and simplify things, but I wanted to push what I had to get some feedback on this direction.

pekkah:
This is looking good. I like the unification. I don't know if anyone actually is hosting two scheman inside same application so I don't think it is that big of a issue.

caoquendo:
Hello! Sorry to bother. When should we expect this PR to be merged and a new version of the lib to be published? I guess this is a long awaited fix. Thanks!

RehanSaeed:
Overall, this PR looks good. I'm interested in configuring certain execution options from `appsettings.json` like this:

```
  "MainSchema": {
    "ComplexityConfiguration": {
      // The total maximum nesting across all queries in a request.
      "MaxDepth": 2,
      "MaxComplexity": 100,
      "FieldImpact": 300
    },
    "EnableMetrics": true
  },
```

It would be nice if you used the `IOptions` pattern, so it's easy to configure. Ideally, you'd register `IOptions<ExecutionOptions>` to the IOC container which would contain a default implementation, developers could then override this default implementation using the `Configure<ExecutionOptions>("foo")` method. Finally, use `IOptions<ExecutionOptions>` wherever you need to it.

pekkah:
@johnrutherford could you resolve the conflicts :)

johnrutherford:
I'm going to finally get back to this and finish it up.

johnrutherford:
@pekkah, sorry for the delay, but I've finally finished up this pull request.

pekkah:
Thank you very much for this!

Rikarin:
Thanks a lot!
Can someone deploy it to the official nuget repository, please? Thanks.

RehanSaeed:
Yes, an update to NuGet would be nice.

src/Core/GraphQLBuilderExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// <param name="builder"></param>
+        /// <param name="assembly"></param>
+        /// <returns></returns>
+        public static IGraphQLBuilder AddGraphTypes(this IGraphQLBuilder builder, Assembly assembly)
+        {
+            // Register all GraphQL types
+            foreach (var type in assembly.GetTypes()
+                .Where(x => !x.IsAbstract && typeof(IGraphType).IsAssignableFrom(x)))
+            {
+                builder.Services.TryAddTransient(type);
```
RehanSaeed: Why are the graph types being added as transient? Shouldn't they be added as singleton? If you were using EF Core, you might want to use scoped instead. Perhaps we should give the user the choice.

README.md
```diff
@@ -29,31 +29,26 @@ public void ConfigureServices(IServiceCollection services)
...
 
 public void Configure(IApplicationBuilder app, IHostingEnvironment env)
 {
     // this is required for websockets support
     app.UseWebSockets();
 
-    // use websocket middleware for ChatSchema at default url /graphql
-    app.UseGraphQLWebSocket<ChatSchema>(new GraphQLWebSocketsOptions());
+    // use websocket middleware for ChatSchema at path /graphql
+    app.UseGraphQLWebSockets<ChatSchema>("/graphql");
```
RehanSaeed: I think it will be rare that someone would want to change the path. Can `/graphql` be defaulted for `UseGraphQLWebSockets` and `UseGraphQL`?

samples/Samples.Server/Startup.cs
```diff
@@ -35,36 +32,29 @@ public void ConfigureServices(IServiceCollection services)
...
+
             app.UseDefaultFiles();
             app.UseStaticFiles();
 
             app.UseWebSockets();
-            app.UseGraphQLWebSocket<ChatSchema>(new GraphQLWebSocketsOptions());
-            app.UseGraphQLHttp<ChatSchema>(new GraphQLHttpOptions());
+            app.UseGraphQLWebSockets<ChatSchema>("/graphql");
+            app.UseGraphQL<ChatSchema>("/graphql");
             app.UseGraphQLPlayground(new GraphQLPlaygroundOptions()
```
RehanSaeed: Perhaps the path `/graphql` could be made part of `GraphQLOptions`, so that the `GraphQLPlaygroundOptions` does not need it.

Also, for consistency, it would be nice if you could do `UseGraphQLPlayground("ui/playground")`.

johnrutherford: That's an interesting idea to add the GraphQL endpoint to `GraphQLOptions`. I agree we should make all of the app builder extensions consistent.

Can you open an issue for this discussion since this PR is closed?

RehanSaeed: For now I've raised a PR https://github.com/graphql-dotnet/server/pull/153 to default the `/graphql` path.

PR #130, opened 2018-07-10
fix for http get support

OneCyrus:
PR #84 broke HTTP GET support. instead of disabling HTTP GET, which is defined by the graphql specs, this PR fixes the null exception when handling variables parsing.

### DIFF
--- a/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
+++ b/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
@@ -1,3 +1,4 @@
+using System;
 using System.IO;
 using System.Linq;
 using System.Net;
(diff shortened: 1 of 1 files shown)


SimonCropp:
@OneCyrus is it possible to add a test to verify GET support so it doesnt break again in the future?

OneCyrus:
sure, but there's no test project at all for transports.

OneCyrus:
btw. issue is already fixed with #128 

though a proper test project wouldn't be too bad.

cyberhck:
I'm seeing that GET is broken currently unfortunately, for some reason when browser tries to perform a GET request for websocket, this gets involved and throws an exception,
image

sungam3r:
It's rather old PR. Current status of code is that middleware supports both GET and POST. See https://github.com/graphql-dotnet/server/blob/master/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs#L16-L27 and https://github.com/graphql-dotnet/server/blob/master/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs#L16-L27

Exception from https://github.com/graphql-dotnet/server/pull/130#issuecomment-787506352 has lready been fixed an will be released in v5.1 soon (days).

PR #139, opened 2018-08-03
Use dependency injection to register custom IValidatorRule's

RehanSaeed:
Passing `IValidatorRule` to the options object is unnecessary and forces you to write a separate method as shown below because the List<IValidatorRule> property is readonly. A better way is to resolve them from the IOC container.

```
application.UseGraphQLHttp<MainSchema>(GetGraphQLHttpOptions(application));

private GraphQLHttpOptions GetGraphQLHttpOptions(IApplicationBuilder application)
{
    var foo = application.ApplicationServices.GetService<IWebHostBuilder>();

    var options = new GraphQLHttpOptions()
    {
        // Show stack traces in exceptions. Don't turn this on in production.
        ExposeExceptions = this.hostingEnvironment.IsDevelopment(),
    };

    // Add your own validation rules e.g. for authentication.
    foreach (var validationRule in application.ApplicationServices.GetServices<IValidationRule>())
    {
        options.ValidationRules.Add(validationRule);
    }

    return options;
}
```

Instead we can just do this:

```
services.AddSingleton<IValidatorRule, MyValidatorRule>();

application.UseGraphQLHttp<MainSchema>(
    new GraphQLHttpOptions()
    {
        // Show stack traces in exceptions. Don't turn this on in production.
        ExposeExceptions = this.hostingEnvironment.IsDevelopment(),
    });
```

### DIFF
--- a/src/AspNetCore/AspNetCore.csproj
+++ b/src/AspNetCore/AspNetCore.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0</TargetFrameworks>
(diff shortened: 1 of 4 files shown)


johnrutherford:
PR #128 implemented this for HTTP and web sockets. All registered `IValidatorRule`s are pulled in for execution with dependency injection.

PR #146, opened 2018-08-12
Release 3.1

pekkah:


pekkah:
@johnrutherford added

build.cake
```diff
@@ -8,10 +8,12 @@ var framework = Argument<string>("framework", "netstandard2.0");
...
 var projectFiles = new [] {
-  "./src/AspNetCore/AspNetCore.csproj",
-  "./src/WebSockets/WebSockets.csproj",
+  "./src/Transports.AspNetCore/Transports.AspNetCore.csproj",
+  "./src/Transports.Subscriptions.Abstractions/Transports.Subscriptions.Abstractions.csproj",
+  "./src/Transports.Subscriptions.WebSockets/Transports.Subscriptions.WebSockets.csproj",
   "./src/Ui.Playground/Ui.Playground.csproj",
-  "./src/Ui.GraphiQL/Ui.GraphiQL.csproj"
+  "./src/Ui.GraphiQL/Ui.GraphiQL.csproj",
+  "./src/Ui.Voyager/Ui.Voyager.csproj"
```
joemcbride: I believe we need to add `GraphQL.Server.Core` to this list so it gets built at as Nuget package?

johnrutherford: It’s not referenced as a NuGet package by the other projects, just as a project reference. So the Core dll should be included the nupkg for the projects that reference it. 

I don’t see much benefit in making it a separate package, but either way should work.

joemcbride: I’m fine with it being included as well. Though if you download the packaged nugets from AppVeyor it is not included.  So need to fix that somehow.

pekkah: I think it should be a NuGet package. Otherwise people will run into problems with different versions of it when installing more than one package including it.

src/Transports.AspNetCore/Transports.AspNetCore.csproj
```diff
@@ -16,10 +16,15 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="GraphQL" Version="2.0.0-alpha-864" />
-    <PackageReference Include="Microsoft.AspNetCore.Http.Abstractions" Version="2.0.1" />
+    <PackageReference Include="GraphQL" Version="[2.0.0, 2.1)" />
```
johnrutherford: I actually think it's better to not specify an upper limit unless a higher version has been released and there are known incompatibilities. 

Otherwise, when GraphQL 2.1 is released, people will have problems updating unless new version of these packages are released at the same time. It will prevent people from upgrading even if there are no compatibility issues. I've seen this happen with other projects and it's really annoying (e.g., https://github.com/openstacknetsdk/openstack.net/issues/656).

build.cake
johnrutherford: Yeah, that's true. Do you need me to do anything on this?

src/Transports.AspNetCore/Transports.AspNetCore.csproj
pekkah: I suppose the upper limit should be < 3.0.

johnrutherford: That would be better, but could still cause issues with GraphQL 3.0, even if it's not incompatible. It doesn't look like Microsoft specifies maximum version numbers for dependencies on their NuGet packages.

joemcbride: I support dropping the range.  If there are known compatibility issues I can see adding one in, but not specifying it gives flexibility on updating.

pekkah: dropping it

build.cake
pekkah: Adding Core to packaging

pekkah:
@joemcbride @johnrutherford could you take a another look at this. Should be ready to ship.

johnrutherford:
Should we squeeze #151 into this release?

Everything else looks good to me.

joemcbride:
Looks good to me.

src/Core/GraphQLBuilderExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// <param name="builder"></param>
+        /// <param name="assembly"></param>
+        /// <returns></returns>
+        public static IGraphQLBuilder AddGraphTypes(this IGraphQLBuilder builder, Assembly assembly)
+        {
+            // Register all GraphQL types
+            foreach (var type in assembly.GetTypes()
+                .Where(x => !x.IsAbstract && typeof(IGraphType).IsAssignableFrom(x)))
+            {
+                builder.Services.TryAddTransient(type);
```
RehanSaeed: Why are the graph types being added as transient? Shouldn't they be added as singleton? If you were using EF Core, you might want to use scoped instead. Perhaps we should give the user the choice.

PR #155, opened 2018-08-15
Add ability to pass service lifetime to AddGraphTypes

RehanSaeed:
This one needs some discussion. This PR adds the ability to set the service lifetime but what should the default be?

I think adding the types as transient is probably wrong. If you are using EF Core, then you want to add the types as scoped, otherwise you probably would be happy with singleton.

### DIFF
--- a/src/Core/GraphQLBuilderExtensions.cs
+++ b/src/Core/GraphQLBuilderExtensions.cs
@@ -27,25 +27,32 @@ namespace GraphQL.Server
         /// Add all types that implement <seealso cref="IGraphType"/> in the calling assembly
         /// </summary>
         /// <param name="builder"></param>
+        /// <param name="serviceLifetime">The service lifetime to register the GraphQL types with</param>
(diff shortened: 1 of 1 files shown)


RehanSaeed:
I noticed that a bunch of other types also being registered as transient in the latest 3.1 release. Can any of these also be registered as singleton?

```
services.AddTransient(typeof(IWebSocketConnectionFactory<>), typeof(WebSocketConnectionFactory<>))
services.AddTransient<IOperationMessageListener, LogMessagesListener>()
services.AddTransient<IOperationMessageListener, ProtocolMessageListener>();
services.AddTransient(typeof(IGraphQLExecuter<>), typeof(DefaultGraphQLExecuter<>));
```

johnrutherford:
Not sure about the message listener ones, but the others cannot or the Schema will always be a singleton.

RehanSaeed:
I noticed that I was registering the message listeners as singleton before this release.

PR #158, opened 2018-08-17
Add support for endpoint authorization

johnrutherford:
Make all IAppBuilders extension methods consistent
Add more tests


src/AspNetCore.Authorization/HttpAuthorizationHelper.cs
```diff
@@ -0,0 +1,82 @@
...
+{
+    public static class HttpAuthorizationHelper
+    {
+        /// <summary>
+        /// Authorize current user based on provided options
+        /// </summary>
+        /// <param name="context"></param>
+        /// <param name="policyName"></param>
+        /// <returns>True if the user was authorized and the request should proceed; False if the request should be ended immediately.</returns>
+        public static async Task<bool> AuthorizeAsync(HttpContext context, string policyName)
```
RehanSaeed: I would suggest modelling this after the `[Authorize]` attribute in ASP.NET Core. In that attribute you can pass:

- AuthenticationSchemes - (This is a comma delimeted string but only because you can't pass an array to an attribute). I think this one is important to provide an overload for.
- RoleName - I would recommend using policies but some people like to shoot themselves in the foot and use role names.
- ActiveAuthenticationSchemes - This is obsolete. Do not add this.

Also, is the `authorization` repo dead after this? I literally submitted a PR yesterday to use the ASP.NET Core Authorization package! Basically, what this PR is all about.

src/AspNetCore.Authorization/GraphQLBuilderExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace GraphQL.Server
+{
+    public static class GraphQLBuilderExtensions
+    {
+        /// <summary>
+        /// Add HTTP authorization services
+        /// </summary>
+        /// <param name="builder"></param>
+        /// <returns></returns>
+        public static IGraphQLBuilder AddHttpAuthorization(this IGraphQLBuilder builder)
```
RehanSaeed: Is the reason this is called `AddHttpAuthorization`, because subscriptions don't support auth?

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -21,27 +21,40 @@ public class GraphQLHttpMiddleware<TSchema>
...
         {
-            if (context.WebSockets.IsWebSocketRequest || !context.Request.Path.StartsWithSegments(_path))
+            if (context.WebSockets.IsWebSocketRequest || !context.Request.Path.StartsWithSegments(_options.Path))
             {
                 await _next(context);
                 return;
             }
 
+            // Authorize request
+            if (!await HttpAuthorizationHelper.AuthorizeAsync(context, _options.AuthorizationPolicyName))
```
RehanSaeed: Ok, I understand this now. So `AuthorizationPolicyName` is like a default policy for the entire API. When you call Microsofts `AddAuthorization` function you can already set a default policy there:

```
services.AddAuthorization(options => options.DefaultPolicy = "FooPolicy");
```

I think you should take a look at the code from my PR https://github.com/graphql-dotnet/authorization/pull/13 to be able to add auth policies to individual graph fields or graph types.

src/AspNetCore.Authorization/HttpAuthorizationHelper.cs
johnrutherford: I think it's best to keep this simple. You can create a policy that specifies the authentication schemes or require role(s). I used a role check for the tests in Transports.AspNetCore.Tests\AuthenticatedTestStartup.cs.

I think the stuff in the authorization repo is more for authorizing individual nodes. I felt there was a separate need to support transport authorization that integrates with ASP.NET Core authentication and will trigger the authentication challenge or forbidden responses.

src/AspNetCore.Authorization/GraphQLBuilderExtensions.cs
johnrutherford: It was more to differentiate it from the authorization of the GraphQL nodes. 

I also added authorization support for the WebSocket transport. WebSockets still uses HTTP(S) before the connection is upgraded.

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
RehanSaeed: I think `UseAuthorization()` actually applies the `DefaultPolicy` above. So there is already middleware that will achieve this built into ASP.NET Core.

johnrutherford: Confirmed it doesn't exist: https://github.com/aspnet/Security/issues/1734.

RehanSaeed: You're absolutely right, I must have imagined it. Looks like they are planning to add a middleware to do this though. If we can't wait for Microsoft to implement it, perhaps it should be a separate `UseDefaultAuthorizationPolicy` middleware.

Either way, we should still not add `AuthorizationPolicyName` to `GraphQLHttpMiddlewareOptions`. Instead we should use `DefaultPolicy` in the authorization options.

johnrutherford: I thought about using the `DefaultPolicy`, but what if you want to use the default policy for MVC and a different policy for your GraphQL endpoint? With the `AuthorizationPolicyName` you're able to specify whichever policy you want to use.

src/AspNetCore.Authorization/AspNetCore.Authorization.csproj
```diff
@@ -0,0 +1,30 @@
...
+    <PackageReference Include="Microsoft.AspNetCore.Authorization" Version="2.0.3" />
+    <PackageReference Include="Microsoft.AspNetCore.Authorization.Policy" Version="2.0.3" />
+    <PackageReference Include="Microsoft.AspNetCore.Http" Version="2.0.2" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Core\Core.csproj" />
+  </ItemGroup>
+
+</Project>
```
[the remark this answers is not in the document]
johnrutherford: Yeah, I was unsure since it brought in additional dependencies, but I think the consolidation would be better.

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -21,27 +21,40 @@ public class GraphQLHttpMiddleware<TSchema>
...
+            {
+                return;
```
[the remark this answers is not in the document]
johnrutherford: Correct.

src/AspNetCore.Authorization/HttpAuthorizationHelper.cs
```diff
@@ -0,0 +1,82 @@
...
+                    }
+                }
+                else
+                {
+                    await context.ForbidAsync();
+                }
+                return false;
+            }
+
+            return false;
```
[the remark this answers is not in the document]
johnrutherford: It shouldn't happen. Yeah, I think it would be a good idea to throw an exception there.

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
johnrutherford: If there was a generic middleware, I guess you could do something like this:

``` csharp
app.Map("/graphql", branch =>
{
    branch.UseAuthorization(policy: "IsAuthenticated");
    branch.UseGraphQL<TestSchema>("");
});
```

johnrutherford:
@joemcbride pushed some additional changes based on your feedback.

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
RehanSaeed: You can run multiple apps in a single ASP.NET Core application with separate DI containers, see [blog post](https://www.strathweb.com/2017/04/running-multiple-independent-asp-net-core-pipelines-side-by-side-in-the-same-application/). Not well known you can do this.

pekkah:
Is this now obsolete?

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -21,27 +21,40 @@ public class GraphQLHttpMiddleware<TSchema>
...
+
             // Handle requests as per recommendation at http://graphql.org/learn/serving-over-http/
             var httpRequest = context.Request;
-            var gqlRequest = new GraphQLRequest();
 
-            var writer = context.RequestServices.GetRequiredService<IDocumentWriter>();
+            if (!HttpMethods.IsGet(httpRequest.Method) && !HttpMethods.IsPost(httpRequest.Method))
+            {
+                context.Response.Headers.Add("Allow", "GET, POST");
+                context.Response.StatusCode = 405; // Method Not Allowed
```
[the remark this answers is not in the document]
benmccallum: The recommended approach from MS is now to use `Microsoft.AspNetCore.Http.StatusCodes`, as it's a static class with const fields therefore implying more codes can come as opposed to the old HttpStatusCode enum. Don't have a source, they explained it better, was in an aspnet repo though.

sungam3r:
@johnrutherford Is this PR still relevant after #440 ?

sungam3r:
I see that at least part of the changes in this PR does what has already been done. Some of the code has already been deprecated and has been removed. The remaining code, as I understand it, establishes certain authorization rules. Authorization is now encapsulated in a separate project. Will you keep working on this PR or is it not worth it?

sungam3r:
@johnrutherford @RehanSaeed friendly bump

sungam3r:
@johnrutherford If there is no more work on this PR, then I'm going to close it after the merging #477 . Otherwise, you need to resolve conflicts and update PR.

sungam3r:
Closed as stale.

PR #171, opened 2018-10-10
Add GraphQL.Server.Authorization.AspNetCore NuGet package

RehanSaeed:
The current [Authorization](https://github.com/graphql-dotnet/authorization) NuGet package duplicates code from `Microsoft.AspNetCore.Authorization` to work around some issues with .NET 4.6 and binding redirects. This causes all kinds of problems for ASP.NET Core users because the API has been slightly changed.

I initially submitted a PR https://github.com/graphql-dotnet/authorization/pull/13 to make use of `Microsoft.AspNetCore.Authorization` directly instead in the [Authorization](https://github.com/graphql-dotnet/authorization) repo. After a long discussion in https://github.com/graphql-dotnet/authorization/issues/11 with @joemcbride and others it was decided that this should be a separate project.

Some changes were made in the authorization repo by @joemcbride to support argument fields. I have updated the code here to reflect that. This PR is mostly copying the code from the authorization repo but adding in ASP.NET Core idioms. It also copies across calls to `.GetAwaiter().GetResult()` which is bad but fixing this requires changes to the core GraphQL project (not looked at how to fix that in too much detail yet) which I think can be done in a separate PR.

### DIFF
--- a/src/Authorization.AspNetCore/Authorization.AspNetCore.csproj
+++ b/src/Authorization.AspNetCore/Authorization.AspNetCore.csproj
@@ -18,6 +18,7 @@
   <ItemGroup Label="Package References">
     <PackageReference Include="GraphQL" Version="2.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.Authorization" Version="2.0.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Http.Abstractions" Version="2.0.0" />
   </ItemGroup>
 
   <ItemGroup Label="Project References">
--- a/src/Authorization.AspNetCore/Authorization.AspNetCore.csproj
+++ b/src/Authorization.AspNetCore/Authorization.AspNetCore.csproj
@@ -18,7 +18,7 @@
   <ItemGroup Label="Package References">
     <PackageReference Include="GraphQL" Version="2.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.Authorization" Version="2.0.0" />
-    <PackageReference Include="Microsoft.AspNetCore.Http.Abstractions" Version="2.0.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Http" Version="2.0.0" />
   </ItemGroup>
 
   <ItemGroup Label="Project References">
--- a/src/Authorization.AspNetCore/Authorization.AspNetCore.csproj
+++ b/src/Authorization.AspNetCore/Authorization.AspNetCore.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup Label="Build">
     <TargetFrameworks>netstandard2.0</TargetFrameworks>
@@ -17,7 +17,7 @@
 
   <ItemGroup Label="Package References">
     <PackageReference Include="GraphQL" Version="2.0.0" />
-    <PackageReference Include="Microsoft.AspNetCore.Authorization" Version="2.1.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Authorization" Version="2.0.0" />
   </ItemGroup>
 
   <ItemGroup Label="Project References">
--- a/src/Authorization.AspNetCore/Authorization.AspNetCore.csproj
+++ b/src/Authorization.AspNetCore/Authorization.AspNetCore.csproj
@@ -1,0 +1,27 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup Label="Build">
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <AssemblyName>GraphQL.Server.Authorization.AspNetCore</AssemblyName>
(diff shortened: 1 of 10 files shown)


benmccallum:
Might wanna just update `GraphQlBuilderExtensions.cs` to `GraphQLBuilderExtensions.cs` for consistency.

RehanSaeed:
Done

benmccallum:
Nice. I guess we see what @pekkah and @joemcbride say 👍

src/Authorization.AspNetCore/Authorization.AspNetCore.csproj
```diff
@@ -0,0 +1,27 @@
...
+    <PackageProjectUrl>https://github.com/graphql-dotnet/server</PackageProjectUrl>
+    <RepositoryUrl>https://github.com/graphql-dotnet/server</RepositoryUrl>
+    <RepositoryType>Git</RepositoryType>
+    <PackageTags>GraphQL authentication authorization middleware</PackageTags>
+    <Copyright>Pekka Heikura</Copyright>
+  </PropertyGroup>
+
+  <ItemGroup Label="Package References">
+    <PackageReference Include="GraphQL" Version="2.0.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Authorization" Version="2.1.0" />
```
johnrutherford: The other projects in this repository are currently referencing 2.0.x releases of the ASP.NET Core packages. I think this should be consistent with those.

src/Authorization.AspNetCore/AuthorizationMetadataExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public static FieldBuilder<TSourceType, TReturnType> AuthorizeWith<TSourceType, TReturnType>(
+            this FieldBuilder<TSourceType, TReturnType> builder, string policy)
+        {
+            builder.FieldType.AuthorizeWith(policy);
+            return builder;
+        }
+
+        public static List<string> GetPolicies(this IProvideMetadata type) =>
+            type.GetMetadata(PolicyKey, new List<string>());
```
johnrutherford: This is going to create a new `List<string>` every time this method is called. We should just let it default to `null`.

src/Authorization.AspNetCore/AuthorizationValidationRule.cs
```diff
@@ -0,0 +1,179 @@
...
+        private readonly IAuthorizationService _authorizationService;
+
+        public AuthorizationValidationRule(IAuthorizationService authorizationService)
+        {
+            _authorizationService = authorizationService;
+        }
+
+        public INodeVisitor Validate(ValidationContext context)
+        {
+            var userContext = context.UserContext as IProvideClaimsPrincipal;
```
johnrutherford: Since this is always going to be in ASP.NET Core, why not inject `IHttpContextAccessor` and use that to get  `HttpContext.User`? 

Then there's no special requirement for the `UserContext`.

src/Authorization.AspNetCore/Authorization.AspNetCore.csproj
RehanSaeed: Done.

src/Authorization.AspNetCore/AuthorizationMetadataExtensions.cs
RehanSaeed: Done.

src/Authorization.AspNetCore/AuthorizationValidationRule.cs
RehanSaeed: Done. Also removed the IProvideClaimsPrincipal interface as I think it's no longer required.

src/Authorization.AspNetCore/GraphQlBuilderExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        /// <summary>
+        /// Adds the GraphQL authorization.
+        /// </summary>
+        /// <param name="builder">The GraphQL builder.</param>
+        /// <returns></returns>
+        public static IGraphQLBuilder AddGraphQLAuthorization(this IGraphQLBuilder builder)
+        {
+            builder
+                .Services
```
johnrutherford: We should add this line in these methods to make sure `IHttpContextAccessor` is registered. It is not by default.

``` csharp
builder.Services.TryAddSingleton<IHttpContextAccessor, HttpContextAccessor>();
```

joemcbride:
> if IValidationRule could provide access to the HttpContext somehow via the ValidationContext we could get rid of IHttpContextAccessor.

Probably won't be adding a direct reference to that anytime soon since `HttpContext` would need to be specific to the server platform you're using (full-framework vs. core), which the main library has no references to.

I could see wanting to pass a ClaimsPrincipal though and provide a way to retrieve it.  Then the end user of the framework can decide how to fetch it.

benmccallum:
Woot. Nice one. Can someone bump this once it's released to nuget? Going to swap over to this when I can 👍

benmccallum:
Any word on this? Doesn't seem to be a new nuget release of current package or any new packages that I can see. Currently running a copy of the proj in my sln that I'd love to clean out :)

PR #172, opened 2018-10-19
Rename GraphQlBuilderExtensions.cs to GraphQLBuilderExtensions.cs

benmccallum:
He'd fixed up the class name but not the file name.


RehanSaeed:
I thought I already did that. Fixing this issue on cases insensitive Windows is a pain.

benmccallum:
Yep, I thought that might be the case. Had th same issue when I pulled it as a local project while I wait for the release. Had to rename twice. All g.

PR #174, opened 2018-10-28
Updated: websocket implementation to support documentwriter.

DSilence:
Implemented support for new DocumentWriter for websocket transport.
Things to note:

1. Writing to socket is achieved by using a proxy stream object - WebsocketWriterStream. It solves several issues:
    
    -  The initial idea of DocumentWriter method has a memory limit of 1 mb per message

    - There is some overly complex code around buffers in DocumentWriter implementation. Reducing it to a single method that writes to stream is great.
There is a con, however. There is always a need to call the websocket API at lease twice. Not sure how this will affect the performance and not sure how to test this (perhaps, some load test?).

2. The ProtocolMessageListener was sending a message.Id twice, once as part of ExecutionResult, and once in OperationMessage. I believe this is not requred, but I may be wrong.

Let me know  what you think! Should this approach be accepted, it should be possible to remove the "extra" method within the DocumentWriter, and always stick to Stream API for serialization.

### DIFF
--- a/src/Transports.Subscriptions.WebSockets/WebsocketWriterStream.cs
+++ b/src/Transports.Subscriptions.WebSockets/WebsocketWriterStream.cs
@@ -33,7 +33,7 @@ namespace GraphQL.Server.Transports.WebSockets
 
         public override int Read(byte[] buffer, int offset, int count)
         {
-            throw new System.NotImplementedException();
+            throw new System.NotSupportedException();
         }
 
         public override long Seek(long offset, SeekOrigin origin)
--- a/src/Transports.Subscriptions.WebSockets/WebsocketWriterStream.cs
+++ b/src/Transports.Subscriptions.WebSockets/WebsocketWriterStream.cs
@@ -21,6 +21,11 @@ namespace GraphQL.Server.Transports.WebSockets
                 cancellationToken);
         }
 
+        public override void Write(byte[] buffer, int offset, int count)
+        {
+            WriteAsync(buffer, offset, count).GetAwaiter().GetResult();
+        }
+
         public override Task FlushAsync(CancellationToken cancellationToken)
         {
             return _webSocket.SendAsync(new ArraySegment<byte>(Array.Empty<byte>()), WebSocketMessageType.Text, true, cancellationToken);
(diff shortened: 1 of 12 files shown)


src/Transports.Subscriptions.WebSockets/WebsocketWriterStream.cs
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        public override void SetLength(long value)
+        {
+            throw new System.NotImplementedException();
+        }
+
+        public override void Write(byte[] buffer, int offset, int count)
+        {
+            throw new System.NotImplementedException("Synchronous methods are not supported by WebsocketWriterStream.");
```
johnrutherford: I believe `JsonTextWriter` will end up calling the synchronous `Write()` method, so I think this will need to be supported.

src/Transports.Subscriptions.WebSockets/WebsocketWriterStream.cs
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        public override void Flush()
+        {
+            throw new System.NotImplementedException("Synchronous methods are not supported by WebsocketWriterStream.");
+        }
+
+        public override int Read(byte[] buffer, int offset, int count)
+        {
+            throw new System.NotImplementedException();
```
johnrutherford: Not a big deal, but in this and other methods, `NotSupportedException` would be more appropriate than `NotImplementedException`.

src/Transports.Subscriptions.WebSockets/WebsocketWriterStream.cs
DSilence: The implementation from Asp.Net Core Mvc (HttpResponseStreamWriter) should always use async method (at least in theory, in my tests, sync methods were never called). 
But I do agree that sync write and flush methods should be implemented (especially considering that alternative DocumentWriter implementation can be used). Perhaps it's worth to add a warning message to logs to notify the users that sync method was used as well.

DSilence: After taking a second look at the code, the fully async flow will only be used if the response is smaller than StreamWriter buffer. Implementing sync method is definitely required.

johnrutherford: Yeah, that's why I think unit tests would be good. With different size responses.

johnrutherford:
@pekkah, the custom stream is required to be able to write to the websocket in chunks from a buffer. The alternative is to serialize the whole response to memory and then write the whole byte array at once, but this uses more RAM.

DSilence:
Added unit tests for payloads of varying size (from several bytes, to 100 mb).

PR #177, opened 2018-11-08
Fix the issue when the client prematurely closes and the server canno…

Eldar-Ahmadov:
…t complete the websocket reader pipeline

When a GraphQL client is subscribed and prematurely closes (e.g. the app crashes, or is simply closed), the   WebSocketReaderPipeline tries to close the socket by calling _socket.CloseAsync which never returns.

Even though Microsoft's documentation on CloseAsync and CloseOutputAsync is quite similar, according to the below links:
- https://stackoverflow.com/a/26744644
- http://www.salmanq.com/blog/5-things-you-probably-didnt-know-about-net-websockets/2013/04/

CloseAsync waits for the client to acknowledge closing of the socket, whereas  CloseOutputAsync works in fire-and-forget mode.

It looks like the use of those methods should be swapped to achieve the correct semantics, so that CloseAsync does not get stuck waiting for the client's acknowledgement (the client may already be closed by now) and hence never complete the reader pipeline.

### DIFF
--- a/src/Transports.Subscriptions.WebSockets/WebSocketReaderPipeline.cs
+++ b/src/Transports.Subscriptions.WebSockets/WebSocketReaderPipeline.cs
@@ -44,25 +44,30 @@ namespace GraphQL.Server.Transports.WebSockets
         {
             if (_socket.State != WebSocketState.Closed && _socket.State != WebSocketState.CloseSent)
             {
-                if (closeStatus == WebSocketCloseStatus.NormalClosure)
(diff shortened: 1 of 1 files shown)


fayilt:
Well spotted!

Eldar-Ahmadov:
Replaced tabs with spaces.

pekkah:
Still some formatting issues in the code. Functions wont align properly. Have you tried format document in VS?

PR #181, opened 2018-11-27
fix Schema.FieldNameConverter not work

Edward-Zhou:
add Schema FieldNameConverter to overrride default CamelCaseFieldNameConverter

### DIFF
--- a/src/Core/Internal/DefaultGraphQLExecuter.cs
+++ b/src/Core/Internal/DefaultGraphQLExecuter.cs
@@ -65,6 +65,9 @@ namespace GraphQL.Server.Internal
                 .Concat(DocumentValidator.CoreRules())
                 .ToList();
 
+            // add Schema FieldNameConverter to overrride default CamelCaseFieldNameConverter
+            opts.FieldNameConverter = Schema.FieldNameConverter;
+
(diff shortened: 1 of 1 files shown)


src/Core/Internal/DefaultGraphQLExecuter.cs
```diff
@@ -65,6 +65,9 @@ protected virtual ExecutionOptions GetOptions(string operationName, string query
                 .Concat(DocumentValidator.CoreRules())
                 .ToList();
 
+            // add Schema FieldNameConverter to overrride default CamelCaseFieldNameConverter
+            opts.FieldNameConverter = Schema.FieldNameConverter;
```
BenjaBobs: I suggest putting this in the initializer where the other options are set, at line 57.

```c#
var opts = new ExecutionOptions()
{
	// ...
	FieldNameConverter = Schema.FieldNameConverter
}
```

Edward-Zhou: I changed it.

sungam3r:
@Edward-Zhou I closed it because the necessary changes are made in #256 .

Issue #184, opened 2018-12-13
playground is loading schema directly from the baseurl, not considering the string path after first slash


pobisw:
playground.cshtml has the below code which sets the url for fetching the schema  in playground
```
  window.addEventListener('load', function (event) {
      GraphQLPlayground.init(document.getElementById('root'),
        {
          setTitle: true,
          endpoint: window.location.protocol + "//" + window.location.host + "@Model.GraphQLEndPoint",
          subscriptionEndpoint: (window.location.protocol === "http:" ? "ws://" : "wss://") + window.location.host + "@Model.GraphQLEndPoint",
          settings: {
            'tracing.hideTracingResponse': false
          }
        });
    })
```
This only consider the localtion.host  part, but not the url path after that 

for example if I load https://basepath/path1/path2/playground 

it will try to load schema from https://basepath/graphql 

because that is the value that comes in window.location.host

Expected Behavior :  It should load schema from 

https://basepath/path1/path2/graphql

joemcbride:
Can you describe why you can’t just set the url of your GraphQL Endpoint to the path that you need?  It appears the changes you are making to this would make it impossible to have different root urls for those two endpoints, which is a valid scenario.

pobisw:
@joemcbride  because we have multiple dev and int endpoints for graphql, and we have can set the graphql endpoint to only one value

pobisw:
> Can you describe why you can’t just set the url of your GraphQL Endpoint to the path that you need? It appears the changes you are making to this would make it impossible to have different root urls for those two endpoints, which is a valid scenario.

@joemcbride  can you please reply, if the explanation answers your concern. If not please suggest what would be the better way to support multiple dev and int environment for graphql ?

subithal:
I'm also facing the same issue.  was it resolved??

subithal:
@sungam3r is this issue resolved?

sungam3r:
I think no. I have no time to work on each project in GraphQL org. Now in my free time I'm busy preparing the GraphQL.NET v4. Then we will publish server v5. It is possible that some of these errors will be fixed in v5 release.

pobisw:
I raised a PR, but nobody approved it, if someone approves it the issue
will get resolved

sungam3r:
Every PR needs time to review.

pobisw:
It's been a year 🙂

sungam3r:
I know. I myself have such PRs that have not being merged for more than a year.

[...]

PR #185, opened 2018-12-14
Fixing path string issue for playground schema fetch

pobisw:
Fix for #184 
Fix to consider path string for creation of graphql schema endpoint in playground

### DIFF
--- a/src/Transports.AspNetCore/Transports.AspNetCore.csproj
+++ b/src/Transports.AspNetCore/Transports.AspNetCore.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0</TargetFrameworks>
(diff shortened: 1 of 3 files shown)


BenjaBobs:
Is this also a problem for GraphiQL?

pobisw:
> Is this also a problem for GraphiQL?

Yeah you are right, the url is extracted in the same way in GraphIql as well 

`fetch(window.location.protocol + "//" + window.location.host + '@Model.GraphQLEndPoint', {
				method: 'post',
				headers: {
					'Accept': 'application/json',
					'Content-Type': 'application/json',
				}`

BenjaBobs:
I suspect it's also the case for Voyager. Maybe there should be a common way to access the full api url.

pobisw:
> I suspect it's also the case for Voyager. Maybe there should be a common way to access the full api url.

yes it's the same code everywhere, let try to make 

> Hmm the commit history looks bit messy
Any Idea what went wrong ?

pekkah:
Hmm I think you created this PR from master and that's why the commits looks funny. Could you rebase against develop?

pobisw:
The merge is not resolving, so closing this and creating a fresh PR here 
#195

PR #195, opened 2019-01-12
fix: url issue for path string for ui tools (playground, voyager, graphiql)

### LINKED ISSUE #184: playground is loading schema directly from the baseurl, not considering the string path after first slash

pobisw:
fixes #184


pobisw:
@pekkah @BenjaBobs  Can you please take a look at this PR 
added the fix for all the 3 UI tool

pobisw:
@joemcbride  @pekkah  please review

src/Ui.GraphiQL/Internal/graphiql.cshtml
```diff
@@ -98,23 +98,22 @@
 			updateURL();
```
joemcbride: This updateUrl is used here but you removed the function.

pobisw: got removed mistakenly, added back

pobisw:
> I mentioned this in the other issue, though this seems to make it so you can’t have your GraphQL endpoint at a different root url, which is a valid scenario. Is there something blocking you from setting that url to the path that you need?

I have replied in the other issue,  restating it here, we have multiple int and dev root urls

sungam3r:
@pobisw Thanks for this PR. I will try to review it if you resolve conflicts. Also, could you not change the indentation in the `cshtml `files so as not to complicate the review?

Shane32:
> @pobisw Thanks for this PR. I will try to review it if you resolve conflicts. Also, could you not change the indentation in the `cshtml `files so as not to complicate the review?

Just FYI, you can turn off whitespace comparison on github's website, @sungam3r

Shane32:
image

sungam3r:
OK

pobisw:
just saw the comments, will make the changes

src/Ui.GraphiQL/Internal/graphiql.cshtml
```diff
@@ -96,25 +96,34 @@
...
+        }
+
+        function stripGraphQLUiEndPointFromPathName(graphQlUiEndpoint) {
+            var currentPathName = window.location.pathname;
+            var regex = new RegExp(graphQlUiEndpoint + "\/*");
+            currentPathName = currentPathName.replace(regex, '');
+            return currentPathName;
+        }
+
+        var strippedCurrentPathName = stripGraphQLUiEndPointFromPathName("@Model.GraphiQLPath");
```
Shane32, line 118: This injected variable appears to be encoded for HTML instead of for Javascript - correct?  Or does Razor detect that it's within a `<script>` tag and encode it differently?

src/Ui.GraphiQL/Internal/graphiql.cshtml
```diff
@@ -96,25 +96,34 @@
...
+            }).map(function (key) {
+                return encodeURIComponent(key) + '=' +
+                    encodeURIComponent(parameters[key]);
+            }).join('&');
+            history.replaceState(null, null, newSearch);
+        }
+
+        function stripGraphQLUiEndPointFromPathName(graphQlUiEndpoint) {
+            var currentPathName = window.location.pathname;
+            var regex = new RegExp(graphQlUiEndpoint + "\/*");
```
Shane32, line 113: Here the path name from the model variable is passed directly into a regular expression without encoding.  Also seems wrong.  Maybe this should not be using regexp at all and just perform a simple replace...

src/Ui.GraphiQL/Internal/graphiql.cshtml
```diff
@@ -96,25 +96,34 @@
...
 
 		// Defines a GraphQL fetcher using the fetch API. You're not required to
 		// use fetch, and could instead implement graphQLFetcher however you like,
 		// as long as it returns a Promise or Observable.
-		function graphQLFetcher(graphQLParams) {
+        function graphQLFetcher(graphQLParams) {
 			// This example expects a GraphQL server at the path /graphql.
 			// Change this to point wherever you host your GraphQL server.
-			return fetch(window.location.protocol + "//" + window.location.host + '@Model.GraphQLEndPoint', {
+            return fetch(window.location.protocol + "//" + window.location.host + strippedCurrentPathName + '@Model.GraphQLEndPoint', {
```
Shane32, line 126: We should update the encoding of `Model.GraphQLEndPoint` if necessary as well.  For example `@JavaScriptEncoder.Default.Encode(Model.GraphQLEndPoint)`.

src/Ui.GraphiQL/Internal/graphiql.cshtml
sungam3r: Razor? No Razor here 🤔 .

src/Ui.GraphiQL/Internal/graphiql.cshtml
sungam3r: `JavaScriptEncoder` ?  Razor stuff?

sungam3r:
> just saw the comments, will make the changes

@pobisw  Thanks for the fix. The main thing here is to resolve conflicts, after which we can continue review.

src/Ui.GraphiQL/Internal/graphiql.cshtml
Shane32: cshtml are razor pages and `@Model.GraphiQLPath` is razor injecting a variable. By default it will format variables with html encoding. So I always use JavaScript encoding as shown below when injecting a variable into a script tag like this. Or encode it as json and then inject it. Or put it into a html tag and read the tag.

sungam3r: > cshtml are razor pages

Yes but ... no Razor engine/stuff/handlers/compilers here.

sungam3r: Just good old `string.Replace`.

Shane32: Ok I see.  Then we should modify the `string.Replace` encode to javascript-encode the variables for use on the page.  Probably not a critical fix as the url is controlled by the setup code, not user input.

sungam3r: 👍

Issue #166, opened 2018-09-18
Enable GraphQL Playground Tracing


RehanSaeed:
GraphQL Playground disables tracing by default. I've enabled it in what I thought would be a PR [here](https://github.com/graphql-dotnet/server/commit/de433c4d92b3c86d0d1ef0979960499f9f8e53f2) but I forgot that I'm now a contributor, so when I hit the edit button it does a commit instead, Doh!

Does someone with greater authority want to review or rollback said commit?

PR #198, opened 2019-01-23
Add GraphQL Playground settings to settable options

Cito:
Make GraphQL Playground settings and GraphQL config settable via `GraphQLPlaygroundOptions`.

Reverted the changed default setting (show tracing, see #166), since this is customizable now. Better to stick with the default settings of Playground (principle of least astonishment).

The Samples.Server shows a usage example.

### DIFF
--- a/src/Ui.Playground/GraphQLPlaygroundOptions.cs
+++ b/src/Ui.Playground/GraphQLPlaygroundOptions.cs
@@ -1,4 +1,5 @@
 using Microsoft.AspNetCore.Http;
+using System.Collections.Generic;
 
 namespace GraphQL.Server.Ui.Playground {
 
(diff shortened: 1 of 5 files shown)


src/Ui.Playground/GraphQLPlaygroundOptions.cs
```diff
@@ -11,6 +11,16 @@ public class GraphQLPlaygroundOptions {
...
 
+        /// <summary>
+        /// The GraphQL Config (as JSON string)
+        /// </summary>
+        public string GraphQLConfig { get; set; } = null;
+
+        /// <summary>
+        /// The GraphQL Playground Settings (as JSON string)
+        /// </summary>
+        public string PlaygroundSettings { get; set; } = null;
```
joemcbride: I think I would lean towards this being a Dictionary<string, object> vs just a string.  Then serialize that to JSON.

joemcbride: I know that is how the other property is setup, just seems a bit odd to me.

Cito: Ok, changed to Dictionary now. Using a string smelled a bit strange, that's right.

ebicoglu:
I need to make `same-origin` as default for `request.credentials`.
when this PR will be merged?

andrejohansson:
Same question here as @ebicoglu, any idea when these changes show up in the nuget releases? Or can I get pre-releases somewhere?

rahulpandey8920:
Need to use `"request.credentials": "include"` so that cookies are automatically sent. Waiting for this to be released. Thanks

PR #210, opened 2019-03-13
Null default options value for all middleware extensions

### LINKED ISSUE #209: null default value for all middleware extensions
It would be good if for the `UseGraphiQLServer/UseGraphQLPlayground/UseGraphQLVoyager` methods it was not necessary to set the options parameter even if it is `null`. Passing `null` looks unnatural.

sungam3r:
Fixes #209 
Also includes code cleanup.


joemcbride:
Hey, I can appreciate trying to clean up the code a bit, but that makes it really difficult to review the actual real changes.  Some things like switching to nested ifs vs. returning early, spacing, are things that really don't need to be changed and are more preferences vs. actual issues.  Can you please do the clean up in a separate PR (with a warning that perhaps it won't be merged) vs. adding / updating features?  Thanks.

sungam3r:
Ok

PR #213, opened 2019-03-18
Register GraphQLDependencyResolver in AddGraphQL

bkaid:


sungam3r:
`IDependencyResolver` was marked as obsolete. You can switch to `System.IServiceProvider` instead. But I think that it is not required because microsoft implementation already has support for getting `IServiceProvider`. With the transition to `IServiceProvider` things are somewhat simplified.

sungam3r:
As you can see in graphql-dotnet/examples#25 di container works fine even without explicit registration of `IDependencyResolver/IServiceProvider` so close it. Feel free to post a new issue if the problem is not resolved.

PR #216, opened 2019-03-22
If metrics are enabled enrich result with Apollo tracing

ahaeber:
Enriching results with metrics seems to be the default case if metrics have been enabled in the GraphQLOptions.

This is inspired by @RehanSaeed 's [Dotnet-Boxed template for GraphQL](https://github.com/Dotnet-Boxed/Templates/blob/master/Docs/GraphQL.md).


sungam3r:
@ahaeber I made a new PR with the necessary edits.

sungam3r:
Сlosed because of #254 .

Issue #217, opened 2019-03-28
Need new prerelease packages for 3.0.0 prerelease version of GraphQL


sungam3r:
See [#1058](https://github.com/graphql-dotnet/graphql-dotnet/pull/1058)

rose-a:
Im running into this issue too! Any news on when that pre-release package will be available?

pekkah:
Latest previews have been always available from myget

sungam3r:
In order for the package to appear on myget, we do not need to merge the code into the master? Will it be enough to make PR with prerelease versions of dependencies?

sungam3r:
The problem is solved.

PR #218, opened 2019-03-28
ExecutionOptions.UserContext: object -> IDictionary<string, object>

sungam3r:
draft fix for #217

### DIFF
--- a/src/Transports.Subscriptions.Abstractions/MessageHandlingContext.cs
+++ b/src/Transports.Subscriptions.Abstractions/MessageHandlingContext.cs
@@ -1,10 +1,11 @@
 ﻿using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 
 namespace GraphQL.Server.Transports.Subscriptions.Abstractions
 {
-    public class MessageHandlingContext : IDisposable
+    public class MessageHandlingContext : Dictionary<string, object>, IDisposable
     {
         private readonly IServerOperations _server;
 
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,6 @@
+<Project>
+  <PropertyGroup>
+    <LangVersion>latest</LangVersion>
+    <GraphQLPackageVersion>3.0.0-preview-1149</GraphQLPackageVersion>
+  </PropertyGroup>
+</Project>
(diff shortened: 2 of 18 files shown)


src/Transports.Subscriptions.Abstractions/MessageHandlingContext.cs
```diff
@@ -1,10 +1,11 @@
 ﻿using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 
 namespace GraphQL.Server.Transports.Subscriptions.Abstractions
 {
-    public class MessageHandlingContext : IDisposable
+    public class MessageHandlingContext : Dictionary<string, object>, IDisposable
```
sungam3r: Not sure about that. Inside there is already a `ConcurrentDictionary` for some properties. Not yet thoroughly understood this repository.

pekkah:
Because originally user context was object we needed to hijack it so we could add more properties into it. That's why on server side it already is a dictionary. The original user context is injected into it and available on resolvers.

sungam3r:
Need to make some more fixes?

joemcbride:
@sungam3r Would you mind fixing this merge conflict and I'll get this merged?  There may be some other refactoring that we need to do for subscriptions, though this should at least get the core project working.

sungam3r:
Yes, I want to do it, but I did not understand the comment from @pekkah 
> Because originally user context was object we needed to hijack it so we could add more properties into it. That's why on server side it already is a dictionary. The original user context is injected into it and available on resolvers.

I suspect I'm doing something wrong. I will come back to this a little later, I need to figure it out.

joemcbride:
> I suspect I'm doing something wrong.

No, it was how it was designed.  Subscriptions needed to pass along some additional metadata.  @pekkah accomplished that with the `UserContext`.  But, that would override the existing `UserContext`.  So he devised a way to work around that.  Now that `UserContext` is a Dictionary this can be refactored.  I can do that work, I just want to get what you have accomplished so far committed.

sungam3r:
You can do it yourself if it is required as quickly as possible. I still need some time to figure it out myself. I do not want to commit things if I do not fully understand what I have done. Now honestly, I am very tired, I can only return to this tomorrow.

joemcbride:
No rush.  In fact, it looks like there is only one file conflicting.  I'll take care of it.

joemcbride:
@sungam3r good news, subscriptions still work with those changes.  🎉

sungam3r:
Good

PR #220, opened 2019-04-05
feat: allow use of custom executor.

benmccallum:
Reason being, I'd like to add some cross-cutting validation rules and this seems like the easiest way.

With this, I can implement my own GraphQLExecuter, inheriting your default, and pass in more validation rules.

Only thing I'm not sure about here is there's no type safety between what user provides and `IGraphQLExecuter<>`. Not sure if we care though, it'll blow up in their faces straight away. Alternatively, I toyed with the idea of making the option a public getter only and requiring it be set through a method that could enforce the type through generics.


benmccallum:
Could obviously just do what you're doing on the `ServiceCollection` directly, but I like the certainty that after an upgrade I won't be missing out on startup config stuff.

benmccallum:
I'm sure I may have a need for more validators and still need this PR, but the one I want to specifically add right now is a "MaximumPageSize" validator for connection fields. Perhaps I should do a PR for that to `graphql-dotnet` adding one under [here](https://github.com/graphql-dotnet/graphql-dotnet/tree/master/src/GraphQL/Validation/Rules)?

benmccallum:
Hey @joemcbride or @pekkah, reckon we can get this merged and released? It's a super simple change :)

src/Core/ServiceCollectionExtensions.cs
```diff
@@ -19,7 +19,7 @@ public static class ServiceCollectionExtensions
         public static IGraphQLBuilder AddGraphQL(this IServiceCollection services, GraphQLOptions options)
         {
             services.TryAddSingleton<IDocumentExecuter, DocumentExecuter>();
-            services.AddTransient(typeof(IGraphQLExecuter<>), typeof(DefaultGraphQLExecuter<>));
+            services.AddTransient(typeof(IGraphQLExecuter<>), options.GraphQLExecuterType);
```
joemcbride: I am thinking this should be instead changed to `TryAddTransient` vs. adding the type to the options.  That way you can control it with your IoC container vs. adding an option (just like the `IDocumentExecuter`).

benmccallum: What would that look like? Since DefaultGraphQLExecutor is generic, I tried this, which compiles but it breaks at runtime. `services.TryAddTransient(typeof(IGraphQLExecuter), typeof(DefaultGraphQLExecuter<>));`

benmccallum: Ignore that. Literally just need to add the word `Try` to the existing code haha.

benmccallum:
Hi @joemcbride, I've changed this over to be configurable by DI as per your suggestion. Ready when you are. Thanks :)

sungam3r:
But what about failed tests?

benmccallum:
> But what about failed tests?

Strange. They pass on my machine. Will take a look.

benmccallum:
Re-pushed a refactor on some unused fields to retrigger build and tests pass... so... 🤷‍♂

benmccallum:
Hey @joemcbride or @pekkah, think we could merge this now and get a release out? All tests passing, literally just gives someone the option to register their own `IGraphQLExecutor`, so no risk. Please and thank you!

PR #235, opened 2019-07-11
Do not dispose request stream

### LINKED ISSUE #234: ObjectDisposedException using middlewares with HttpRequest.EnableBuffering()
`GraphQLHttpMiddleware` uses `StreamReader` to read data from request stream. `StreamReader` closes underlying stream (`leaveopen` parameter default value - `false`). So if using other "wrapping" middlewares (for example one for tracing request/response body) with call to `HttpRequest.EnableBuffering()` it leads to `ObjectDisposedException`. We should leave the underlying stream open to allow other components to use it when necessary. And ASP.NET Core request pipeline will take care of the destruction of this stream at the right time.

sungam3r:
fixes #234

### DIFF
--- a/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
+++ b/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
@@ -1,9 +1,3 @@
-using System.Collections.Generic;
-using System.IO;
-using System.Linq;
-using System.Net;
-using System.Net.Http.Headers;
-using System.Threading.Tasks;
 using GraphQL.Http;
 using GraphQL.Server.Internal;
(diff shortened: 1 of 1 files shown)


sungam3r:
@joemcbride  Please, can you check PR it if you have time? One of our services is currently not working due to this error. Of course, I think that we can find another solution on our side, a messier one, but I would just like to upgrade the package to a new preview version. Thanks!

sungam3r:
@joemcbride  Thank you very much! I took the packages from MyGet feed. Today we will try to install them.

sungam3r:
Yes, the error is gone.

Issue #236, opened 2019-07-12
Working with replaced request body stream


Donnatello:
In Asp Net Core `HttpRequest.Body` has public setter and it makes it possible to replace original stream to another. The standard way to do this is to use the extension methods for `HttpRequest` `EnableBuffering` or `EnableRewind`.
Except for #234, in this situation the position in the stream may be different from 0 and reading the content is not correct now.
In this case should:
1. Leave opened stream after reading (for another middleware)
2. Set stream position to 0 (if it possible)

sungam3r:
Resolved

PR #237, opened 2019-07-12
Seek and leaveOpen for request stream (#236)

Donnatello:
`StreamReader` ctor called with default parameters except leaveOpen
See: https://github.com/dotnet/corefx/blob/master/src/Common/src/CoreLib/System/IO/StreamReader.cs

Fixes: #236


sungam3r:
Сlosed because the necessary changes are already in develop branch. Feel free to reopen if necessary.

PR #239, opened 2019-07-18
New release?

benmccallum:
Are we in a position to do a proper version release of stuff currently in alpha? cc: @joemcbride


benmccallum:
I know that at least my feat 37a7658, and probably a few others haven't made it to an alph arelease though as it seems the last alpha release was back on June 5.

joemcbride:
You can use the myget feed to get more alpha builds.

https://www.myget.org/F/graphql-dotnet/api/v3/index.json

benmccallum:
Thanks @joemcbride, I'd be keen to swap over to alpha releases but we're using this in pretty heavily in prod so it makes me a bit nervous. 😝

joemcbride:
@benmccallum What's currently in the develop branch points to the alpha build of the core GraphQL project.  Your feature was added after that update.  So it will only be alpha builds of this project as well until the core project is no longer in alpha.

sungam3r:
> but we're using this in pretty heavily in prod so it makes me a bit nervous

@benmccallum We are in the same situation. We actively use graphql in prod. I’m already used to occasionally manually download packages from myget and place them in a private corporate package repository. At first it seemed very burdensome, but then I got used to it. Moreover, now I often make changes to this package and underlying graphql-dotnet package. The fact that it is in alpha makes these edits easier, since there is no need to worry about backward compatibility (although it’s still worth a bit). The main drawback of the fact that the package has been in the alpha version for so long is that for many developers all the achievements made are still inaccessible. Many companies may have strict rules allowing the use of only release packages. And developers over and over again encounter those errors that have already been fixed in the alpha version. This is frustrating.

In any case, graphql-dotnet release should be made first.

sungam3r:
> You can use the myget feed to get more alpha builds.

I just found that only the GraphQL package automatically gets there. 
image

And the latest server packages there are only for August:
image

@joemcbride , @pekkah  What do I need to do to publish the latest server packages there? I do not yet have experience with AppVeyor.

pekkah:
Server packages are also pushed to my get.

sungam3r:
But I only see version alpha0020.  Could you check please that the packages are really pushed?

hmvs:
It would be great to see the release.

benmccallum:
I think it'd be good to try and get a new major release out in a couple of weeks that puts support for .net core 3.0 into master. What do you think @sungam3r? Do we need Pekka to help us release to NuGet proper?

I can whip up some changes to the readme to let folks know of 3.0 and what's required. (Basically point to the sample).

sungam3r:
Do you remember that graphql-dotnet is still in 3.0.0-preview?

benmccallum:
Ah, so because you're pointing `develop` at 3.0.x-preview of graphql-dotnet we're stuck. Hmm, that's a shame.

benmccallum:
I'm going to close this. It's just noise. We'll create a develop > master PR if the main project ever does a 3.0 proper release.

sungam3r:
Agree.

PR #241, opened 2019-07-19
Support for batches

sungam3r:
Fixes graphql-dotnet/graphql-dotnet#1128


sungam3r:
Backward compatibility is maintained. It seems to be quite simple. @joemcbride Is there any specification violation?

sungam3r:
https://blog.apollographql.com/query-batching-in-apollo-63acfd859862

sungam3r:
@joemcbride I find it somewhat burdensome to require a `graphql-batch` (or any other) header from the client. Probably worth checking for an array - `[`?

sungam3r:
Relates to graphql-dotnet/graphql-client#68

joemcbride:
> Probably worth checking for an array - [?

That is probably the most sure way to check.  It is probably a little less performant since you would deserialize to string first, but that is probably acceptable.

sungam3r:
> It is probably a little less performant since you would deserialize to string first

No premature string allocation, `StreamReader.Peek()` to the rescue. Also no more allocating `JsonSerializer` on each request.

Issue #251, opened 2019-08-20
ConfigureAwait seems to be applied inconsistently


gbiellem:
I noticed the that ConfigureAwait seems to be inconsistently applied throughout this codebase.
Since fixing this would touch a lot of files  I'd be curious if you'd like one big PR or several small ones

sungam3r:
Can you explain what type of inconsistecy are you talking about?

gbiellem:
@sungam3r Sure. 

For all await calls in  libraries it's generally best practice is  to configure the `await` call by adding `ConfigureAwait(false);`

For ref:  https://markheath.net/post/async-antipatterns and https://blog.stephencleary.com/2012/02/async-and-await.html

I did a quick search over the code base on GraphQL.Net  and GraphQL.Net Server

For GraphQL.net there are 202 `await` statements,  28 of which have been decorated with `ConfigureAwait(false);`

For GraphQL.Net Server there are 117 `await` statements and 5 of which have been decorated with `ConfigureAwait(false);`

Since undecorated awaits will default to `ConfigureAwait(true)` it's inconsistent.

sungam3r:
I didn’t think there were so many of them. Is it without tests? In the tests there is little sense in changing.You are welcome to make a PR.

gbiellem:
FYI... @SimonCropp  has already done the same in GraphQL.Net

SimonCropp:
here is the one in graphql .net  https://github.com/graphql-dotnet/graphql-dotnet/pull/1263

sungam3r:
Closed via #253

PR #253, opened 2019-08-21
Add ConfigureAwait(false) consistently

gbiellem:
As discussed in #251,   This just adds ConfigureAwait(false) to all awaits.    I didn't bother with changing this in any of the test projects

### DIFF
--- a/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
+++ b/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
@@ -49,7 +49,7 @@ namespace GraphQL.Server.Transports.AspNetCore
         {
             if (context.WebSockets.IsWebSocketRequest || !context.Request.Path.StartsWithSegments(_path))
             {
-                await _next(context);
+                await _next(context).ConfigureAwait(false);
                 return;
             }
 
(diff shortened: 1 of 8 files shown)


src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -211,7 +211,7 @@ private static async Task<string> ReadAsStringAsync(Stream s)
             // This leads to the inability to use the stream further by other consumers/middlewares of the request processing
             // pipeline. In fact, it is absolutely not dangerous not to dispose StreamReader as it does not perform any useful
             // work except for the disposing inner stream.
-            return await new StreamReader(s).ReadToEndAsync();
+            return await new StreamReader(s).ReadToEndAsync().ConfigureAwait(false); ;
```
SimonCropp: unrelated to the PR, but should this StreamReader be disposed of?

SimonCropp: this can be a direct return to avoid the async state machine

```
        private static Task<string> ReadAsStringAsync(Stream s)
        {
            // Do not explicitly or implicitly (via using, etc.) call dispose because StreamReader will dispose inner stream.
            // This leads to the inability to use the stream further by other consumers/middlewares of the request processing
            // pipeline. In fact, it is absolutely not dangerous not to dispose StreamReader as it does not perform any useful
            // work except for the disposing inner stream.
            return new StreamReader(s).ReadToEndAsync();
        }
```

SimonCropp: i will do this in a new PR after this is merged

src/Transports.Subscriptions.WebSockets/WebSocketConnection.cs
```diff
@@ -18,9 +18,9 @@ public class WebSocketConnection
 
         public async Task Connect()
         {
-            await _server.OnConnect();
-            await _server.OnDisconnect();
-            await _transport.CloseAsync();
+            await _server.OnConnect().ConfigureAwait(false); ;
+            await _server.OnDisconnect().ConfigureAwait(false); ;
```
sungam3r: Please remove duplicated semicolons here and there.

SimonCropp: @sungam3r done

PR #265, opened 2019-09-30
Made the Deserialize method asynchronous

MikeyFriedChicken:
This prevents the InvalidOperationException which disallow async operations.

### DIFF
--- a/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
+++ b/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
@@ -9,6 +9,9 @@ using Newtonsoft.Json.Linq;
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Threading.Tasks;
 
@@ -65,13 +68,17 @@ namespace GraphQL.Server.Transports.AspNetCore
                 switch (mediaTypeHeader.MediaType)
                 {
                     case JsonContentType:
-                        if (!Deserialize(httpRequest.Body, out gqlRequest, out gqlBatchRequest))
+                        var result = await Deserialize(httpRequest.Body).ConfigureAwait(false);
+
+                        if (!result.Success)
                         {
                             await WriteBadRequestResponseAsync(context, writer, "Body text could not be parsed. Body text should start with '{' for normal graphql query or with '[' for batched query.").ConfigureAwait(false);
                             return;
                         }
-                        break;
 
+                        gqlRequest = result.Request;
+                        gqlBatchRequest = result.Requests;
+                        break;
                     case GraphQLContentType:
                         gqlRequest.Query = await ReadAsStringAsync(httpRequest.Body).ConfigureAwait(false);
                         break;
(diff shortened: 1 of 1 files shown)


MikeyFriedChicken:
Error looks like this:

image

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -159,40 +166,53 @@ private Task WriteBadRequestResponseAsync(HttpContext context, IDocumentWriter w
             return writer.WriteAsync(context.Response.Body, result);
         }
 
-        private Task WriteResponseAsync<TResult>(HttpContext context, IDocumentWriter writer, TResult result)
+        private async Task WriteResponseAsync<TResult>(HttpContext context, IDocumentWriter writer, TResult result)
```
sungam3r: Why change the method signature if you can simply return the task?

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -159,40 +166,53 @@ private Task WriteBadRequestResponseAsync(HttpContext context, IDocumentWriter w
...
             // work except for the disposing inner stream.
             var reader = new StreamReader(stream);
-            using (var jsonReader = new JsonTextReader(reader) { CloseInput = false })
+
+            // There is no async peek, so need to check the type of jToken returned
+            // If anything other than { or ] or some other error then return false
+            try
             {
-                switch (reader.Peek())
+                using (var jsonReader = new JsonTextReader(reader) {CloseInput = false})
```
sungam3r: Please return spaces

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -159,40 +166,53 @@ private Task WriteBadRequestResponseAsync(HttpContext context, IDocumentWriter w
...
-                    default:
-                        return false; // fast return with BadRequest without reading request stream
+                        default:
+                            return (false, null, null, null);
+                    }
                 }
             }
+            catch (Exception exception)
+            {
+                return (false, null, null, exception);
```
sungam3r: You are swallowing an exception. I don’t see that it is used somehow. The current implementation allows the exception to flow up the call stack and be processed by the corresponding http request middleware. I don't think swallowing an exception is the right way.

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -159,40 +166,53 @@ private Task WriteBadRequestResponseAsync(HttpContext context, IDocumentWriter w
...
+            // If anything other than { or ] or some other error then return false
+            try
             {
-                switch (reader.Peek())
+                using (var jsonReader = new JsonTextReader(reader) {CloseInput = false})
                 {
-                    case '{':
-                        single = _serializer.Deserialize<GraphQLRequest>(jsonReader);
-                        return true;
+                    var jToken = await JToken.LoadAsync(jsonReader).ConfigureAwait(false);
```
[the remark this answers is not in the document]
sungam3r: This is my main concern.

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
MikeyFriedChicken: With the addition of Async/Await, the out parameters were not allowed.

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
MikeyFriedChicken: I agree with you here.  It was because I could not use the Peek() method asynchronously so relied on the exception instead.   However, as you said all exceptions will get consumed.  Let's abandon this request, and will look at the code again.

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
sungam3r: Sorry but I don't understand. What out parameters?

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
sungam3r: Will you close this PR?

MikeyFriedChicken: Yep will do.  By the way, just spent some time going through AspNetCore source code to see how the framework handles this as I was puzzled why this error was not coming up all the time.  Throughout the code it gets around this by taking the incoming HttpRequestStream and asynchronously reading it into FileBufferringReadStream which then gets used with the json deserialiser.

So if we apply the same pattern the code change to fix this will be quite minimal.

MikeyFriedChicken:
Closing this as a better fix will be to apply the same pattern used in asp.net core source code, that is to create a FileBufferringReadStream from the HttpRequestStream and asynchronously read first.

PR #271, opened 2019-10-04
Synchronous operations are disallowed on ASP.NET Core 3

edsondewes:
To use the lib on an ASPNET Core 3 project, you must manually allow syncronous operations.

```csharp
// Startup.ConfigureServices
services.Configure<IISServerOptions>(options =>
{
  options.AllowSynchronousIO = true;
});
```

This pull request is to make it work without changing the default config.
"Deserialize" and "Peek" were making syncronous reads on the Body stream, causing exceptions.

I dont know if is the best way to do this, but I couldn't find better async methods in Newtonsoft.Json.

Related to: graphql-dotnet/graphql-dotnet#1325


sungam3r:
Duplicate #265 - here you can read about the disadvantages of such a solution.

edsondewes:
Sorry, I didn't see the other pull request.

sungam3r:
It was closed, no problems.

PR #277, opened 2019-10-16
Fix Synchronous IO

ganhammar:
When targeting latest stable version of GraphQL.Server in a .Net Core 3.0 project you will get an error regarding Synchronous IO, this will resolve that issue.

### DIFF
--- a/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
+++ b/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
@@ -62,7 +62,7 @@ namespace GraphQL.Server.Transports.AspNetCore
                 switch (mediaTypeHeader.MediaType)
                 {
                     case JsonContentType:
-                        gqlRequest = Deserialize<GraphQLRequest>(httpRequest.Body);
+                        gqlRequest = await Deserialize<GraphQLRequest>(httpRequest.Body);
                         break;
                     case GraphQLContentType:
                         gqlRequest.Query = await ReadAsStringAsync(httpRequest.Body);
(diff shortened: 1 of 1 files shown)


sungam3r:
Please see #271 and related issues

ganhammar:
Okay, so you're saying that you're aware of it and it will be address properly in future version? Just for my understanding, will close the pr then.

sungam3r:
Yes, I am aware of it. System.Text.Json will most likely be used.

ganhammar:
I switched out newtonsoft for system.text.json in the commit below, but for the develop branch, switching to netcoreapp3.0 as well. Something like that?
https://github.com/ganhammar/server/commit/fc74f104fefbb54fce25012d628cd3bbfc734fa5

sungam3r:
@ganhammar  Yes, something like that except for inefficient allocation of extra strings in the heap.
The essence of the transition to a new serializer is that 1) **nowhere** to allocate strings 2) **everywhere** work asynchronously. I left my comments there. I also think we will be forced (at least for a while) to target two platforms - both netstandard2.0 and netstandard2.1.

ganhammar:
@sungam3r thanks for the comments! If I look at the things you're mentioning, would it be relevant with a pr?
Is the plan to address this in the master branch as well or will it be with next release?

sungam3r:
Current branch is `develop`. First, the graphql-dotnet project should go to the master - 3.0.0 release. This now is the main blocker. See graphql-dotnet/graphql-dotnet#1321. Also if we change the serializer, then we need to target 2 platforms.

ganhammar:
Thanks for clarifying @sungam3r!

I made two more commits where I addressed your comments. I didn't figure out a good way to do the equivalent of Peek async so I consumed the whole stream and then looked at the first character, there is probably better ways to handle that. I also changed the target platforms, I kept the test and sample projects as netcore3.0.

https://github.com/ganhammar/server/commit/86428c8bcf360e6fc32ffee2b1bad65d647acae0
https://github.com/ganhammar/server/commit/02d1bfc015ce9b550ef45bcaa9228789c698eb7b

benmccallum:
Will be addressed in PR [here](https://github.com/graphql-dotnet/server/pull/299).

Issue #276, opened 2019-10-15
Support for ASP.NET Core 3.0 ?


EdgarMaucourant:
Hi Guys,

Is there a plan to support  Dotnet core 3.0 ?
As Microsoft is terminating the support of .net core 2.2 in December we are a bit in a hurry to migrate our applications to .netcore 3.0.

Best regards
Edgar

sungam3r:
I am sure that this is inevitable. The migration process should not be difficult. Maybe a little later. Most likely, another target will appear in the package, there will be no separate package.

benmccallum:
I'm trying to update, but the issue I'm currently having is that `AddAuthorization` has been renamed to `AddAuthorizationCore`, so the call in the `AddGraphQLAuthorization` is failing on method not found. 

https://github.com/aspnet/Announcements/issues/386

benmccallum:
I might be able to give the multi-targeting issue a go today, since it's blocking me.

EdgarMaucourant:
Hi benmccallum,

I also did try, and fixed the AddAuthorizationCore (by editing the source directly), but then I run also on conflict in the GraphQL code, so I gave up for the moment, and will try later. I let you know if I manage to get this working, but as I'm on tight deadline for the project, it will be not before 10 of November in think.

Edgar

benmccallum:
Do you have a branch you could push for this repo?

You had an issue with the main GraphQL package too? Hmm... what I'm unsure about is whether we can keep Microsoft.Extensions.Xyz 2.x versions, or they need to be upgraded to 3.0.0, which probably requires netcoreapp3.0, and then it's a big issue.

sungam3r:
Multitargeting with Conditional csproj switches all the way...

benmccallum:
And by all the way you mean in this repo and graphql? Hmm, I can give it a go I suppose :)

benmccallum:
I think the biggest pain point is that some of upgraded to 2.2 but its only 2.1 that is LTS, so we're like... up sh*t creek unless we wanna back down to 2.1 which is just nasty.

benmccallum:
@sungam3r , I'm going to create a branch for this project support 3.0.0, should I do it off develop or master?

sungam3r:
> And by all the way you mean in this repo and graphql? Hmm, I can give it a go I suppose :)

netstandard2.0 and netstandard2.1 in graphql-dotnet

sungam3r:
[...]

PR #280, opened 2019-10-18
Support ASP.NET Core 3.0

benmccallum:
Addresses #276


benmccallum:
I think this should do it, but @EdgarMaucourant did mention he had issues in GraphQL proper over on the issue, so potentially we need to do a similar thing over in that project.

My biggest unknown about all this is can an ASP.NET Core 3.0 app be referencing projects that target `netstandard2.x` and reference packages like `Microsoft.Extensions.Xyz` at v2.x.x?

I've got no idea what would happen. I assume the ASP.NET Core 3.0 app will get the shared framework, which potentially includes v3.0.0 of those, so it would have to be able to fwd and be compatible. This really isn't my area of expertise.

benmccallum:
Few more things to sort out, but getting there slowly...

benmccallum:
So I've got all tests passing. Outstanding is:
1. WebSocketConnectionFacts tests are failing under netcoreapp3.0, this is above my head, perhaps @pekkah would know a solution :P Getting a `System.InvalidOperationException : Incomplete handshake, status code: 404` which could be a breaking change in 3.0 *shrugs*
2. Cake build file updates, would love if someone could jump in here for that too. Potentially that involves AppVeyor configuration as well depending on where the SDK/s for builds are configured. Haven't really looked into it yet.
3. Find out if this even works on GraphQL proper.

Directory.Build.props
```diff
@@ -11,19 +11,30 @@
     <Company>graphql-dotnet</Company>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <IsNetCore30OnwardsTarget>$(TargetFramework.Equals('netcoreapp3.0'))</IsNetCore30OnwardsTarget>
+  </PropertyGroup>
+  
   <PropertyGroup Label="Package dependency versions">
-    <MicrosoftAspNetCoreVersion>2.2.0</MicrosoftAspNetCoreVersion>
+    <MicrosoftAspNetCoreVersion>2.2.0</MicrosoftAspNetCoreVersion>
```
sungam3r: Please check for such additional whitespaces all over PR to reduce noise.

samples/Samples.Server30/GraphQLHttpMiddlewareWithLogs.cs
```diff
@@ -0,0 +1,37 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+using System;
+using System.Threading.Tasks;
+
+namespace GraphQL.Samples.Server
+{
+    // Example of a custom GraphQL Middleware which logs errors to Microsoft.Extensions.Logging API
+    public class GraphQLHttpMiddlewareWithLogs<TSchema> : GraphQLHttpMiddleware<TSchema>
```
sungam3r: No need a separate file, just add existing as link.

samples/Samples.Server30/Samples.Server30.csproj
```diff
@@ -0,0 +1,33 @@
...
+
+  <ItemGroup>
+    <PackageReference Include="GraphQL" Version="$(GraphQLVersion)" />
+    <PackageReference Include="Serilog" Version="$(SerilogVersion)" />
+    <PackageReference Include="Serilog.AspNetCore" Version="$(SerilogAspNetCoreVersion)" />
+    <PackageReference Include="Serilog.Sinks.Console" Version="$(SerilogSinksConsoleVersion)" />
+  </ItemGroup>
+  
+  <ItemGroup>
+    <DotNetCliToolReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Tools" Version="2.0.0" />
```
sungam3r: This can be removed from both current and new sample project.

samples/Samples.Server30/Startup.cs
```diff
@@ -0,0 +1,90 @@
...
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Server.Kestrel.Core;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using System;
+using System.Collections.Generic;
+
+namespace GraphQL.Samples.Server
```
sungam3r: I would prefer not to copy this file but to manage one. This will be a good example of source level backward compatibility.

samples/Samples.Server30/package-lock.json
```diff
@@ -0,0 +1,95 @@
+{
+  "name": "samples.server",
```
sungam3r: This file together with package.json file can be removed from both samples. I do not write JS code but I see that this stuff is not needed here. Everything should work without it.

sungam3r: Also all files in wwwroot folder

sungam3r: Apparently these files appeared a long time ago and are not required for work now.

src/Authorization.AspNetCore/Authorization.AspNetCore.csproj
```diff
@@ -10,6 +10,13 @@
 
   <ItemGroup Label="Package References">
     <PackageReference Include="GraphQL" Version="$(GraphQLVersion)" />
+  </ItemGroup>
+
+  <ItemGroup Label="Package references for .NET Core" Condition="'$(IsNetCore30OnwardsTarget)' == 'True'">
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
```
sungam3r: It looks strange to refer from a library to a meta-package. It seems to me that the documentation talked about using this method for end applications, not libraries.

src/Authorization.AspNetCore/GraphQLBuilderExtensions.cs
```diff
@@ -18,7 +18,11 @@ public static IGraphQLBuilder AddGraphQLAuthorization(this IGraphQLBuilder build
             builder.Services
                 .AddHttpContextAccessor()
                 .AddTransient<IValidationRule, AuthorizationValidationRule>()
+#if NETCOREAPP3_0
```
sungam3r: Whole method impl can be changed to
```c#
 public static IGraphQLBuilder AddGraphQLAuthorization(this IGraphQLBuilder builder) => builder.AddGraphQLAuthorization(opt => { });
```

src/Core/Core.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.0;netstandard2.0</TargetFrameworks>
```
sungam3r: This is not required. You do not need to add a second platform if there is no specific api for it.

src/Transports.AspNetCore/Transports.AspNetCore.csproj
```diff
@@ -10,10 +10,17 @@
 
   <ItemGroup>
     <PackageReference Include="GraphQL" Version="$(GraphQLVersion)" />
-    <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="$(MicrosoftAspNetCoreHostingAbstractionsVersion)" />
     <PackageReference Include="Microsoft.Extensions.Options" Version="$(MicrosoftExtensionsOptionsVersion)" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(IsNetCore30OnwardsTarget)' == 'True'">
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
```
sungam3r: The same question.

src/Transports.Subscriptions.WebSockets/Transports.Subscriptions.WebSockets.csproj
```diff
@@ -17,6 +17,15 @@
     <PackageReference Include="Newtonsoft.Json" Version="$(NewtonsoftJsonVersion)" />
     <PackageReference Include="System.Reactive" Version="$(SystemReactiveVersion)" />
   </ItemGroup>
+  
+  <ItemGroup Condition="'$(IsNetCore30OnwardsTarget)' == 'True'">
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
```
sungam3r: one more

src/Ui.GraphiQL/Ui.GraphiQL.csproj
```diff
@@ -12,8 +12,12 @@
 	  <EmbeddedResource Include="Internal\graphiql.cshtml" />
 	</ItemGroup>
 
-	<ItemGroup>
-		<PackageReference Include="Microsoft.AspNetCore.Http.Abstractions" Version="$(MicrosoftAspNetCoreHttpAbstractionsVersion)" />
-	</ItemGroup>
+  <ItemGroup Condition="'$(IsNetCore30OnwardsTarget)' == 'True'">
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
```
sungam3r: Documentation says that `Microsoft.AspNetCore.App` should be removed since ASP.NET Core now is a share framework - https://docs.microsoft.com/en-us/aspnet/core/migration/22-to-30?tabs=visual-studio&view=aspnetcore-2.2#remove-obsolete-package-references

sungam3r: But it seems that it is only for Microsoft.NET.Sdk.Web SDK.

tests/Samples.Server.Tests/BaseTest.cs
```diff
@@ -1,20 +1,29 @@
...
 {
-    public abstract class BaseTest : IDisposable
+    public abstract class BaseTest : IClassFixture<WebApplicationFactory<Startup>>
     {
-        protected BaseTest()
+        private readonly WebApplicationFactory<Startup> _factory;
+
+        public BaseTest(WebApplicationFactory<Startup> factory)
         {
-            Server = new TestServer(Program.CreateWebHostBuilder(Array.Empty<string>()));
```
sungam3r: Why not to stay `TestServer`? It is available in ASP.NET Core 3.0 and its api seems cleaner than `WebApplicationFactory<Startup>`. Although I suspect that something has changed there.

src/Authorization.AspNetCore/Authorization.AspNetCore.csproj
benmccallum: See https://github.com/aspnet/AspNetCore/issues/6040 and https://github.com/aspnet/AspNetCore/issues/3756. It's just the way it is now for 3.0. The dependencies that were previously exposed as separate packages are only available in the shared framework now so you don't have a choice.

samples/Samples.Server30/Startup.cs
benmccallum: Do you want me to Include Existing as Link for Program.cs and Startup.cs? 

There's quite a few changes as some core types change. I can see the need to not manage two files, but I can also see that we don't wanna scare people off with compiler conditionals. There's a case to be made for both :P Let me know.

src/Transports.AspNetCore/Transports.AspNetCore.csproj
benmccallum: Same answer - we need stuff that was previously in `Microsoft.AspNetCore.Hosting.Abstractions` and is now only available in the shared framework.

src/Transports.Subscriptions.WebSockets/Transports.Subscriptions.WebSockets.csproj
benmccallum: Same answer - we need stuff that was previously in Microsoft.AspNetCore.Hosting.Abstractions and is now only available in the shared framework.

src/Ui.GraphiQL/Ui.GraphiQL.csproj
benmccallum: Yea if you're targeting `netcoreapp3.0` and using the `Microsoft.NET.Sdk.Web` SDK then the framework reference is implicitly added for you. Since this is just using the "Microsot.NET.Sdk" SDK (for libraries) then we need to explicitly add it, otherwise, it's the same answer:
> we need stuff that was previously in Microsoft.AspNetCore.Hosting.Abstractions and is now only available in the shared framework.

src/Core/Core.csproj
benmccallum: Yep, removing this seemed to be OK; build and tests. Since we don't need to toggle with compiler flags for this framework or pull in different references.

tests/Samples.Server.Tests/BaseTest.cs
benmccallum: Yea I tried staying with `TestServer` but the issue was that the call pattern changed. It still takes a `WebHostBuilder` even though in 3.0 your `Program` now creates a generic `HostBuilder`. I couldn't figure out the call pattern, so swapped over to the well-documented integration test approach which worked.

However, swapping over to the factory approach didn't work for the WebSockets test (see it's ctor) as it relies on the Startup class's assembly having a `Program` type with the `CreateHostBuilder` function (or `CreateWebHostBuilder` in 2.x) but those test/s are using a custom `Startup` straight in the test project.

I can swap this one back, but it's actually tidier like this (the new call pattern is quite long) and more standard/documented. I'd prefer to align the websockets one to this but obviously we can't. Also haven't got the WebSockets tests working yet, so...

Let me know.

benmccallum:
Hey @sungam3r, ready for another review. 
* I've resolved all the ones that I agreed with and changed.
* I've commented on the references ones that have to be like that - linked some sources. Resolve these when you can.
* Outstanding is: 
** Linked and compiler toggled Program and Startup? Or keep separate?
** Go back to TestServer or leave as standard/documented approach?
** Someone to help me fix the WebSockets tests.

src/Authorization.AspNetCore/Authorization.AspNetCore.csproj
sungam3r: Wow, I did not notice that here `FrameworkReference ` instead of `PackageReference `.
Well, the dependencies issue is the first thing that bothered me. Now I understand that my concerns about shifting from fine grained netstandard2.0-targeted packages to one shared netcoreapp3-targeted framework were well founded. Now I understand what you meant when you said that netstandard is dead. Let's just say it is dead only for applications and libraries that are built on top of ASP.NET Core stack. Now I have gotten more work to migrate our applications to ASP.NET Core 3.0 - we will have to use **Trimmed** Self-Contained Deployment. Thanks for the useful information!

benmccallum: Yea trimming will be the way to go if doing self-contained, that's for sure. 

I think it's probably going to be dead for non-aspnetcore stacks too, as .NET 5 will be .NET Core and only supported by `netcoreapp5.0` I'd imagine. I'm talking dead in a long way off though. As long as we support netfx it'll be useful, but come .NET 6 and 7, etc, I imagine the idea really is that netcoreapp is xplat enough that it can run everywhere. Whether that's the case though is yet to be seen. Potentially IoT and some targets will mean that netstandard is still useful over there.

samples/Samples.Server30/Startup.cs
sungam3r: > Do you want me to Include Existing as Link for Program.cs and Startup.cs?

Yes. 

> There's quite a few changes as some core types change. I can see the need to not manage two files, but I can also see that we don't wanna scare people off with compiler conditionals. 

Agree. If the code is very different for the two platforms, then using the preprocessor makes it an unreadable stuff. But if you only need a couple of such switches, then using one file makes sense - the reader focuses on the switch point. In the case of different examples, he does not even bother to see how they differ. I am for the thoughtful reader to understand what the difference is right away, and not be forced to conduct a manual comparison.

tests/Samples.Server.Tests/BaseTest.cs
sungam3r: Does [this](https://github.com/aspnet/AspNetCore/issues/7754#issuecomment-543813345) not work?

sungam3r:
1. Linked and compiler toggled Program and Startup? Or keep separate?
2. Go back to TestServer or leave as standard/documented approach?
3. Someone to help me fix the WebSockets tests.
4. Someone to help me fix the build on AppVeyor for 3.0 target

____
1. I am for linked.
2. I am for TestServer. And isn't it documented?
3. I can’t do it yet.
4. I guessed that it would have to be done.  I am not familiar with appveyor but I think that you should start with changing 2017 to 2019 and 2.2.104 to 3.0.100 in [config](https://github.com/graphql-dotnet/server/blob/develop/appveyor.yml)

src/Authorization.AspNetCore/Authorization.AspNetCore.csproj
sungam3r: Let's not be upset ahead of time.

tests/Samples.Server.Tests/BaseTest.cs
benmccallum: Yea it does, it's just more verbose and not documented. Want me to use that?

sungam3r: Let's go this way. I would not say that the code is complicated. He is very straightforward.

benmccallum: Another benefit of the factory is that it's the same code for both 2.x and 3.0, I'll have to dupe the blocks a bit to go back to the old way

sungam3r: Well... I do not know. The testserver option seems simpler. Nevertheless the main requirement is that this just should work.

benmccallum: Haha, I swapped over and it doesn't work. The response from the client is a 404... Honestly not keen on spending any more time trying to figure out why, though it could be the same issue affecting the websockets tests... hmmm.

tests/Samples.Server.Tests/BaseTest.cs
```diff
@@ -39,10 +58,20 @@ public virtual void Dispose()
         {
             Client.Dispose();
             Server.Dispose();
+
+#if NETCOREAPP2_2
+#else
```
sungam3r: can be changed to one line #if NETCOREAPP3_0

tests/Samples.Server.Tests/BaseTest.cs
benmccallum: Nevermind, got it working. Alright, will consolidate the startup and program and then we'll be there almost.

tests/Samples.Server.Tests/BaseTest.cs
benmccallum: Yea I've done this so that we can avoid changing it in the future, e.g. when 4_0 comes out. Assuming we can use the same code we shouldn't have to do anything other than deal with package versions in csproj files.

benmccallum: Guess I can do `!NETCOREAPP2_2` as we prob won't add any other older targets. Done.

samples/Samples.Server30/Startup.cs
benmccallum: Done!

benmccallum:
Almost there. 

I up'd the appveyor.yml versions and now it all builds, but it's failing on the Tests step. It will eventually fail on the WebSockets ones until I can fix them, but it's strange that it's failing earlier. Seems like it's trying to run the tests for the 3.0 sample project against 2.2... Tried commenting out the specification of the framework in the cake file but no luck. I feel like we might need to bump cake's version somewhere, but I'm not sure where that would be specified - build.ps1 I guess but running that on my machine works and gets all the way to the websockets tests...

Need to move on right now, but maybe someone will jump in.

benmccallum:
Oh I see, version is in /tools/packages.config... still, it works on my machine so... I can bump I guess in case it's cached in the AppVeyor build pipeline.

pekkah:
Cake version shouldn't matter. I would check the framework param in build. Also you can run the build locally to verify it works. Could also switch to using Cake.Tool instead of the manual bootstrapper scripts.

pekkah:
Did you change this already in appveyor.yaml
```
 - ps: choco install dotnetcore-sdk --no-progress --confirm --version 2.2.104
```

benmccallum:
Yea I've bumped that to 3.0.100, and dropped out all the FE stuff re: node, etc. Also bumped the VS version.

See [here](https://github.com/graphql-dotnet/server/pull/280/files?file-filters%5B%5D=.cake&file-filters%5B%5D=.config&file-filters%5B%5D=.cs&file-filters%5B%5D=.csproj&file-filters%5B%5D=.css&file-filters%5B%5D=.html&file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.ps1&file-filters%5B%5D=.yml#diff-180360612c6b8c4ed830919bbb4dd459L20).

Is there a cache in AppVeyor that needs clearing? It's strange that it works on my machine when I run build.ps1, but I guess that could be because I've got something different installed. I'm on 3.0.100 too,

benmccallum:
It's gotta be the fact it doesn't understand that the 3.0 only samples project should only be run with netcoreapp3.0... not sure how we control that though. Should be taken care of by the dotnet cli

pekkah:
Check the build task in cake. I remember it passing the framework as net standard 2.0.

pekkah:
Maybe just commenting it out is enough

benmccallum:
Yea sorry, forgot to mention I'd commented that out. Didn't help

pekkah:
Ok. I got that same error in other projects due to NuGet or dotnet sdk version being too old. Happened at restore stage.

sungam3r:
изображение
@benmccallum Not bad, nice cleaning :)

benmccallum:
Found the AppVeyor issue. Fresh eyes. This was the fix. https://github.com/graphql-dotnet/server/pull/280/commits/bdd139b260fe44b94885143346c03684f094b925

I realised that the thing it was complaining about "GraphQL.Samples.Server doesn't support netcoreapp2.0, it's netcoreapp3.0 only" or whatever, was specifically referring to the assembly name in the .csproj, as the project name doesn't include `GraphQL.`. So then I found that I hadn't updated the assembly name for the new project and so that messes up the test run. Kinda weird it worked locally though...

So the last thing here is the WebSockets unit tests still failing. I'm really not across this stuff but I'll have a quick look. @pekkah, sorry to bother, but you seem good at this, mind taking a quick look at it?

benmccallum:
Latest commit gets rid of the 3.0 specific samples server project all together, since I realised all the files were linked anyway...

sungam3r:
> Latest commit gets rid of the 3.0 specific samples server project all together, since I realised all the files were linked anyway...

It is interesting to observe such an evolution. I was hoping that everything would come down to one sample project with multitargeting. It seems to me that this is better than individual examples. It’s easier to focus and understand the difference between the platforms.

benmccallum:
Boom, got it. 

Bit painful, but looks like if you call `.Configure(() => {})` on a `IWebHostBuilder` instance it'll take precedence over the `Configure` method in `TStartup` when calling `.UseStartup<TStartup>()`. Middleware wasn't being hooked up, hence the 404 behaviour and the middleware not firing.

What's next @sungam3r, is this ready for prime time? Did a bit more clean up as I went through all this at least :P

sungam3r:
> Bit painful, but looks like if you call .Configure(() => {}) on a IWebHostBuilder instance it'll take precedence over the Configure method in TStartup when calling .UseStartup<TStartup>(). 

It's good that everything ended successfully. I will look again later. I don’t think I will have any new comments.

samples/Samples.Server/Program.cs
```diff
@@ -1,5 +1,6 @@
 using Microsoft.AspNetCore;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Hosting;
```
sungam3r: Maybe take `using Microsoft.AspNetCore.Hosting;` and `using Microsoft.Extensions.Hosting;` under #if / #else too?

samples/Samples.Server/Program.cs
```diff
@@ -34,10 +39,23 @@ public static int Main(string[] args)
...
         }
 
+#if NETCOREAPP2_2
         public static IWebHostBuilder CreateWebHostBuilder(string[] args)
         {
             return WebHost.CreateDefaultBuilder<Startup>(args)
                 .UseSerilog();
         }
+#else
+                public static IHostBuilder CreateHostBuilder(string[] args)
```
sungam3r: indentation

tests/Samples.Server.Tests/BaseTest.cs
```diff
@@ -1,19 +1,38 @@
 ﻿using GraphQL.Samples.Server;
 using GraphQL.Server.Transports.AspNetCore.Common;
-using Microsoft.AspNetCore.TestHost;
 using Newtonsoft.Json;
-using System;
 using System.Net.Http;
 using System.Text;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Hosting;
```
sungam3r: under #if

tests/Samples.Server.Tests/Samples.Server.Tests.csproj
```diff
@@ -1,13 +1,13 @@
...
-    <TargetFramework>netcoreapp2.2</TargetFramework>
+    <TargetFrameworks>netcoreapp3.0;netcoreapp2.2</TargetFrameworks>
     <Description>End to end tests for Samples.Server project</Description>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.App" />
-    <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="$(MicrosoftAspNetCoreTestHostVersion)" />
+    <PackageReference Include="GraphQL" Version="$(GraphQLVersion)" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="$(MicrosoftAspNetCoreMvcTestingVersion)" />
```
sungam3r: `Microsoft.AspNetCore.TestHost` should be enough. There are no any MVC things in this repo. Also no need in `MicrosoftAspNetCoreMvcTestingVersion` in Directory.Build.props.

tests/Transports.Subscriptions.WebSockets.Tests/Transports.Subscriptions.WebSockets.Tests.csproj
```diff
@@ -16,16 +16,23 @@
   <ItemGroup>
     <PackageReference Include="Castle.Core" Version="$(CastleCoreVersion)" />
     <PackageReference Include="GraphQL" Version="$(GraphQLVersion)" />
-    <PackageReference Include="Microsoft.AspNetCore" Version="$(MicrosoftAspNetCoreVersion)" />
-    <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="$(MicrosoftAspNetCoreTestHostVersion)" />
-    <PackageReference Include="Microsoft.AspNetCore.WebSockets" Version="$(MicrosoftAspNetCoreWebSocketsVersion)" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="$(MicrosoftAspNetCoreMvcTestingVersion)" />
```
sungam3r: TestHost instead

tests/Transports.Subscriptions.WebSockets.Tests/WebSocketsConnectionFacts.cs
```diff
@@ -50,5 +76,13 @@ public async Task should_not_accept_websocket_with_wrong_protocol()
             /* Then */
             Assert.Equal(WebSocketCloseStatus.ProtocolError, received.CloseStatus);
         }
+
+        public void Dispose()
+        {
+            _server.Dispose();
+#if !(NETFRAMEWORK || NETCOREAPP2_2)
+        _host.Dispose();
```
sungam3r: indentation

tests/Samples.Server.Tests/BaseTest.cs
```diff
@@ -39,10 +58,18 @@ public virtual void Dispose()
...
+            Host.Dispose();
+#endif
         }
 
         protected TestServer Server { get; }
 
         protected HttpClient Client { get; }
+
+#if !NETCOREAPP2_2
+        protected IHost Host { get; }
```
sungam3r: Protected? Any usages?

samples/Samples.Server/Program.cs
benmccallum: Yea I wasn't sure what was best practice with that. Will #if/#else them to make it clearer.

samples/Samples.Server/Program.cs
benmccallum: Interesting Ctrl, K + F, only works on current target code. Fixed.

tests/Samples.Server.Tests/BaseTest.cs
benmccallum: Done.

tests/Samples.Server.Tests/Samples.Server.Tests.csproj
benmccallum: Done.

tests/Transports.Subscriptions.WebSockets.Tests/Transports.Subscriptions.WebSockets.Tests.csproj
benmccallum: Done.

tests/Transports.Subscriptions.WebSockets.Tests/WebSocketsConnectionFacts.cs
benmccallum: Fixed.

tests/Samples.Server.Tests/BaseTest.cs
benmccallum: Nope, neither for the two above. All private now.

benmccallum:
@sungam3r , all feedback addressed. LGTM

sungam3r:
Finally!

PR #283, opened 2019-11-05
GraphQLRequestExecutionResult instead of multiple params

sungam3r:
move all params in GraphQLRequestExecutionResult
add Elapsed
bump GraphQL


src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -98,18 +99,19 @@ public async Task InvokeAsync(HttpContext context)
                 userContext = new Dictionary<string, object>(); // in order to allow resolvers to exchange their state through this object
 
             var executer = context.RequestServices.GetRequiredService<IGraphQLExecuter<TSchema>>();
-
+
```
sungam3r: oops, it seems time to add editorconfig

PR #315, opened 2020-02-16
Add .editorconfig aligned with one from graphql-dotnet

sungam3r:


.editorconfig
```diff
@@ -1,9 +1,24 @@
...
+# .NET naming conventions for EditorConfig
+# https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-naming-conventions?view=vs-2019
+
+# Top-most EditorConfig file
 root = true
 
-# Unix-style newlines with a newline ending every file
+# Editor default newlines with a newline ending every file
 [*]
 end_of_line = lf
```
sungam3r: I saw the difference only in this line and did not change it.

PR #318, opened 2020-02-19
Add deserializer benchmarks

benmccallum:


benmccallum:
Woah this is weird. Looks like I really need to rebase ay. Ugh.

PR #321, opened 2020-02-21
fix subscriptions regression on S.T.J

### LINKED ISSUE #320: Subscription transport protocol has wrong casing when using System.Text.Json
It appears that #192 has made a comback with System.Text.Json.
I believe the reason is that DataMember is not supported.

I worked around this by using a custom JsonConverter for serialization.
The better solution is perhaps a naming policy or setting JsonPropertyName

```c#
    public class OperationMessageConverter : JsonConverter<OperationMessage>
    {
        public override OperationMessage Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
        {
            throw new NotImplementedException();
        }

        public override void Write(Utf8JsonWriter writer, OperationMessage message, JsonSerializerOptions options) {
            writer.WriteStartObject();
            writer.WritePropertyName("id");
            writer.WriteStringValue(message.Id);

            writer.WritePropertyName("type");
            writer.WriteStringValue(message.Type);

            writer.WritePropertyName("payload");
            JsonSerializer.Serialize(writer, message.Payload, options);
            
            writer.WriteEndObject();
        }
    }
```

sungam3r:
fixes #320

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -50,7 +50,7 @@
     <CastleCoreVersion>4.4.0</CastleCoreVersion>
     <NSubstituteVersion>3.1.0</NSubstituteVersion>
     <NewtonsoftJsonVersion>12.0.3</NewtonsoftJsonVersion>
-    <GraphQLVersion>3.0.0-preview-1490</GraphQLVersion>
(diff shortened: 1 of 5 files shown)


sungam3r:
@benmccallum we broke subscriptions a bit

PR #323, opened 2020-02-22
Remove ConfigureAwait(false)

sungam3r:
I propose to remove `ConfigureAwait(false)` calls because it is nothing more than noise for ASP.NET Core projects that have no synchronization context. I did not remove `ConfigureAwait(false)` only from Transports.Subscriptions.Abstractions project. With high probability they are not needed there, but it will be part of another PR.

@benmccallum @pekkah @joemcbride Please review.


benmccallum:
Almost wonder if you should just change the targeting to explicitly be netcoreapp2.1 and netcoreapp2.2 rather than netstandard2.0. It's kinda implied by the AspNetCore packagerefs like we've previously discussed, but yea. Maybe just a note in the readme even. I'll leave that up to you.

sungam3r:
Maybe in small steps we will eventually come to this.

PR #326, opened 2020-02-23
Add GraphQL Playground HTTP headers

### LINKED ISSUE #311: Setting HTTP Headers programatically in GraphQL Playground
We want to be able to set a default Authorization HTTP Header programmatically in the GraphQL playground so that when we start the playground, we will be able see something like this:

image

I don't quite see how we can achieve this by using `GraphQLPlaygroundOptions.GraphQLConfig`.
Can someone guide me in the right direction? An example would be appreciated.

**GraphQL for .NET:** 3.0.0-preview-1352
**GraphQL.Server.Ui.Playground**: 3.5.0-alpha0027

sungam3r:
fixes #311

### DIFF
--- a/samples/Samples.Server/Startup.cs
+++ b/samples/Samples.Server/Startup.cs
@@ -77,7 +77,12 @@ namespace GraphQL.Samples.Server
                 {
                     ["editor.theme"] = "light",
                     ["tracing.hideTracingResponse"] = false,
-                }
+                },
+                Headers = new Dictionary<string, object>
+                {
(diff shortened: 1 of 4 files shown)


Askar938:
@sungam3r do you know when this will be released?
I tried to download `GraphQL.Server.Ui.Playground 3.5.0-alpha0046` but that one didn't have it. Also I couldn't find any information about a future release.

sungam3r:
See readme:

изображение

PR #327, opened 2020-02-23
Sync with FieldMiddleware changes in graphql-dotnet repo

sungam3r:
@Shane32 There are few changes, it was enough to register the middleware.


Shane32:
Yup, figured if it was using `DefaultServiceProvider` then it probably wasn't supposed to.

Shane32:
oh, you're still using it.  well, I know nothing about this project

PR #334, opened 2020-02-26
Project reorganization

sungam3r:
* make pubinternal classes public
* move versionprefix and versionsuffix into Directory.Build.props
* flatten namespaces, i.e. remove *.common
* remove unnecessary csproj settings
* README tweaks


src/Core/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -1,10 +1,10 @@
 using System;
+using GraphQL;
 using GraphQL.Instrumentation;
-using GraphQL.Server.Internal;
-using Microsoft.Extensions.DependencyInjection;
+using GraphQL.Server;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 
-namespace GraphQL.Server
+namespace Microsoft.Extensions.DependencyInjection
```
sungam3r: This is namespace of `this` argument.

sungam3r:
@benmccallum @joemcbride @shane32 thoughts?

src/Core/Extensions/ServiceCollectionExtensions.cs
Shane32: Correct me if I'm wrong, but I was under the impression that if a namespace was imported, all extensions in static classes of that namespace are then available.  So with this change you would need to have `using Microsoft.Extensions.DependencyInjection;` in order to use these extensions.  Is that what is desired?  No opinion here, just want to clarify.

Shane32:
I have not used this project, so I don't have much of an opinion.  It looks like you're just doing some housekeeping; if we are releasing 3.0 soon, better before than after.

src/Core/Extensions/ServiceCollectionExtensions.cs
sungam3r: Right. The fact is that in most cases the namespace of the `this` parameter has already been imported. Due to this, IntelliSense discoverability is provided. See https://github.com/graphql-dotnet/server/blob/develop/samples/Samples.Server/Startup.cs#L10

`using Microsoft.Extensions.DependencyInjection` is already here.

BenjaBobs: It's actually possible in VS2019 >16.4 (and iirc also Rider) to do this without having them in the same namespace.
For VS2019:
image

Note that this is not a suggestion to change it back, simply wanted to make sure you had all the information for when making the decision.

sungam3r: OK. Thanks for information, I did not know about this setting.

benmccallum: Didn't know about this setting either. VS has always baffled me when it sometimes finds extensions and sometimes doesn't. I'm guilty of making my namespaces either non-existant or very simple as a result so people find them. Thanks for the tip!

benmccallum: image LOL. I wonder what that means exactly...

BenjaBobs: It means partially enabled. You can press `Alt+A` when the suggestion window is open. But I just have it fully enabled at all times, it's much easier and great for discovery.

README.md
```diff
@@ -75,25 +75,24 @@ public void Configure(IApplicationBuilder app)
     // this is required for websockets support
     app.UseWebSockets();
 
-    // use websocket middleware for ChatSchema at path /graphql
-    app.UseGraphQLWebSockets<ChatSchema>("/graphql");
+    // use websocket middleware for ChatSchema at default url /graphql
```
benmccallum: Internally our team tends to stick with "path" for the url path and "url" for an absolute one, but that's just nitpicking in a doc file. You could make it "url path" if you wanted but eh, up to you.

tests/Transports.Subscriptions.Abstractions.Tests/Transports.Subscriptions.Abstractions.Tests.csproj
```diff
@@ -2,9 +2,6 @@
 
   <PropertyGroup>
     <TargetFrameworks>netcoreapp3.1;netcoreapp3.0;netcoreapp2.2;net471</TargetFrameworks>
-    <IsPackable>false</IsPackable>
-    <AssemblyName>GraphQL.Server.Transports.Subscriptions.Abstractions.Tests</AssemblyName>
-    <RootNamespace>GraphQL.Server.Transports.Subscriptions.Abstractions.Tests</RootNamespace>
```
benmccallum: You'll want to keep the root namespace as otherwise adding files to the project will get added as the project name, which doesn't have `GraphQL.Server` at the start.

tests/Transports.Subscriptions.WebSockets.Tests/Transports.Subscriptions.WebSockets.Tests.csproj
```diff
@@ -2,17 +2,8 @@
 
   <PropertyGroup>
     <TargetFrameworks>netcoreapp3.1;netcoreapp3.0;netcoreapp2.2;net471</TargetFrameworks>
-    <IsPackable>false</IsPackable>
-    <AssemblyName>GraphQL.Server.Transports.WebSockets.Tests</AssemblyName>
-    <RootNamespace>GraphQL.Server.Transports.WebSockets.Tests</RootNamespace>
```
benmccallum: Same here. Suggest you keep `RootNamespace`.

README.md
sungam3r: OK

tests/Transports.Subscriptions.Abstractions.Tests/Transports.Subscriptions.Abstractions.Tests.csproj
sungam3r: OK

src/Transports.Subscriptions.WebSockets/Transports.Subscriptions.WebSockets.csproj
```diff
@@ -2,10 +2,6 @@
   
   <PropertyGroup>
     <TargetFrameworks>netcoreapp3.1;netcoreapp3.0;netstandard2.0</TargetFrameworks>
-    <AssemblyName>GraphQL.Server.Transports.WebSockets</AssemblyName>
```
sungam3r: **Attention!**
@pekkah @joemcbride I noticed a naming inconsistency.

Before:
```
ProjectName = Transports.Subscriptions.WebSockets
AssemblyName = PackageName = GraphQL.Server.Transports.WebSockets
```

After:
```
ProjectName = Transports.Subscriptions.WebSockets
AssemblyName = PackageName = GraphQL.Server.Transports.Subscriptions.WebSockets
```

All other projects consistently add `GraphQL.Server` to the name.
Moreover 
```
<Description>WebSockets transport for subscriptions</Description>
```
clearly says that this is a transport for subscriptions, not some general transport on web sockets. So I propose to change package name to `GraphQL.Server.Transports.Subscriptions.WebSockets`. See https://ci.appveyor.com/project/graphql-dotnet-ci/server/builds/31132815/artifacts

Directory.Build.props
```diff
@@ -4,11 +4,16 @@
...
     <RepositoryUrl>https://github.com/graphql-dotnet/server</RepositoryUrl>
-    <RepositoryType>Git</RepositoryType>
+    <RepositoryType>git</RepositoryType>
     <Copyright>Pekka Heikura</Copyright>
     <Authors>Pekka Heikura</Authors>
     <Product>graphql-dotnet server</Product>
     <Company>graphql-dotnet</Company>
+    <VersionPrefix>3.5.0</VersionPrefix>
+    <VersionSuffix>alpha</VersionSuffix>
+    <AssemblyName>GraphQL.Server.$(MSBuildProjectName)</AssemblyName>
```
sungam3r: This will allow to unify naming and in case of new packages not to make mistakes with naming conventions.

tests/Transports.Subscriptions.WebSockets.Tests/Transports.Subscriptions.WebSockets.Tests.csproj
sungam3r: Done by using naming  template in Directory.Build.props

pekkah:
The original names were not changed due to it being a major breaking change for anyone using them as simple package upgrade would not work for them anymore. If that is now OK then just change them.

sungam3r:
I understand the consequences of changing the name of the package. I'm just asking for advice on whether it’s acceptable to change the name now.

sungam3r:
@Shane32 @BenjaBobs @benmccallum I resolved conflicts and if there are no objections I would like to go on with 5.0.0-preview on `develop`.

Shane32 (commented):
I would not put GraphQL extension methods into the Microsoft namespace.  Besides that, I have no comments.

BenjaBobs:
> I would not put GraphQL extension methods into the Microsoft namespace. Besides that, I have no comments.

That's a good point, borh Visual Studio and Rider supports auto-completion from unimported namespaces, so extension method discovery is no longer a problem.

benmccallum (approved):

sungam3r:
But not everyone knows about it. It is just a good convention to place extensions inside "appropriate" namespaces and make them work out of the box. MS tends to follow this convention along .net core codebase.

PR #340, opened 2020-02-29
add GraphQLHttpMiddleware docs

### LINKED ISSUE #249: Mutation in GET request
According to [this spec](https://github.com/APIs-guru/graphql-over-http):

> GET requests can be used for executing ONLY queries. If the values of query and operationName indicates that a non-query operation is to be executed, the server should immediately respond with an error status code, and halt execution.

Now `GraphQLHttpMiddleware` does not have such restrictions, it simply parses a request from the  query and executes it. Is it worth it to forbid executing mutations in GET http requests?

sungam3r:
fixes #249

Honestly, I'm not sure that such a restriction should be made (breaking although it meets the specification), so for now I have limited myself to just documenting the current behavior. We can add a restriction in the future.

### DIFF
--- a/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
+++ b/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
@@ -13,6 +13,17 @@ using System.Threading.Tasks;
 
 namespace GraphQL.Server.Transports.AspNetCore
 {
+    /// <summary>
+    /// ASP.NET Core middleware for processing GraphQL requests.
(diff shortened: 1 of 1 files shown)


sungam3r:
@benmccallum Agree?

benmccallum:
The only reason I guess we should implement the restriction would be if there was a security vulnerability. I know there can be issues returning JSON over GET (in restful services in general), but can't remember the exact issue. Session hi-jacking maybe.

I'm happy with this for now.

PR #371, opened 2020-07-08
Use replace instead of add

sungam3r:
Before this change I got error from
```c#
 throw new InvalidOperationException(
                    "IDocumentWriter not set in DI container. " +
                    "Add a IDocumentWriter implementation, for example " +
                    "GraphQL.SystemTextJson.DocumentWriter or GraphQL.NewtonsoftJson.DocumentWriter." +
                    "For more information, see: https://github.com/graphql-dotnet/graphql-dotnet/blob/master/README.md and https://github.com/graphql-dotnet/server/blob/develop/README.md.");
```
if I call GetService**S**.

### DIFF
--- a/src/Transports.AspNetCore.NewtonsoftJson/GraphQLBuilderExtensions.cs
+++ b/src/Transports.AspNetCore.NewtonsoftJson/GraphQLBuilderExtensions.cs
@@ -29,7 +29,7 @@ namespace GraphQL.Server
             Action<JsonSerializerSettings> configureSerializerSettings = null)
         {
             builder.Services.AddSingleton<IGraphQLRequestDeserializer>(p => new GraphQLRequestDeserializer(configureDeserializerSettings));
-            builder.Services.Replace(new ServiceDescriptor(typeof(IDocumentWriter), p => new DocumentWriter(configureSerializerSettings), ServiceLifetime.Singleton));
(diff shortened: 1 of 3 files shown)


BenjaBobs:
Good catch! Maybe it should be `builder.Services.TryAdd()` though, so the consumer and specify an override?

sungam3r:
We should **replace** stab added for just show warning about serializer. `TryAdd` will not work.

sungam3r:
@benmccallum Are you OK with these changes? IIRC you are the author of serialization extensions.

benmccallum:
Hey @sungam3r , will mirror my comment on another ticket. We've recently moved to HotChocolate so I've shifted my focus there.

benmccallum:
I guess if there's a unit tests and docs/examples showing a working setups, and like @BenjaBobs  said, people can still sub in their own implementation, I'd agree with any approach you take

BenjaBobs:
@benmccallum I know the pr is closed, but may I ask why you switched to HotChocolate? Because of something you were able to with that library that you aren't able to do with GraphQL Dotnet? Or do you simply prefer the design of the api? And probably more importantly, what do you feel is lacking in this library?

benmccallum:
Hey @BenjaBobs and @sungam3r, sorry for the late reply, I will get back to giving you my thoughts, it'll just take me a while to complete them and we're not 100% on HotChoc yet for me to draw final conclusions. I've bookmarked this notification though, so I'll come back to it when I get a sec!

PR #372, opened 2020-07-12
Extensions for GraphQLRequest

### LINKED ISSUE #352: Add Extensions to GraphQLRequest
Hello, currenly we implement [Apollo Persisted Query](https://www.apollographql.com/docs/apollo-server/performance/apq/) in custom GraphQLHttpMiddleware.
After split request deserialization in separate nuget it got a little harder to do)
Can we have extensions field in [GraphQLRequest](https://github.com/graphql-dotnet/server/blob/develop/src/Core/Common/GraphQLRequest.cs) like Inputs?

sungam3r:
fixes #352 

@benmccallum @SimonCropp @joemcbride What do you think? I see that work is underway on the first release of the http transport specification. Should we consider merge in the master branch of http spec as an indicator that we need to move implementation forward?

### DIFF
--- a/src/Core/Common/GraphQLRequest.cs
+++ b/src/Core/Common/GraphQLRequest.cs
@@ -5,9 +5,11 @@ namespace GraphQL.Server.Common
         public const string OPERATION_NAME_KEY = "operationName";
         public const string QUERY_KEY = "query";
         public const string VARIABLES_KEY = "variables";
+        public const string EXTENSIONS_KEY = "extensions";
 
         public string OperationName { get; set; }
         public string Query { get; set; }
         public Inputs Inputs { get; set; }
+        public Inputs Extensions { get; set; }
     }
 }
--- a/src/Core/Common/GraphQLRequest.cs
+++ b/src/Core/Common/GraphQLRequest.cs
@@ -1,15 +1,53 @@
 namespace GraphQL.Server.Common
 {
+    /// <summary>
+    /// Represents data sent by client to GraphQL server.
+    /// See https://github.com/graphql/graphql-over-http/blob/master/spec/GraphQLOverHTTP.md#request
+    /// </summary>
     public class GraphQLRequest
     {
+        /// <summary>
+        /// Name for the operation name parameter.
+        /// See https://github.com/graphql/graphql-over-http/blob/master/spec/GraphQLOverHTTP.md#request-parameters
+        /// </summary>
         public const string OPERATION_NAME_KEY = "operationName";
+
+        /// <summary>
+        /// Name for the query parameter.
+        /// See https://github.com/graphql/graphql-over-http/blob/master/spec/GraphQLOverHTTP.md#request-parameters
+        /// </summary>
         public const string QUERY_KEY = "query";
+
+        /// <summary>
+        /// Name for the variables parameter.
+        /// See https://github.com/graphql/graphql-over-http/blob/master/spec/GraphQLOverHTTP.md#request-parameters
+        /// </summary>
         public const string VARIABLES_KEY = "variables";
+
+        /// <summary>
+        /// Name for the extensions parameter.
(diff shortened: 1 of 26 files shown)


src/Core/Common/GraphQLRequest.cs
```diff
@@ -5,9 +5,11 @@ public class GraphQLRequest
         public const string OPERATION_NAME_KEY = "operationName";
         public const string QUERY_KEY = "query";
         public const string VARIABLES_KEY = "variables";
+        public const string EXTENSIONS_KEY = "extensions";
 
         public string OperationName { get; set; }
         public string Query { get; set; }
         public Inputs Inputs { get; set; }
+        public Inputs Extensions { get; set; }
```
sungam3r, line 13: Using `Inputs` type here is a temporary measure. We can introduce a new type, although this one should fit.

benmccallum:
Sorry @sungam3r , we've been moving to HotChocolate so my focus is really over there now.

sungam3r:
@benmccallum It is a little sad for the development of this project. If this is not a secret, can you talk about the reasons?

SimonCropp:
i think we need to ship a new stable asap

sungam3r:
Well, it is a very good question. Relates to https://github.com/graphql-dotnet/graphql-dotnet/issues/1321

benmccallum:
As I mentioned on another ticket, I'll give some proper feedback when I'm fully on HotChocolate and have had a moment to reflect, but as @SimonCropp alludes to, a real release cycle (including previews, rcs, blog posts, Slack group, requests for feedback) was definitely one of the biggest factors. I was always happy to use graphql-dotnet's previews, but at a certain point it's not enough as the project can't gain any momentum/clear direction without major/minor releases. There's nothing to gather around, plan around or get excited about. 

To be clear, I think you (and others) @sungam3r have done an absolutely amazing job to make real progress from where it was say a year ago! And with some release cycles and some community stuff to make it all more public, it'd continue to do really well. But HotChocolate is already there. The community is thriving and we're getting excited about things like v11 which has a new execution engine with almost zero allocations, (+ they've done metrics and it's numbers of times faster than graphql-dotnet) + it's gonna drop some real-time features like `@defer` and `@stream`. 

I'll list out some more things when I get some more time.

benmccallum:
Essentially just biting the bullet and get a v3 out would be huge. I know you guys make breaking changes quite often, and that may be a reason for the v3 delay to finalise a bunch of breaking things into one, but frankly it'd be better to have v3, v4, v5 in quick succession, as it's much clearer than breaking changes between preview 1423 and 1424 lol. + it's much easier to document each time. + it shows progression.

I know this isn't any of your fault though, you're just doing the best you can with the access you've got!

sungam3r:
@benmccallum We migrated to GitHub Actions CI and finally published v3.0.0 (and 3.1.0/3.1.1 already). See https://github.com/graphql-dotnet/graphql-dotnet/issues/1321#issuecomment-717656610

src/Core/Common/GraphQLRequest.cs
sungam3r: I think to keep `Inputs` type for this property.

sungam3r:
Also this PR fixes a lot of warnings from check annotations. See https://github.com/graphql-dotnet/server/pull/418/files

sungam3r:
@Shane32 @joemcbride I want to merge and release ASAP after review.

Shane32:
I’ll review today

Shane32 (approved):
As long as you’re sure we don’t need any of the items removed from gitignore, I’m good. What version number do you want to publish as?

Shane32:
4.1.1 to mirror 3.1.1 release of GraphQL.NET?

sungam3r:
There is new public property so 4.2.2
изображение

PR #378, opened 2020-08-12
if any result has any error Data should be null

miladkakouli:
if result had any errors Data should be null so that relay can CatchError


BenjaBobs:
The [spec](https://spec.graphql.org/June2018/#sec-Data) actually says that this should only happen if the error prevents a valid response:

> If an error was encountered during the execution that prevented a valid response, the data entry in the response should be null.

Maybe `ExecutionError` should have an `IsFatal` property that can be set to indicate whether it is such an error?

PR #398, opened 2020-09-17
Error code and number for AuthorizationValidationRule

### LINKED ISSUE #390: Regression: Unauthorized access to node returns error with code "VALIDATION_ERROR" instead of "authorization"
Hi there,

I was playing around with the new GraphQL.Server 4.0.0 and its dependencies and bumped into a minor problem.

I spotted a slight error in Authorization.AspNetCore\AuthorizationValidationRule.cs, in method AuthorizationValidationRule.AuthorizeAsync(...).

`new ValidationError(context.OriginalQuery, "authorization", stringBuilder.ToString(), node)`

The constructor of ValidationError had a minor change in GraphQL 3.0.0.
**Previously:**
`public ValidationError(string originalQuery, string code, string message, params INode[] nodes)
`
where "code" is some form of technical error code that a developer could define.

**Now:**
`public ValidationError(string originalQuery, string number, string message, params INode[] nodes)
`
where "number" is the section number of the GraphQL specification that the error violates.

Technically, this is only a minor change as it does not break compilation, but it is a actually a change of method signature.

In order the make sure the error code is properly set, you can set it explicitly with the corresponding `Code` property.
```
new ValidationError(context.OriginalQuery, "6.1.1", stringBuilder.ToString(), node)
{
	Code = "authorization"
}
```
Alternatively, you could implement your own error error class that derives from ValidationError. That would probably be a better approach.

The value "6.1.1" of parameter "number" refers to the section about "Validating Requests" in the GraphQL spec. You might be able to find a more appropriate number.

Kind regards,
Reureu

sungam3r:
Fixes #390

### DIFF
--- a/src/Authorization.AspNetCore/AuthorizationValidationRule.cs
+++ b/src/Authorization.AspNetCore/AuthorizationValidationRule.cs
@@ -107,8 +107,10 @@ namespace GraphQL.Server.Authorization.AspNetCore
                         AppendFailureLine(stringBuilder, failure);
                     }
 
-                    context.ReportError(
-                        new ValidationError(context.OriginalQuery, "authorization", stringBuilder.ToString(), node));
+                    context.ReportError(new ValidationError(context.OriginalQuery, "6.1.1", stringBuilder.ToString(), node)
+                    {
+                        Code = "authorization",
+                    });
                 }
             }
         }


Shane32 (approved):
That will work great.  I would probably add a class but it isn't necessary.

PR #427, opened 2020-11-02
netcoreapp3.1 only

### LINKED ISSUE #304: End support for netcoreapp2.2
It's now hit EOL, so we should just document in the readme the last version that will support it and drop it.

To change:
* Samples.Server.csproj
** Startup.cs 
* Samples.Server.Tests.csproj
* Directory.Build.props

To remove:
* Serialization.NewtonsoftJson (unless we want to say it can be used with `AllowSynchronousIO` set to true, but that to me is a really bad idea)

**Update** netcoreapp2.1 (LTS release) EOLs on 2021-08-21, which is ages away. So, given we don't target 2.2 specifically (except in the sample projects), we'd almost need to re-target them to 2.1. It's a brave new world.

sungam3r:
Removes 2.2 and 3.0 TFMs. Fixes #304 . I would like to move on with `develop` branch and bump version to 5.0.0-preview-XXX . We will still support version 4.x.x on `master`. The release of the 5.0.0 version may take place some time later, maybe next year. I still have some thoughts on what needs to be redone.


sungam3r:
ping @benmccallum

benmccallum (approved):
Looks good. Nice cleanup of all the stuff is put in to multi target. :)

Shane32:
This is great @sungam3r !  Side note: we should add support for .Net 5 ...

Shane32 (approved):

sungam3r:
I would not rush with net5. When it comes out, then it will be possible to think about supporting it, but for now there is a lot of other work.

Shane32:
> I would not rush with net5. When it comes out, then it will be possible to think about supporting it, but for now there is a lot of other work.

It should be released next week. It’s on RC2 now.

PR #434, opened 2020-11-07
Trigger CI without tests

sungam3r:


sungam3r:
It works! Finally resolves #40.

PR #440, opened 2020-11-08
Rework of all middlewares

### LINKED ISSUE #175: Playground running on /graphql
On the pre-release version, we had configured the path for playground to be /graphql/ui/playground which worked fine. 
This was done to have a single endpoint of our api gateway for grapghql.
However on the latest version this does not work anymore , it appears all requests are routed to the graphql engine, including graphql/ui/playground.
So when you try to run playground , you get the error "GraphQL.ExecutionError: A query is required"

Is there any way we could keep our current playground path with the new release?

### LINKED ISSUE #430: Include extensions for ASP.NET Core 3.x Endpoints for easy use of Authorization
Been looking for information on how-to make use of `Endpoints` in ASP.NET Core 3.x to hook in the middleware. Was expecting it to exist in the library. Is there something similar to the sample below? If not, is it worth a PR or worth to be included in docs somewhere?

```csharp
    public static class GraphQlEndpointRouteBuilderExtensions
    {
        public static IEndpointConventionBuilder MapGraphQl<TSchema>(
            this IEndpointRouteBuilder endpoints, 
            string pattern = "/graphql") where TSchema : ISchema
        {
            var pipeline = endpoints
                .CreateApplicationBuilder()
                .UseMiddleware<GraphQLHttpMiddleware<TSchema>>(new PathString())
                .Build();
                
            return endpoints.Map(pattern, pipeline);
        }
    }
```

Which then allows you to do the following in e.g. `Startup`:

```csharp
public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
{
    if (env.IsDevelopment())
    {
        app.UseDeveloperExceptionPage();
    }

    app.UseHttpsRedirection();

    app.UseRouting();

    app.UseAuthorization();

    app.UseEndpoints(endpoints =>
    {
        endpoints
            .MapGraphQl<PetsSchema>()
            .RequireAuthorization();
    });
}
```

sungam3r:
- add ASP.NET Core routing support and example in readme
- add setting for separate subscriptions endpoint to Altair, Playground, GraphiQL
- add headers setting to GraphiQL
- unify naming across middlewares, options, extension methods, classes

fixes #430 
fixes #175 

See https://docs.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-3.1


sungam3r:
@Shane32 @benmccallum I would really appreciate a review. There are a lot of changes and I had to do a lot of monotonous typing across all projects to bring everything to a similar look.

Shane32 (approved):

README.md
```diff
@@ -112,20 +115,77 @@ public void Configure(IApplicationBuilder app)
...
+        // map GraphiQL middleware at default path /ui/graphiql with default options
+        endpoints.MapGraphQLGraphiQL();
+
+        // map Altair middleware at default path /ui/altair with default options
+        endpoints.MapGraphQLAltair();
+
+        // map Voyager middleware at default path /ui/voyager with default options
+        endpoints.MapGraphQLVoyager();
+}
+```
```
benmccallum, line 187: Worth taking a look at @SimonCropp 's MarkdownSnippet for these. https://github.com/SimonCropp/MarkdownSnippets

sungam3r: I roughly understood what this is for, but it seems too complicated to me. It is not difficult for me to update the readme manually.

Issue #451, opened 2020-11-11
Possibility to add "GraphiQL Explorer"?


varasinp:
I would like to know if there is a plan to include GraphiQL Explorer (part of GraphiQL).
GraphiQL Explorer is very useful. It allows users to point and click on GraphQL node and it auto-generate the query.
Here's the example of GraphiQL Explorer https://api.spacex.land/graphql/

sungam3r:
I've seen this UI somewhere. GraphiQL middleware just returns "simple" html - https://github.com/graphql-dotnet/server/blob/master/src/Ui.GraphiQL/Internal/graphiql.cshtml. I am not an expert in JS and frontend. Do you know how to change html? Are more serious server side changes required to support this functionality?

varasinp:
Sorry, I'm not an expert when it comes to front-end stuff like JavaScript or TypeScript. However, I don't believe there's a need to change a lot of code on the server-side. I believe only the GraphiQL version update is required.

I look into the "graphiql.cshtml" file that you mentioned and I noticed this (line 42 & 43)

```html
<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/graphiql/0.13.0/graphiql.min.css" />
<script src="https://cdnjs.cloudflare.com/ajax/libs/graphiql/0.13.0/graphiql.min.js"></script>
```
As you can see, the graphiql version is "0.13.0" which is the older version. Maybe all that needs to be done is just to use the latest version which is "1.0.6"? https://cdnjs.com/libraries/graphiql 

So line 42 & 43 in "graphiql.cshtml" should be modified to use version "1.0.6" instead probably. I hope that's will do it but I didn't try that myself just yet.
```html
<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/graphiql/1.0.6/graphiql.min.css" />
<script src="https://cdnjs.cloudflare.com/ajax/libs/graphiql/1.0.6/graphiql.min.js"></script>
```

I went to the https://github.com/graphql/graphiql and I saw this picture. They specicy GraphiQL Explorer as an external key (not sure what that means)
![GraphiQL](https://raw.githubusercontent.com/graphql/graphiql/main/resources/images/current-ecosystem.jpg)

If this work, maybe others UI (Altair, Playground, Voyager) need a version update also probably?

varasinp:
[...]

PR #455, opened 2020-11-21
Replaced GraphiQL with GraphiQLWithExtensions

imolorhe:
This replaces the default GraphiQL
with one that includes GraphiQL explorer

Addresses #451

### DIFF
--- a/src/Ui.GraphiQL/Internal/graphiql.cshtml
+++ b/src/Ui.GraphiQL/Internal/graphiql.cshtml
@@ -40,7 +40,7 @@
 	  favored resource bundler.
 	 -->
 	<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/graphiql/0.13.0/graphiql.min.css" />
-	<script src="https://cdnjs.cloudflare.com/ajax/libs/graphiql/0.13.0/graphiql.min.js"></script>
+	<script src="https://cdn.jsdelivr.net/npm/graphiql-with-extensions@0.14.3/graphiqlWithExtensions.min.js"></script>
 
 	<script src="https://unpkg.com/subscriptions-transport-ws@0.8.3/browser/client.js"></script>
 	<script src="https://unpkg.com/graphiql-subscriptions-fetcher@0.0.2/browser/client.js"></script>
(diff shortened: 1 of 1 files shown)


sungam3r:
So simple?!

src/Ui.GraphiQL/Internal/graphiql.cshtml
```diff
@@ -40,7 +40,7 @@
 	  favored resource bundler.
 	 -->
 	<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/graphiql/0.13.0/graphiql.min.css" />
```
sungam3r, line 42: 1.0.6 ?

src/Ui.GraphiQL/Internal/graphiql.cshtml
```diff
@@ -40,7 +40,7 @@
...
-	<script src="https://cdnjs.cloudflare.com/ajax/libs/graphiql/0.13.0/graphiql.min.js"></script>
+	<script src="https://cdn.jsdelivr.net/npm/graphiql-with-extensions@0.14.3/graphiqlWithExtensions.min.js"></script>
 
 	<script src="https://unpkg.com/subscriptions-transport-ws@0.8.3/browser/client.js"></script>
```
sungam3r, line 45: 0.9.18 ?

sungam3r:
Thanks @imolorhe , it works! I just have one question about bumping versions for two js packages. Is it worth doing this? We haven't changed this cshtml for a long time.

imolorhe:
> So simple?!

😄 Ideally it's meant to be more complicated but the folks that built the GraphiQL explorer anticipated this kind of use case and created an already bundled version that is just a drop-in replacement for GraphiQL.

varasinp:
I didn't realize that Onegraph people already created that "graphiqlWithExtensions.js:" 
Thank you!!

I also wonder whether or not we should upgrade the version number for the two js packages like @sungam3r  mentioned.

imolorhe:
I've update the package versions and tested them

sungam3r (approved):

PR #457, opened 2020-11-26
Rename GraphQLBuilderExtensions to GraphQLAuthorizationBuilderExtensions

### LINKED ISSUE #456: GraphQL.Server.GraphQLBuilderExtensions class is defined multiple times in different assemblies
[GraphQL.Server.GraphQLBuilderExtensions in GraphQL.Sever.Authorization.AspNetCore assembly](https://github.com/graphql-dotnet/server/blob/master/src/Authorization.AspNetCore/GraphQLBuilderExtensions.cs)
[GraphQL.Server.GraphQLBuilderExtensions in GraphQL.Sever.Core assembly](https://github.com/graphql-dotnet/server/blob/master/src/Core/GraphQLBuilderExtensions.cs)

This may be the cause of some issues such as in #353 because it becomes near impossible to reference methods from both.

infogulch:
Fixes #456

### DIFF
--- a/Tests.props
+++ b/Tests.props
@@ -9,7 +9,7 @@
     <PackageReference Include="Shouldly" Version="4.0.1" />
     <PackageReference Include="xunit" Version="2.4.1" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3" />
-    <PackageReference Include="Moq" Version="4.15.1" />
+    <PackageReference Include="Moq" Version="4.15.2" />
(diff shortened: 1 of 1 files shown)


sungam3r:
You renamed only file name.

infogulch:
Oops

infogulch:
I accidentally clicked close, then I tried to rebase and lost my changes, then I committed with the wrong email.  Sorry!  These should be fixed now.

infogulch:
Ok github doesn't like this PR anymore.  I opened a new one at #461

PR #461, opened 2020-11-27
Rename all GraphQLBuilderExtensions classes to unique names

infogulch:
Take 2!  Sorry for the noise.  😄

### DIFF
--- a/src/Core/Extensions/GraphQLBuilderServerCoreExtensions.cs
+++ b/src/Core/Extensions/GraphQLBuilderServerCoreExtensions.cs
@@ -14,7 +14,7 @@ namespace GraphQL.Server
     /// <summary>
     /// GraphQL specific extension methods for <see cref="IGraphQLBuilder"/>.
     /// </summary>
-    public static class GraphQLBuilderExtensions
+    public static class GraphQLBuilderServerCoreExtensions
     {
         // Provides integration with Microsoft.Extensions.Options so the caller may use services.Configure<ErrorInfoProviderOptions>(...)
         private sealed class InternalErrorInfoProvider : ErrorInfoProvider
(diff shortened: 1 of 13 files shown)


sungam3r:
In principle, I am not against such changes, although they will be useful in a very limited number of cases. Since these are breaking changes you should change target branch to `develop`. @Shane32 please review.

infogulch:
I changed the target to the develop branch.

I agree, this does change the API surface so it is technically a breaking change.  Though, as you said, referencing these extension classes by name is `different from the "usual" use of the extension methods of these classes`, so it should not affect any user in practice.  In fact, the `using` statement I mentioned in #456 demonstrates that if any user *did* reference these classes by name they likely encountered a compiler error.

sungam3r:
Now you have done a PR with 16 commits.

infogulch:
I have redone the changes on top of the develop branch instead.

infogulch:
Should we consider these changes a breaking change when no user could have referenced any of the changed APIs by name without encountering a compiler error?

sungam3r:
It's binary breaking changes, yes.

src/Core/Extensions/GraphQLBuilderServerCoreExtensions.cs
```diff
@@ -14,7 +14,7 @@ namespace GraphQL.Server
     /// <summary>
     /// GraphQL specific extension methods for <see cref="IGraphQLBuilder"/>.
     /// </summary>
-    public static class GraphQLBuilderExtensions
+    public static class GraphQLBuilderServerCoreExtensions
```
sungam3r, line 17:
```suggestion
    public static class GraphQLBuilderCoreExtensions
```

Shane32 (approved):

infogulch:
Thank you for being patient while we worked through the issues on this PR!

sungam3r (approved):

PR #469, opened 2020-12-06
Correct launch urls in sample project

mungojam:
When I first tried launching the project (using Rider and .net 5 SDK), it started up `http://localhost:60341//ui/playground` in the browser which didn't work due to the double `/`.

This is fixed by removing the trailing slash from the `applicationUrl`.


sungam3r:
Well, it seems to be an issue with Rider. I see that VS 2019 works both ways - with or without trailing slash.

PR #478, opened 2020-12-21
Delegate generation of authorization failure message to injected class

rose-a:
Allows customization of authorization error messages through dependency injection

### DIFF
--- a/src/Authorization.AspNetCore/AuthorizationValidationRule.cs
+++ b/src/Authorization.AspNetCore/AuthorizationValidationRule.cs
@@ -1,13 +1,10 @@
 using System;
 using System.Collections.Generic;
-using System.Linq;
-using System.Text;
 using System.Threading.Tasks;
 using GraphQL.Language.AST;
 using GraphQL.Types;
 using GraphQL.Validation;
 using Microsoft.AspNetCore.Authorization;
-using Microsoft.AspNetCore.Authorization.Infrastructure;
 using Microsoft.AspNetCore.Http;
 
 namespace GraphQL.Server.Authorization.AspNetCore
@@ -20,16 +17,18 @@ namespace GraphQL.Server.Authorization.AspNetCore
     {
         private readonly IAuthorizationService _authorizationService;
         private readonly IHttpContextAccessor _httpContextAccessor;
+        private readonly IAuthorizationFailureDescriptionGenerator _failureDescriptionGenerator;
 
         /// <summary>
         /// Creates an instance of <see cref="AuthorizationValidationRule"/>.
         /// </summary>
         /// <param name="authorizationService"> ASP.NET Core <see cref="IAuthorizationService"/> to authorize against. </param>
         /// <param name="httpContextAccessor"> ASP.NET Core <see cref="IHttpContextAccessor"/> to take user (HttpContext.User) from. </param>
-        public AuthorizationValidationRule(IAuthorizationService authorizationService, IHttpContextAccessor httpContextAccessor)
+        public AuthorizationValidationRule(IAuthorizationService authorizationService, IHttpContextAccessor httpContextAccessor, IAuthorizationFailureDescriptionGenerator failureDescriptionGenerator)
         {
             _authorizationService = authorizationService;
             _httpContextAccessor = httpContextAccessor;
+            _failureDescriptionGenerator = failureDescriptionGenerator;
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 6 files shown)


Shane32 (commented):
I'm not sure about this PR.  Usually one does not see interfaces/classes for generating error messages.  I would think a more typical coding pattern would be to have each of these different error messages return a specific class (that inherits `ExecutionError`), with properties for the specifics of the error, and then after the document was executed, some code can enumerate the errors and replace them as necessary with custom error messages or classes.

For example (untested, just a sample):
```cs
for (int i = 0; i <= result.Errors.Count; i++)
{
    var error = result.Errors[i];
    if (error is NotAuthorizedError notAuthError)
        error.Message = $"You are not authorized for policy '{notAuthError.PolicyName}'";
}
```

src/Authorization.AspNetCore/IAuthorizationFailureDescriptionGenerator.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    /// <summary>
+    /// The interface for an authorization failure description generator
+    /// </summary>
+    public interface IAuthorizationFailureDescriptionGenerator
+    {
+        /// <summary>
+        /// Generates a failure description for the given <paramref name="result"/> 
+        /// </summary>
+        /// <param name="result">the ASP.NET Core authorization result</param>
```
sungam3r, line 14:
```suggestion
        /// <param name="result">The ASP.NET Core authorization result</param>
```

src/Authorization.AspNetCore/IAuthorizationFailureDescriptionGenerator.cs
```diff
@@ -0,0 +1,19 @@
...
+        /// <param name="operationType">the GraphQL operation type</param>
```
sungam3r, line 15:
```suggestion
        /// <param name="operationType">The GraphQL operation type. Not set if authorization occurs for the schema itself</param>
```

src/Authorization.AspNetCore/DefaultAuthorizationFailureDescriptionGenerator.cs
```diff
@@ -0,0 +1,106 @@
...
+            foreach (var failure in result.Failure.FailedRequirements)
+                AppendFailureLine(messageBuilder, failure);
+
+            return messageBuilder.ToString();
+        }
+
+        /// <summary>
+        /// Transforms the given <paramref name="operationType"/> into a string to be used in a failure description
+        /// </summary>
+        /// <param name="operationType">the GraphQL operation type</param>
```
sungam3r, line 30:
```suggestion
        /// <param name="operationType">The GraphQL operation type</param>
```

src/Authorization.AspNetCore/DefaultAuthorizationFailureDescriptionGenerator.cs
```diff
@@ -0,0 +1,106 @@
+using System.Linq;
+using System.Text;
+using GraphQL.Language.AST;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Authorization.Infrastructure;
+
+namespace GraphQL.Server.Authorization.AspNetCore
+{
+    /// <summary>
+    /// The default authorization failure description generator for this library
```
sungam3r, line 10:
```suggestion
    /// The default authorization failure description generator.
```

src/Authorization.AspNetCore/DefaultAuthorizationFailureDescriptionGenerator.cs
```diff
@@ -0,0 +1,106 @@
...
+                OperationType.Query => "query",
+                OperationType.Mutation => "mutation",
+                OperationType.Subscription => "subscription",
+                _ => "operation",
+            };
+
+        /// <summary>
+        /// Generates a description for a single failed <see cref="IAuthorizationRequirement"/>
+        /// </summary>
+        /// <param name="messageBuilder">the message string builder</param>
```
sungam3r, line 44:
```suggestion
        /// <param name="messageBuilder">The message string builder to accumulate generated text</param>
```

src/Authorization.AspNetCore/DefaultAuthorizationFailureDescriptionGenerator.cs
```diff
@@ -0,0 +1,106 @@
...
+        /// <param name="authorizationRequirement">the failed authorization requirement</param>
+        protected virtual void AppendFailureLine(StringBuilder messageBuilder, IAuthorizationRequirement authorizationRequirement)
```
sungam3r, line 46: messageBuilder -> error

sungam3r:
изображение

sungam3r:
> I would think a more typical coding pattern would be to have each of these different error messages return a specific class (that inherits ExecutionError), with properties for the specifics of the error, and then after the document was executed, some code can enumerate the errors and replace them as necessary with custom error messages or classes.

This solution can be applied, but it is more difficult to maintain. In addition, we already have ExecutionError with a very **definite meaning**, we do not need **new types** of errors, the whole question is to get an auxiliary description.

sungam3r:
@Shane32 My previous comment was wrong. Indeed, it is possible to make separate errors instead of _generic_ one. It might have been worth doing right from the start of project, but now it would be a backward incompatible change and would require a clients change. I like the idea with separate errors. @rose-a ?

Shane32:
Another idea would be to simply add protected methods to the `AuthorizationValidationRule` class to generate the respective error messages.  If desired, one can simply replace the validation rule with their own implementation, overriding the members required.

I also drafted a sample of a non-breaking change to the existing codebase that demonstrates moving all the error message generation code into a new `AuthorizationError` class.  Unfortunately `Exception.Message` is a read-only property so it is not easy to override a message once the class has been constructed.  However, by making `AuthorizationValidationRule.AddValidationError` protected, it is easy to replace the entire validation rule with an implementation that creates a custom derivative of `AuthorizationError`.  This can be further separated by having the code within `AddValidationError` create a specific class (deriving from `AuthorizationError`) depending on which exact type of `IAuthorizationRequirement` failed.

I suggest we also add a PR to the main project that adds a `SetMessage(string message)` method to `ExecutionError`.  This would allow for an easier way to adjust messages after the class has been created.

Please keep in mind that I don't actually use the server project, and so you both are probably more familiar with how this would work in practice @sungam3r @rose-a .

sungam3r:
> I suggest we also add a PR to the main project that adds a SetMessage(string message) method to ExecutionError. This would allow for an easier way to adjust messages after the class has been created.

I would not do that.

Shane32:
> > I suggest we also add a PR to the main project that adds a SetMessage(string message) method to ExecutionError. This would allow for an easier way to adjust messages after the class has been created.
> 
> I would not do that.

By that logic, `AddLocation` and `Code { set; }` and `Number { set; }` should be `protected`, not `public`.  But if everything is writable, then `Message` should be as well.

rose-a:
I'm ok with creating "error classes" like @Shane32 started in #479... But I'd go further and make `AuthorizationError` abstract and create a `MissingClaimsAuthorizationError`, `NameAuthorizationError`, ... which provide the appropriate message.

I'd still like to put the switch logic in a factory class to be able to inject a custom version which returns customized errors in certain cases...

Another way would be to put the error generation for each case in a separate virtual method to be able to easily override just one of them and leave the switch logic alone...

rose-a:
> 
> 
> > > I suggest we also add a PR to the main project that adds a SetMessage(string message) method to ExecutionError. This would allow for an easier way to adjust messages after the class has been created.
> > 
> > 
> > I would not do that.
> 
> By that logic, `AddLocation` and `Code { set; }` and `Number { set; }` should be `protected`, not `public`. But if everything is writable, then `Message` should be as well.

I think that's because the `Message` property of `Exception` is not writeable...

rose-a:
> I'm not sure about this PR. Usually one does not see interfaces/classes for generating error messages. 

Well, it was the simplest way to create a custom error message for my project 😉

Shane32:
> I'm ok with creating "error classes" like @Shane32 started in #479... But I'd go further and make `AuthorizationError` abstract and create a `MissingClaimsAuthorizationError`, `NameAuthorizationError`, ... which provide the appropriate message.

I agree

Shane32:
> Another way would be to put the error generation for each case in a separate virtual method to be able to easily override just one of them and leave the switch logic alone...

I like that also

sungam3r:
> But if everything is writable, then Message should be as well.

This is a consequence of inheritance from `System.Exception`. We should to take it as it is.

sungam3r:
> But I'd go further and make AuthorizationError abstract and create a MissingClaimsAuthorizationError, NameAuthorizationError, ... which provide the appropriate message.

Of course it should be done.

src/Authorization.AspNetCore/AuthorizationValidationRule.cs
```diff
@@ -20,16 +17,19 @@ public class AuthorizationValidationRule : IValidationRule
...
         private readonly IHttpContextAccessor _httpContextAccessor;
+        private readonly IAuthorizationFailureDescriptionGenerator _failureDescriptionGenerator;
 
         /// <summary>
         /// Creates an instance of <see cref="AuthorizationValidationRule"/>.
         /// </summary>
         /// <param name="authorizationService"> ASP.NET Core <see cref="IAuthorizationService"/> to authorize against. </param>
         /// <param name="httpContextAccessor"> ASP.NET Core <see cref="IHttpContextAccessor"/> to take user (HttpContext.User) from. </param>
-        public AuthorizationValidationRule(IAuthorizationService authorizationService, IHttpContextAccessor httpContextAccessor)
+        /// <param name="failureDescriptionGenerator"> The generator for authorization failure description messages</param>
```
sungam3r, line 27:
```suggestion
        /// <param name="failureDescriptionGenerator"> The generator for authorization failure description messages. </param>
```

Shane32:
> > But I'd go further and make AuthorizationError abstract and create a MissingClaimsAuthorizationError, NameAuthorizationError, ... which provide the appropriate message.
> 
> Of course it should be done.

Once completed, an easy solution to the original problem is to provide an `IErrorInfoProvider` implementation like this:

```cs
    public class MyErrorInfoProvider : ErrorInfoProvider
    {
        public override ErrorInfo GetInfo(ExecutionError executionError)
        {
            var info = base.GetInfo(executionError);
            info.Message = executionError switch
            {
                SyntaxError syntaxError => "Syntax error. READY",
                ValidationError validationError => $"I'm afraid I can't do that, Dave.  You've got a {validationError.Number}",
                _ => info.Message,
            };
            return info;
        }
    }
```
Injecting an `IErrorInfoProvider` is already built into the framework.

sungam3r:
So as you demonstrated there is no need to have setter for `ExecutionError.Message` since we can return arbitrary text via `ErrorInfo`.

rose-a:
What do we do about `InternalErrorInfoProvider` inside `GraphQLBuilderCoreExtensions`?

rose-a:
After playing around a little bit i don't think it makes sense to make separate classes for the different authorization errors... since one `AuthorizationError` can contain multiple failed requirements...

I`m willing close this PR in favour of #479...

sungam3r:
OK. You may branch from #479 and move on.

rose-a:
closing this in favour of #480

PR #479, opened 2020-12-21
Add AuthorizationError class

Shane32:
See discussion within #478


src/Authorization.AspNetCore/AuthorizationError.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// </summary>
+        public AuthorizationError(INode node, ValidationContext context, string message, AuthorizationResult result)
+            : base(context.OriginalQuery, "6.1.1", message, node == null ? Array.Empty<INode>() : new INode[] { node })
+        {
+            Code = "authorization";
+            AuthorizationResult = result;
+        }
+
+        /// <summary>
+        /// Returns the result of the Asp.Net Core authorization request
```
sungam3r, line 35:
```suggestion
        /// Returns the result of the ASP.NET Core authorization request
```

src/Authorization.AspNetCore/AuthorizationValidationRule.cs
```diff
@@ -102,100 +103,21 @@ private async Task AuthorizeAsync(INode node, IProvideMetadata type, ValidationC
...
-                OperationType.Subscription => "subscription",
-                _ => "operation",
-            };
         }
 
-        private static void AppendFailureLine(StringBuilder error, IAuthorizationRequirement authorizationRequirement)
+        /// <summary>
+        /// Adds an authorization failure error to the document response
+        /// </summary>
+        protected void AddValidationError(INode node, ValidationContext context, OperationType? operationType, AuthorizationResult result)
```
sungam3r, line 115: Not virtual? So how to change  `new AuthorizationError` to something other?

Shane32: `protected` infers `virtual`.  It is illegal to write `protected virtual`.

Shane32: (or at least the code style rules disallow it)

rose-a:
I like the error class because it took me a while to find out where that mysterious number "6.1.1" came from ;)

The injection of `AuthorizationValidationRule` in [GraphQLBuilderAuthorizationExtensions](https://github.com/graphql-dotnet/server/blob/develop/src/Authorization.AspNetCore/GraphQLBuilderAuthorizationExtensions.cs) should be changed to `TryAddTransient` to not overwrite a previous injection of a custom class....

rose-a:
The `AddValidationError` method is IMO too complex for easily creating custom error messages ... The user should not have to care about calling `context.ReportError()`...

src/Authorization.AspNetCore/AuthorizationValidationRule.cs
sungam3r: I read your comment several times. It feels like 2 + 2 = 5 )

Shane32:
@rose-a @sungam3r One of you want to pick this up?  I really was just providing a sample; I wish I had more time to spend on this but I don't.  Either just branch from this and make a new PR, or edit this branch; I don't care either way.

Shane32:
Keep in mind that this is currently 100% backwards compatible.  If you want, I could make these changes against the master branch.  Probably not worth it if we are making additional changes and releasing 5.0 soon.

Another idea: we could merge this PR as-is and make a new PR with the additional classes.  Again, whatever you all want is fine.

src/Authorization.AspNetCore/AuthorizationValidationRule.cs
Shane32: Yeah, the xml comments were spur-of-the-moment.  I figured you could tweak them up a bit, but I wanted to jot down something so that you could change them via github code review.

sungam3r:
I'm for v5.

src/Authorization.AspNetCore/AuthorizationValidationRule.cs
sungam3r: I talked about this: 
изображение

I almost fell off my chair.

Shane32: oh - lol - i certainly could be wrong!

Shane32: no surprise, I am wrong!  I felt sure I tried that first, and had to change it.  I must be losing my mind lol

rose-a:
Please add a public property `IEnumerable<IAuthorizationRequirement> FailedRequirements { get }` to `AuthorizationError`.
This could be used inside `IErrorInfoProvider` to "re-generate" the error message... then we'd have a "non-breaking" solution to #478...

rose-a:
Ah forget my latest comment... the `AuthorizationResult` prop is already there ;)

Shane32:
Replaced by #480

PR #493, opened 2021-01-15
Change 500 to 400 http error codes

sungam3r:
Not ideal solution.

See `DocumentExecuter.ExecuteAsync`:
> Cannot execute request if no query is specified

In general, we do not know why this or that error occurred and the client or server is to blame for this. Some cases can be handled on the server side without ever passing control inside GraphQL.NET. For example, this one is just to check the query string for `null` but in general it is impossible to predict all situations. What code should be returned? Obviously not 500.


Shane32 (approved):

Shane32:
The changes here should be reviewed in light of #581 and tests should be added

sungam3r:
Closed in favor of #584 .

PR #521, opened 2021-03-18
Authorization.AspNetCore split proposal

Xelit3:
Hi all,

First, thank you all for reviweing and having a look to this PR, let me expose briefly the idea behind this proposal:

**Basically, in those projects, where GraphQL objects/queries are in a different layer of the Web API, as it could happen in a n-Layered structure, unnecessary AspNetCore dependencies are carried in the case of implementing security.

This proposal is intended to avoid carrying those dependencies in layers that don't require it, when in reality, the only dependency that are strictly required comes from the Core in that case.**

I understand that this PR, if it is accepted according to your most expert criteria, still has pending points to implement, probably  tests or documentation...

If you agree with the idea, I will be happy to add what you may consider neccesary

Thank you all, have a nice day!


sungam3r:
Superseded by #522. See https://graphql-dotnet.github.io/docs/migrations/migration4/#authorization-extension-methods

sungam3r:
Also see https://github.com/graphql-dotnet/graphql-dotnet/pull/2451 . This attribute will be removed from this repo too.

Issue #505, opened 2021-02-17
Extend schema-first authorization policy annotations to work on class properties


infogulch:
Currently, `[GraphQLAuthorize(Policy = "...")]` can only be applied to methods and classes. I propose extending the `GraphQLAuthorizeAttribute` to allow one to apply it to class properties as well.

It turns out you can technically apply `GraphQLAuthorize` specifically to a getter or setter directly (as in `public int MyProperty { [GraphQLAuthorize(Policy="😲")] get; set; }`) , but that didn't magically work how I hoped it would.

I think this feature would be useful because sometimes the level that you want to populate/resolve data for a subquery is not perfectly aligned with the level you want to apply access authorization policies. For example, your backend queries the whole record, but some users should only have permission to access a subset of the fields¹. Yes, this strategy does not utilize 100% of the potential efficiency of graphql, but in some cases that optimization is not possible (e.g. a coarse rest backend), or desired (optimizing for implementation simplicity), etc.

¹ The nearest workaround that I've found is to wrap every record class in a second class that exposes the wrapped class' properties as methods with the same name that simply return the field from the contained record; adding the authorize attribute to those methods then allows one to define access policies. IMO this unnecessary duplication is clearly undesirable.

Thoughts?

sungam3r:
Both `GraphQLAuthorizeAttribute` from this project and https://github.com/graphql-dotnet/authorization inherit `GraphQLAttribute` from GraphQL.NET : 
```c#
 [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = true)]
    public abstract class GraphQLAttribute : Attribute
    {
```

Moreover I can see code like ` public static PropertyInfo PropertyForField(this Type type, string field)` from GraphQL.NET that checks for this attribute applied to property. It looks like we just forgot to allow this attribute to be applied to properties and didn't notice it.

infogulch:
So does this mean that we should change `GraphQLAttribute` to allow properties like this:

```c#
    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method | AttributeTargets.Property, AllowMultiple = true)]
    public abstract class GraphQLAttribute : Attribute
    {
```

Or should `GraphQLAuthorizeAttribute` in this repo be changed to specify these attribute targets?

sungam3r:
`GraphQLAttribute` should be changed.

infogulch:
[...]

PR #542, opened 2021-04-01
Add tests for property authorization & recursive authorization

infogulch:
This PR adds tests that reproduce an issue described in #505.

The PR is broken up into two commits:

* "Add tests for AuthorizeAttribute on properties" - simply adds tests to check that the newly available `GraphQLAuthorizationAttribute` can be applied to properties; these tests succeed as expected.
* "Add tests for object return value authorize attribute" - Adds tests to check if an object returned by a resolver can have *its* properties checked for authorization. As of the opening of this PR the `nested_property_policy_fail` test fails.


sungam3r:
@infogulch Thanks for the repro. See my comments in #505 .

Issue #580, opened 2021-05-20
Missing/invalid JSON causes an unhandled exception 500 error


jeffw-wherethebitsroam:
I set up a test graphql server using:

```
            services.AddSingleton<TestSchema>()
                .AddGraphQL()
                .AddSystemTextJson()
                .AddGraphTypes(typeof(TestSchema));
```

```
            app.UseEndpoints(endpoints =>
            {
                endpoints.MapGraphQL<TestSchema>();
            });
```

If the body of the graphql request is then invalid, I get a 500 response and an error like:

```
      An unhandled exception has occurred while executing the request.
      System.InvalidOperationException: Cannot execute request if no query is specified
         at GraphQL.DocumentExecuter.ExecuteAsync(ExecutionOptions options) in /_/src/GraphQL/Execution/DocumentExecuter.cs:line 66
```

Firstly, shouldn't the middleware return some sort of 400 error? This is clearly a client error.

Secondly, the `System.InvalidOperationException: Cannot execute request if no query is specified` completely obscures the actual error, making it very hard to debug misbehaving clients.

For example, we recently had a problem with a client sending a newline in their JSON, i.e.

```
curl -v https://localhost:5001/graphql -d '{"query": "{ hello\n }"}' -H 'application/json'
```

And the only error reported is `Cannot execute request if no query is specified` which is very misleading.

alilengrai:
UP

sungam3r:
Probably fixed by #493. I have not been engaged in this project for a long time. Closer to winter I plan to return to support.

Shane32:
Should check again - may have already been solved by #581

sungam3r:
Fixed in https://github.com/graphql-dotnet/server/releases/tag/5.1.0

PR #581, opened 2021-05-20
Catch JSON deserialization exceptions and use them to return 400 errors

jeffw-wherethebitsroam:
This code adds the JSON deserialization exception to
GraphQLRequestDeserializationResult and sets IsSuccessful = false.
Also, unmatched content-type should return 415 error code: Unsupported Media Type


jeffw-wherethebitsroam:
This addresses some of the issues from https://github.com/graphql-dotnet/server/issues/580

Shane32 (approved):
Looks good to me, but @sungam3r should review.

tests/Samples.Server.Tests/ResponseTests.cs
```diff
@@ -98,7 +98,7 @@ public async Task Batched_Query_Should_Return_Multiple_Results()
             {
                 HttpMethod.Post,
                 new StringContent(Serializer.ToJson(new GraphQLRequest { Query = "query { __schema { queryType { name } } }" }), Encoding.UTF8, "something/unknown"),
-                HttpStatusCode.BadRequest,
+                HttpStatusCode.UnsupportedMediaType,
```
sungam3r, line 101: > // POST with unsupported mime type should be a bad request

Maybe change this comment ?

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -101,7 +102,12 @@ public async Task InvokeAsync(HttpContext context)
                         break;
 
                     default:
-                        await WriteErrorResponseAsync(httpResponse, writer, cancellationToken, $"Invalid 'Content-Type' header: non-supported media type. Must be of '{MediaType.JSON}', '{MediaType.GRAPH_QL}' or '{MediaType.FORM}'. {DOCS_URL}");
+                        await WriteErrorResponseAsync(
+                            httpResponse,
+                            writer,
+                            cancellationToken,
+                            $"Invalid 'Content-Type' header: non-supported media type. Must be of '{MediaType.JSON}', '{MediaType.GRAPH_QL}' or '{MediaType.FORM}'. {DOCS_URL}",
+                            415); // UnsupportedMediaType
```
sungam3r, line 110: Let's change `httpStatusCode` argument type from `WriteErrorResponseAsync` to `System.Net.HttpStatusCode.BadRequest` and pass enum so no comment is needed. Then cast argument to `int` inside.

src/Transports.AspNetCore.NewtonsoftJson/GraphQLRequestDeserializer.cs
```diff
@@ -44,12 +44,28 @@ public Task<GraphQLRequestDeserializationResult> DeserializeFromJsonBodyAsync(Ht
                 switch (firstChar)
                 {
                     case '{':
-                        result.Single = ToGraphQLRequest(_serializer.Deserialize<InternalGraphQLRequest>(jsonReader));
+                        try
```
sungam3r, line 47: I would not add try/catch inside each deserializer. Let's exception bubble up the stack and catch it in `GraphQLHttpMiddleware` so you don't need new `Exception` property.  Even further. Now I think that `IsSuccessful` property is redundant since we can just throw exception. In this case, API will become simpler - either good deserialized result or exception, no third option via `bool` flag.

sungam3r (changes requested):
Need small changes in the API design. I'm OK to merge after that.

tests/Samples.Server.Tests/ResponseTests.cs
jeffw-wherethebitsroam: Yep, done.

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
jeffw-wherethebitsroam: I've change the status code int to a HttpStatusCode.

src/Transports.AspNetCore.NewtonsoftJson/GraphQLRequestDeserializer.cs
jeffw-wherethebitsroam: The was my first thoughts as well, but _serializer.Deserialize will use either `Newtonsoft.Json` or `System.Text.Json`. Which means we would either have to:
1. Catch both `Newtonsoft.Json.JsonException` and `System.Text.Json.JsonException` in middleware, which makes for a very leaky abstraction.
2. Catch `Exception` and assume it was a a json parsing exception :(
3. Make a wrapper exception, e.g. `GraphQLRequestDeserializerJsonException` to be thrown be both deserializers.

SK201:
Hello,
We are currently affected by this issue and would love to see the PR merged.

Thanks in advance !

jeffw-wherethebitsroam:
I was waiting on feedback on how to propagate the exceptions and then forgot about this. I can update the PR :)

src/Transports.AspNetCore.NewtonsoftJson/GraphQLRequestDeserializer.cs
jeffw-wherethebitsroam: I have implemented 3. using a wrapper exception.

Shane32:
Thanks @jeffw-wherethebitsroam !  Did the API surface change?  We might need the apiapprovals file updated.  Also it looks like it is failing the formatting check.

Shane32:
@jeffw-wherethebitsroam I found the issue with the formatting tool; if you can merge master back into this branch, then it should run the CI tests properly.

jeffw-wherethebitsroam:
@Shane32 I've merged master. The API surface did change (removed IsSuccessful from GraphQLRequestDeserializationResult, added GraphQLRequestDeserializationException), but the apiapprovals file is updated.

jeffw-wherethebitsroam:
@Shane32 It looks like I need a maintainer to approve the workflows: "First-time contributors need a maintainer to approve running workflows"

Shane32:
I think these changes look great!  However, it contains breaking changes in the API surface, and has breaking changes to the semantics of the current request parsing methods.  It really belongs in the next major version, I would think.

To make this a non-breaking change, we could retain the `IsSuccessful` property and add another `Exception` property of type `GraphQLRequestDeserializationException`, or `null` if no error.  This really would perform identically, even though it's not as clean of a design.  Then the decision would become if we should clean up the design for the next major version or not.

The remainder of the semantic changes, such as returning the proper status code upon failure, would only necessitate a bump to the minor version number.

There is one other alternative to getting these changes in the current version, and that would be to leave the `IsSuccessful` property there, but always return `true`, and mark it as obsolete.  The property would need to be monitored in the middleware in case someone is using a custom JSON transport that returns `false` for that property.  I don't really care for this idea.

Thoughts?

jeffw-wherethebitsroam:
Ok, we're now pretty much back to the original version :) Its a bit uglier now since `IsSuccessful = false` and `Exception != null` both mean that the deserialization failed and either or both can be set. But it would be better to clean that up in the next major release.

Shane32 (approved):

Shane32:
I just have so little time to spend on the next version, and there’s a whole lot of cleanup that could be done. Better to get this fix in the current version.

sungam3r:
@jeffw-wherethebitsroam Thanks for this fix. Just reviewed it. LGTM, just a few cosmetic changes [here](#651). I had a long break from OSS this year but I returned. I think that when I read the remaining emails for this repo (~60 left), then I can release v5.1 including this fix (and others of course).

PR #584, opened 2021-05-25
GraphQLHttpMiddleware returns 400 when query is missing

### LINKED ISSUE #583: GraphQL endpoint throwing HTTP 500 on HTTP GET
## Description

The graphQL server throws an InvalidOperationException("Cannot execute request if no query is specified") when HTTP GET is requested on graphql endpoint.

This is a very common scenario (people opening the graphql endpoint in browser). It is causing an internal server error. It would be better if the GraphQL framework could handle this situation. Instead the error is propagated to ASP.NET middleware.

### Steps to reproduce

Open url `http://{server}/graphql` in browser which points to the graphql endpoint.

### Expected result

Bad http request (400)

### Actual result

Unhandled exception is thrown.
Returned Http Response 500.

sharppanda:
Fixes #583

### DIFF
--- a/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
+++ b/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
@@ -121,7 +121,7 @@ namespace GraphQL.Server.Transports.AspNetCore
                     OperationName = urlGQLRequest.OperationName ?? bodyGQLRequest?.OperationName
                 };
 
-                if (string.IsNullOrWhiteSpace(urlGQLRequest.Query))
+                if (string.IsNullOrWhiteSpace(gqlRequest.Query))
                 {
                     await WriteErrorResponseAsync(httpResponse, writer, cancellationToken,
                         "GraphQL query is missing.",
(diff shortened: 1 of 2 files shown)


src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -120,6 +120,15 @@ public async Task InvokeAsync(HttpContext context)
                     Extensions = urlGQLRequest.Extensions ?? bodyGQLRequest?.Extensions,
                     OperationName = urlGQLRequest.OperationName ?? bodyGQLRequest?.OperationName
                 };
+
+                if (string.IsNullOrWhiteSpace(urlGQLRequest.Query))
```
Shane32, line 124:
```suggestion
                if (string.IsNullOrWhiteSpace(gqlRequest.Query))
```

Shane32 (changes requested):

Shane32 (approved):
Looks good to me but @sungam3r needs to approve it.

sharppanda:
Do you have an ETA when this could be reviewed by @sungam3r ?

Shane32:
@sungam3r bump

sungam3r:
@sharppanda Thanks for fixing this. 

> Do you have an ETA when this could be reviewed by @sungam3r ?

I had a long break from OSS this year so...just reviewed! LGTM

PR #590, opened 2021-05-31
Add custom index page support to graphiql

### LINKED ISSUE #589: GraphiQL: support custom index page
We want to modify GraphiQL UI to add user authentication - this is easily done by customizing GraphiQL React components in index.html file.

It would be nice to be able to set custom index.html with additional data model in `GraphiQLOptions` for GraphiQL customization.

I have a small PR to add that support 🙌

whyleee:
Fixes #589

### DIFF
--- a/src/Ui.GraphiQL/GraphiQLOptions.cs
+++ b/src/Ui.GraphiQL/GraphiQLOptions.cs
@@ -1,4 +1,7 @@
+using System;
 using System.Collections.Generic;
+using System.IO;
+using System.Reflection;
 using Microsoft.AspNetCore.Http;
 
 namespace GraphQL.Server.Ui.GraphiQL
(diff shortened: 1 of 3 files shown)


src/Ui.GraphiQL/GraphiQLOptions.cs
```diff
@@ -22,5 +25,16 @@ public class GraphiQLOptions
         /// HTTP headers with which the GraphiQL will be initialized.
         /// </summary>
         public Dictionary<string, string> Headers { get; set; }
+
+        /// <summary>
+        /// Function to get index.html page with GraphiQL setup.
+        /// </summary>
+        public Func<Stream> IndexStream { get; set; } = () => typeof(GraphiQLOptions).GetTypeInfo().Assembly
```
Shane32, line 32:
```suggestion
        public Func<Stream> IndexStream { get; set; } = () => typeof(GraphiQLOptions).Assembly
```

Shane32:
@whyleee I guess I’m sorta wondering why this feature is necessary for an ASP.Net server. The custom index page can be implemented as an ASP.Net page or view extremely easily, and is then very easy to modify or extend, rather having to mess with resource streams.

whyleee:
@Shane32 The idea is you create a custom "cshtml" file as usual, then pass it as a stream to GraphiQL. The file can be bundled into dll or deployed as an ordinary file - this is abstracted away from GraphiQL middleware by using generic `Stream` type. Do you think there is a better way to do this?

I actually got this idea from https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/master/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIOptions.cs#L17..L21 as it looks very flexible and easy to use.

Shane32:
But why go to all that effort? If someone needs a custom page for graphiql, just add the controller-view or “page” to the existing ASP.NET infrastructure. No need to mess with streams or embedded files or anything else.

sungam3r:
@Shane32 You are right about one can easily add custom middleware, copy-paste code, etc. It's funny, but now I needed to do the same for Playground for my work. I believe that it is better to have such an option than not to have. It will not be difficult to me to add such feature to all ui middlewares.

sungam3r:
@whyleee Thanks. Closed in favor of #652. Let's review together.

PR #622, opened 2021-09-24
Potential fix for issue 621

### LINKED ISSUE #621: GraphQL Subscription does not close sockets to Load Balancer
Hello Group, 

Great work with V5.0.  Lots of great improvements.  

Having an issue with the latest WebSockets (Subscriptions).  Prior to V5, #255 was an on-going issue, where we would get these constant disconnects in an aborted status.  This is now gone. However, I have noticed that when behind a load balancer we have some major issues with sockets not being closed on the .Net side.  

image

This is my server running for just a few minutes, as you can see 4,600~ connections are open.  We currently have about 40 people on the site, thus, its not possible.  How did it get so high? Watch. Now, what I will do is force my load balancer to crash, if we did have 4,600 connections, and if .net releases the connections, we would see it drop to ~0.

image

Here is what we get, the existing connections stay open, and an additional ~300 connections are created. Meaning that the connection is not being closed on the .Net side when there is no way the connection can still be open as I crashed the load balancer.  

I can see my DevTools is working correctly to the site (React) is re-creating the connection after the failure of a socket (Reboot).

image

(Yes, I rebooted many times, which is why the socket closes).  Also, If no traffic is sent for 60 seconds, the connection is closed via the load balancer.  This is configurable from the load balancer side. Since the .Net side is not closing the socket, eventually the server crashes with `503.2 Too many connections.`  

Any assistance in a quick resolution would be awesome otherwise, I have to downgrade back to V3.

Shane32:
May fix #621

### DIFF
--- a/src/Transports.Subscriptions.WebSockets/WebSocketTransport.cs
+++ b/src/Transports.Subscriptions.WebSockets/WebSocketTransport.cs
@@ -32,7 +32,7 @@ namespace GraphQL.Server.Transports.WebSockets
 
         public Task CloseAsync()
         {
-            if (_socket.State != WebSocketState.Open)
+            if (_socket.State == WebSocketState.Closed || _socket.State == WebSocketState.Aborted)
                 return Task.CompletedTask;
 
             if (CloseStatus.HasValue)


wpoPR:
it's related to this issue? https://github.com/graphql-dotnet/server/issues/621

Shane32:
We were hoping it would fix #621 but it does not appear to.

sungam3r:
@Shane32 What is next? Close? Rewrite?

Shane32:
Do you think these changes should be merged anyway?  If not, let's close this PR.  We can always re-open it or start a new one if a solution is found.

sungam3r:
I don't know. Just sorting emails. Closing for now.

Issue #504, opened 2021-02-16
[GraphiQL] Add headers


jasonlaw:
Hi,

May I know how to add headers with GraphiQL?

Thank you.

sungam3r:
Most likely, this question is worth asking the GraphiQL authors. I found that [thread](https://github.com/graphql/graphiql/issues/500).

nur858:
@jasonlaw Is GraphiQL absolute necessity for you? Otherwise I found [altair](https://github.com/imolorhe/altair) to be much richer UI with ability to add headers.

jasonlaw:
> @jasonlaw Is GraphiQL absolute necessity for you? Otherwise I found [altair](https://github.com/imolorhe/altair) to be much richer UI with ability to add headers.

Not really, but I found the explorer is very convenient.

dasco144:
> > @jasonlaw Is GraphiQL absolute necessity for you? Otherwise I found [altair](https://github.com/imolorhe/altair) to be much richer UI with ability to add headers.
> 
> Not really, but I found the explorer is very convenient.

There does seem to be a plugin that can add a GraphiQL like explorer into Altair.

![image](https://user-images.githubusercontent.com/10575019/111600762-64e0b380-87da-11eb-90eb-656b0487dba2.png)

sungam3r:
Closed as answered.

PR #649, opened 2021-11-21
Upgrade GraphiQL to 1.5.1

### LINKED ISSUE #544: Subscriptions didnt work out of the box in current examples and readme.md
Maybe I miss something, but I copied the example server over to play with it and subscriptions always raise a not supported exception.

It took me some time until I noticed that this line is missing both in the example server and in the readme.md.

`services.AddSingleton<IDocumentExecuter,SubscriptionDocumentExecuter>();`

### LINKED ISSUE #588: Wrong document executer registered for subscription support
Due to the following code, `DocumentExecuter` is registered instead of `SubscriptionDocumentExecuter`, and there is nothing to override this within `.AddWebSockets()`.

https://github.com/graphql-dotnet/server/blob/4bb8cc3e87a92c179f2af8b2422fe303d51ef0c5/src/Core/Extensions/ServiceCollectionExtensions.cs#L53

This causes the following error to be thrown when attempting to use subscriptions, unless a specific implementation of `IDocumentExecuter` has been registered within the DI framework:

```
DocumentExecuter does not support executing subscriptions. You can use SubscriptionDocumentExecuter from GraphQL.Subscriptions package to execute subscriptions.
```

Current workaround is to register the correct document executer manually:

```cs
services.AddSingleton<IDocumentExecuter, SubscriptionDocumentExecuter>();
```

This is either a bug, or not properly documented and demonstrated in the readme.

See https://github.com/graphql-dotnet/graphql-dotnet/pull/1973

boulc:
# Context

The goal is to add support request headers in GraphiQL (discussed in https://github.com/graphql-dotnet/server/issues/504).

# Description

Request headers support has been added in recent versions of GraphiQL. Unfortunately, [GraphiQL with extensions](https://github.com/OneGraph/graphiql-with-extensions) does not support it. As it wraps the GraphiQL component and masks the props, upgrading has no effect. There is not much activity on graphiql-with-extensions repo since some time now.

The current integration of GraphiQL has a broken support of subscription (see #544), due to the deprecation of `subscribe` function in https://github.com/apollographql/subscriptions-transport-ws/pull/272. Also this dependency is still required though not [maintained anymore](https://the-guild.dev/blog/graphql-over-websockets) until #492 is resolved. For the same reason, the current integration relies on [graphiql-subscriptions-fetcher](https://github.com/apollographql/GraphiQL-Subscriptions-Fetcher) which is archived and not maintained anymore, also wrapping the `GraphQLFetcher` and masking the additional options necessary to support request headers.

# Proposal

1. Upgrading GraphiQL to version 1.5.1: cshtml file is inspired from [CDN example](https://github.com/graphql/graphiql/blob/main/examples/graphiql-cdn/index.html).

2. Adding an option to enable/disable GraphiQL with extensions: `ExplorerExtensionEnabled`, enabled by default for backward compatibility.

3. Adding an option to enable/disable header editor: `HeaderEditorEnabled`, enable by default (same as in GraphiQL).

4. Wrapping `subscriptionsFetcher` to pass additional option to support request headers editor.

5. Downgrade subscriptions-transport-ws to version 0.8.2 to fix #544.

# Tests

Screenshot of side by side example of a mutation on the left leading to a working subscription on the right:

Screenshot 2021-11-20 210857

On the previous screenshot, the request headers editor is visible on the right and the authorization header is available in HTTP context when debugging:

Screenshot 2021-11-20 210339

### DIFF
--- a/samples/Samples.Server/Startup.cs
+++ b/samples/Samples.Server/Startup.cs
@@ -7,6 +7,7 @@ using GraphQL.Server.Ui.Altair;
 using GraphQL.Server.Ui.GraphiQL;
 using GraphQL.Server.Ui.Playground;
 using GraphQL.Server.Ui.Voyager;
+using GraphQL.SystemReactive;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.Configuration;
@@ -34,6 +35,7 @@ namespace GraphQL.Samples.Server
             services.AddSingleton<IChat, Chat>();
 
             MicrosoftDI.GraphQLBuilderExtensions.AddGraphQL(services)
+                .AddSubscriptionDocumentExecuter()
                 .AddServer(true)
                 .AddSchema<ChatSchema>()
                 .ConfigureExecution(options =>
--- a/samples/Samples.Server/Startup.cs
+++ b/samples/Samples.Server/Startup.cs
@@ -43,6 +43,9 @@ namespace GraphQL.Samples.Server
                 .AddWebSockets()
                 .AddDataLoader()
                 .AddGraphTypes(typeof(ChatSchema));
+
+            services
+                .AddSingleton<IDocumentExecuter, SubscriptionDocumentExecuter>();
         }
 
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
--- a/samples/Samples.Server/Startup.cs
+++ b/samples/Samples.Server/Startup.cs
@@ -43,9 +43,6 @@ namespace GraphQL.Samples.Server
                 .AddWebSockets()
                 .AddDataLoader()
                 .AddGraphTypes(typeof(ChatSchema));
-
-            services
-                .AddSingleton<IDocumentExecuter, SubscriptionDocumentExecuter>();
         }
 
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
--- a/src/Ui.GraphiQL/GraphiQLOptions.cs
+++ b/src/Ui.GraphiQL/GraphiQLOptions.cs
@@ -40,6 +40,9 @@ namespace GraphQL.Server.Ui.GraphiQL
         /// Enables the header editor when <c>true</c>.
         /// Not supported when <see cref="ExplorerExtensionEnabled"/> is <see langword="true"/>.
         /// </summary>
+        /// <remarks>
+        /// Original setting from <see href="https://github.com/graphql/graphiql/blob/08250feb6ee8335c3b1ca83a912911ae92a75722/packages/graphiql/src/components/GraphiQL.tsx#L186">GraphiQL</see>.
+        /// </remarks>
         public bool HeaderEditorEnabled { get; set; } = true;
 
         /// <summary>
--- a/src/Ui.GraphiQL/GraphiQLOptions.cs
+++ b/src/Ui.GraphiQL/GraphiQLOptions.cs
@@ -43,7 +43,7 @@ namespace GraphQL.Server.Ui.GraphiQL
         public bool HeaderEditorEnabled { get; set; } = true;
 
         /// <summary>
-        /// Enables the explorer extension when <c>true</c>.
+        /// Enables the explorer extension when <see langword="true"/>.
         /// </summary>
         public bool ExplorerExtensionEnabled { get; set; } = true;
     }
--- a/src/Ui.GraphiQL/GraphiQLOptions.cs
+++ b/src/Ui.GraphiQL/GraphiQLOptions.cs
@@ -38,7 +38,7 @@ namespace GraphQL.Server.Ui.GraphiQL
 
         /// <summary>
         /// Enables the header editor when <c>true</c>.
-        /// Not supported when ExplorerExtensionEnabled is <c>true</c>.
+        /// Not supported when <see cref="ExplorerExtensionEnabled"/> is <see langword="true"/>.
         /// </summary>
         public bool HeaderEditorEnabled { get; set; } = true;
 
--- a/src/Ui.GraphiQL/GraphiQLOptions.cs
+++ b/src/Ui.GraphiQL/GraphiQLOptions.cs
@@ -22,5 +22,16 @@ namespace GraphQL.Server.Ui.GraphiQL
         /// HTTP headers with which the GraphiQL will be initialized.
         /// </summary>
         public Dictionary<string, string> Headers { get; set; }
+
+        /// <summary>
+        /// Enables the header editor when <c>true</c>.
+        /// Not supported when ExplorerExtensionEnabled is <c>true</c>.
+        /// </summary>
+        public bool HeaderEditorEnabled { get; set; } = true;
+
+        /// <summary>
+        /// Enables the explorer extension when <c>true</c>.
+        /// </summary>
+        public bool ExplorerExtensionEnabled { get; set; } = true;
     }
 }
--- a/src/Ui.GraphiQL/GraphiQLOptions.cs
+++ b/src/Ui.GraphiQL/GraphiQLOptions.cs
@@ -37,7 +37,7 @@ namespace GraphQL.Server.Ui.GraphiQL
         public Func<GraphiQLOptions, string, string> PostConfigure { get; set; } = (options, result) => result;
 
         /// <summary>
-        /// Enables the header editor when <c>true</c>.
+        /// Enables the header editor when <see langword="true"/>.
         /// Not supported when <see cref="ExplorerExtensionEnabled"/> is <see langword="true"/>.
         /// </summary>
         /// <remarks>
--- a/src/Ui.GraphiQL/Internal/graphiql.cshtml
+++ b/src/Ui.GraphiQL/Internal/graphiql.cshtml
@@ -68,7 +68,6 @@
   <div id="graphiql">Loading...</div>
   <script
     src="https://unpkg.com/graphiql@1.5.1/graphiql.min.js"
-    type="application/javascript"
   ></script>
   <script>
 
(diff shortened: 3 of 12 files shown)


samples/Samples.Server/Startup.cs
```diff
@@ -43,6 +43,9 @@ public void ConfigureServices(IServiceCollection services)
                 .AddWebSockets()
                 .AddDataLoader()
                 .AddGraphTypes(typeof(ChatSchema));
+
+            services
+                .AddSingleton<IDocumentExecuter, SubscriptionDocumentExecuter>();
```
sungam3r, line 48: Not needed anymore.

boulc: Right, following #653. Thanks.
Didn't it also resolve #588?

Shane32: Looks like it to me.  I'll close that issue.

src/Ui.GraphiQL/GraphiQLOptions.cs
```diff
@@ -35,5 +35,16 @@ public class GraphiQLOptions
         /// Gets or sets a delegate that is called after all transformations of the GraphiQL UI page.
         /// </summary>
         public Func<GraphiQLOptions, string, string> PostConfigure { get; set; } = (options, result) => result;
+
+        /// <summary>
+        /// Enables the header editor when <c>true</c>.
+        /// Not supported when ExplorerExtensionEnabled is <c>true</c>.
```
sungam3r, line 41: Why so ?

src/Ui.GraphiQL/GraphiQLOptions.cs
```diff
@@ -35,5 +35,16 @@ public class GraphiQLOptions
...
+        /// </summary>
+        public bool HeaderEditorEnabled { get; set; } = true;
```
sungam3r, line 43:
```suggestion
        public bool HeadersEditorEnabled { get; set; } = true;
```
?

boulc: I would chose the original naming from graphiql over the plural: https://github.com/graphql/graphiql/blob/08250feb6ee8335c3b1ca83a912911ae92a75722/packages/graphiql/src/components/GraphiQL.tsx#L186

src/Ui.GraphiQL/GraphiQLOptions.cs
boulc: When `ExplorerExtensionsEnabled` is true, GraphiQL with extensions is registered. This is the current existing behavior, which is maintained for backward compatibility.
Unfortunately, GraphiQL with extensions does not support the header editor.
So to enable it, you have to disable GraphiQL with extensions.

src/Ui.GraphiQL/GraphiQLOptions.cs
sungam3r: OK. Please copy link to this "source-settings" into xml comments for GraphiQLOptions.

src/Ui.GraphiQL/GraphiQLOptions.cs
sungam3r: OK

sungam3r, line 41:
```suggestion
        /// Not supported when <see cref="ExplorerExtensionEnabled"/> is <see langword="true"/>.
```

src/Ui.GraphiQL/GraphiQLOptions.cs
```diff
@@ -35,5 +35,16 @@ public class GraphiQLOptions
         /// Gets or sets a delegate that is called after all transformations of the GraphiQL UI page.
         /// </summary>
         public Func<GraphiQLOptions, string, string> PostConfigure { get; set; } = (options, result) => result;
+
+        /// <summary>
+        /// Enables the header editor when <c>true</c>.
```
sungam3r, line 40:
```suggestion
        /// Enables the header editor when <see langword="true"/>.
```

src/Ui.GraphiQL/GraphiQLOptions.cs
```diff
@@ -35,5 +35,16 @@ public class GraphiQLOptions
...
+        /// Not supported when ExplorerExtensionEnabled is <c>true</c>.
+        /// </summary>
+        public bool HeaderEditorEnabled { get; set; } = true;
+
+        /// <summary>
+        /// Enables the explorer extension when <c>true</c>.
```
sungam3r, line 46:
```suggestion
        /// Enables the explorer extension when <see langword="true"/>.
```

boulc:
chore: code review suggestion

Shane32 (approved):

sungam3r:
I need some time to check.

src/Ui.GraphiQL/Internal/graphiql.cshtml
```diff
@@ -8,148 +8,175 @@
...
+    If you do not want to rely on a CDN, you can host these files locally or
+    include them directly in your favored resource bunder.
+  -->
+  <script
+    src="https://cdn.jsdelivr.net/npm/whatwg-fetch@2.0.3/fetch.min.js"
+    integrity="sha384-dcF7KoWRaRpjcNbVPUFgatYgAijf8DqW6NWuqLdfB5Sb4Cdbb8iHX7bHsl9YhpKa"
+    crossorigin="anonymous"
+  ></script>
+  <script
+    src="https://unpkg.com/react@16.14.0/umd/react.production.min.js"
```
sungam3r, line 38: Why do you use SRI in some script/link tags and in some not ?

src/Ui.GraphiQL/Internal/graphiql.cshtml
```diff
@@ -8,148 +8,175 @@
...
+  <script src="https://unpkg.com/graphiql-subscriptions-fetcher@0.0.2/browser/client.js"></script>
 
 </head>
 <body>
-	<div id="graphiql">Loading...</div>
-	<script>
+  <div id="graphiql">Loading...</div>
+  <script
+    src="https://unpkg.com/graphiql@1.5.1/graphiql.min.js"
+    type="application/javascript"
```
sungam3r, line 71: I think that type attribute is deprecated. No?

sungam3r:
@boulc Thanks for the fix. I don't know JS. I roughly understood what was happening here. I have two questions after which I will be glad to merge.

src/Ui.GraphiQL/Internal/graphiql.cshtml
boulc: Most of the code has been retrieved from the CDN example file of graphiql repository: https://github.com/graphql/graphiql/blob/main/examples/graphiql-cdn/index.html and actually there are no SRI in there. And for the other dependencies, I added it whenever possible. I could try to add them everywhere,in this pull request or another. What do you think?

sungam3r: Please add them here. https://www.srihash.org/ helps a lot.

src/Ui.GraphiQL/Internal/graphiql.cshtml
boulc: Actually, the type is defaulted to `text/javascript` when not provided. This stack overflow answer seems the most accurate with references to the HTML specs per version: https://stackoverflow.com/a/4195504. I wonder why the maintainer of graphiql chose `application/javascript` over `text/javascript` 🤔.
RFC 4329 indicates that `text/javascript` mime type is obsklet though: https://datatracker.ietf.org/doc/html/rfc4329#section-7.1
I guess we could remove it, assuming all recent browsers support those specs.

boulc: I will do some tests with various browsers and revert back.

sungam3r: I would remove. If one really needs support for old browsers then he can use `GraphiQLOptions.PostConfigure` to inject/ovewrite any part of html.

Shane32:
@boulc @sungam3r Status?  Waiting on this PR to release 5.2

src/Ui.GraphiQL/Internal/graphiql.cshtml
boulc: I removed it, without any issue 👍

src/Ui.GraphiQL/Internal/graphiql.cshtml
boulc: And done! No issue while testing it. Great tool you provided, thanks.

boulc:
I had some issues with missing `SubscriptionDocumentExecuter` while testing my latest changes so I had to add the dependency to `GraphQL.SystemReactive` to call `AddSubscriptionDocumentExecuter` extension (see 18cfa9d).
Also, I am suggesting another tweak to the sample, changing `sentAt` attribute from Date to DateTime nullable type and defaulting it to DateTime.UtcNow so it changes each time the mutation is executed: in the example below, the subscription reflects the mutation execution more accurately without having to change the mutation input to validate:

Animation

Shane32:
Looks good.  @sungam3r all set?

sungam3r:
Wait a minute.

samples/Samples.Server/Startup.cs
```diff
@@ -34,6 +35,7 @@ public void ConfigureServices(IServiceCollection services)
             services.AddSingleton<IChat, Chat>();
 
             MicrosoftDI.GraphQLBuilderExtensions.AddGraphQL(services)
+                .AddSubscriptionDocumentExecuter()
```
sungam3r, line 38: Examples are "completely" fixed in #693 and will be merged for v6.

sungam3r (approved):

samples/Samples.Server/Startup.cs
boulc: Oh nice! 💯
I anticipate a slight conflict with this addition but nothing major.

sungam3r: Yes. That is why I needed an extra time for review.

PR #656, opened 2021-11-23
Fix content-type bug

sungam3r:
I checked all middlewares in Firefox first. Then I noticed that Playground doesn't work in Chrome. See comments.

### DIFF
--- a/src/Ui.Playground/Internal/PlaygroundPageModel.cs
+++ b/src/Ui.Playground/Internal/PlaygroundPageModel.cs
@@ -27,7 +27,16 @@ namespace GraphQL.Server.Ui.Playground.Internal
                 var headers = new Dictionary<string, object>
                 {
                     ["Accept"] = "application/json",
-                    ["Content-Type"] = "application/json",
+                    // TODO: investigate, fails in Chrome
+                    // {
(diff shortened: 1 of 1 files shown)


sungam3r:
I think that it should be released as 5.1 - just recreate a tag after regenerating nuget token.

Shane32 (approved):

basvandriel:
This still seems to be an issue. Any update on here?

Shane32:
You downloaded version 5.1.0 (released yesterday) and still having an issue @basvandriel  ?

basvandriel:
Yeah! I still get the content-type bug and the 2 prefixed headers.

sungam3r:
Unfortunately this commit is not included in v5.1.0

sungam3r:
Wait a bit. I'll release v5.1.1.

sungam3r:
https://github.com/graphql-dotnet/server/releases/tag/5.1.1

basvandriel:
Did the job! Why is there a default` "Accept": "application/json"` though? Is it nessacery? @sungam3r

sungam3r:
I don't know: https://github.com/graphql-dotnet/server/blob/master/src/Ui.Playground/Internal/PlaygroundPageModel.cs#L30

PR #666, opened 2021-12-09
add GraphQL Playground UI Tabs

leonanpereirapinto:
Hello,

I recently had to add tabs by default on the playground and couldn't find an easy way to do this.
This will allow the Playground UI to show the tabs added by default in the playground options setup.

The result:

image

Thank you.


leonanpereirapinto:
Wrong PR

Issue #667, opened 2021-12-13
Probable memory leak in WebSocketReaderPipeline


GTechene:
Hello,

During a subscription, when a websocket is aborted (e.g. by closing the client-side abruptly), the _socket.State property is set to Aborted in the Complete() method of the WebSocketReaderPipeline. That means the if clause will be completely skipped, meaning the _startBlock will never be completed, leaving a list of objects in memory.

Context : for some reason, our backend has a copy/paste of an old version of GraphQl Server. Our ops team has detected there was a probable memory leak on our backend and my investigations pointed to this ISourceBlock not being completed. I have written a blog post here : https://guillaume.techene.net/2021/12/fixing-a-memory-leak-in-net-core/ 

I have not tried reproducing the scenario in a completely new environment that would use the latest version of GraphQl Server. However, given the code is similar, I suppose the issue is there too. It's probably exploitable so I would advise trying to reproduce/fix quickly.

Thanks for reading !

Shane32:
@GTechene Do you want to take a look at #668 and see if this is what you had in mind?  Thanks for reporting!!

sungam3r:
@GTechene Waiting feedback from you.

GTechene:
@sungam3r Aside from my previous comment 28 minutes ago, what kind of feedback do you need exactly ? The fix looks fine to me, I can't test it properly but it seems OK.

sungam3r:
OK

PR #668, opened 2021-12-14
Patch memory leak with subscriptions when client abruptly terminates

Shane32:
See:
- #667 
- #621

### DIFF
--- a/src/Transports.Subscriptions.WebSockets/WebSocketReaderPipeline.cs
+++ b/src/Transports.Subscriptions.WebSockets/WebSocketReaderPipeline.cs
@@ -42,11 +42,11 @@ namespace GraphQL.Server.Transports.WebSockets
 
         public async Task Complete(WebSocketCloseStatus closeStatus, string statusDescription)
         {
-            if (_socket.State != WebSocketState.Closed &&
+            try
+            {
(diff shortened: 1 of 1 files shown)


Shane32:
@sungam3r I'm going to merge this so it is testable via GitHub Packages.  Please let me know if you have any comments and I can make changes.

src/Transports.Subscriptions.WebSockets/WebSocketReaderPipeline.cs
```diff
@@ -42,26 +42,28 @@ public void LinkTo(ITargetBlock<OperationMessage> target)
...
-                }
-                finally
-                {
-                    _startBlock.Complete();
+                            closeStatus,
+                            statusDescription,
+                            CancellationToken.None);
                 }
+            }
+            finally
```
Shane32, line 63: Basically, the only change here is to move the if condition inside the try block, so the finally block executes regardless of the websocket state.

Shane32: (The finally block terminating the dataflow)

sungam3r:
ok

PR #671, opened 2021-12-20
Dispose of WebSocket when done with it

Shane32:

PR #673, opened 2021-12-20
Better Dispose pattern for WebSocketConnection

Shane32:
Follow-up to PR:
- #671

### DIFF
--- a/src/Transports.Subscriptions.WebSockets/GraphQLWebSocketsMiddleware.cs
+++ b/src/Transports.Subscriptions.WebSockets/GraphQLWebSocketsMiddleware.cs
@@ -63,6 +63,7 @@ namespace GraphQL.Server.Transports.WebSockets
                     var connectionFactory = context.RequestServices.GetRequiredService<IWebSocketConnectionFactory<TSchema>>();
                     using var connection = connectionFactory.CreateConnection(socket, context.Connection.Id);
 
+                    // Wait until the websocket has disconnected (and all subscriptions ended)
                     await connection.Connect();
                 }
             }
--- a/src/Transports.Subscriptions.WebSockets/GraphQLWebSocketsMiddleware.cs
+++ b/src/Transports.Subscriptions.WebSockets/GraphQLWebSocketsMiddleware.cs
@@ -61,7 +61,7 @@ namespace GraphQL.Server.Transports.WebSockets
                 using (_logger.BeginScope($"GraphQL websocket connection: {context.Connection.Id}"))
                 {
                     var connectionFactory = context.RequestServices.GetRequiredService<IWebSocketConnectionFactory<TSchema>>();
-                    var connection = connectionFactory.CreateConnection(socket, context.Connection.Id);
+                    using var connection = connectionFactory.CreateConnection(socket, context.Connection.Id);
 
                     await connection.Connect();
                 }
--- a/src/Transports.Subscriptions.WebSockets/WebSocketConnection.cs
+++ b/src/Transports.Subscriptions.WebSockets/WebSocketConnection.cs
@@ -1,9 +1,10 @@
+using System;
 using System.Threading.Tasks;
 using GraphQL.Server.Transports.Subscriptions.Abstractions;
 
 namespace GraphQL.Server.Transports.WebSockets
 {
-    public class WebSocketConnection
+    public class WebSocketConnection : IDisposable
     {
         private readonly WebSocketTransport _transport;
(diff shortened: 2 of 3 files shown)


src/Transports.Subscriptions.WebSockets/GraphQLWebSocketsMiddleware.cs
```diff
@@ -61,7 +61,7 @@ public async Task InvokeAsync(HttpContext context)
                 using (_logger.BeginScope($"GraphQL websocket connection: {context.Connection.Id}"))
                 {
                     var connectionFactory = context.RequestServices.GetRequiredService<IWebSocketConnectionFactory<TSchema>>();
-                    var connection = connectionFactory.CreateConnection(socket, context.Connection.Id);
+                    using var connection = connectionFactory.CreateConnection(socket, context.Connection.Id);
```
sungam3r, line 64: Why?? We have just created a connection to communicate with client.

sungam3r: Yes, connection is disposable object, of course. But lifetime of connection is much-much longer.

Shane32: Well, with the changes in this PR, it is now disposable.

The idea was that `await connection.Connect();` would wait until all clients disconnected.  But if PR #671 doesn't work at all, then I have to revert that PR and abandon this one as well.  Or figure out what's really going on.

sungam3r: 😞

Shane32: I can't find anything wrong with PR #671 - and everything seems fine with this PR as well.

But what about this change @sungam3r ?

```suggestion
                    var connection = connectionFactory.CreateConnection(socket, context.Connection.Id);
                    context.Response.RegisterForDispose(connection);
```

Shane32: I did a lot of testing, everything seems fine.  `await connection.Connect();` does not return until the connection has terminated.  And I don't think the `WebSocket` really needs to be disposed, because as soon as the ASP.NET Core middleware returns (exits this method), the connection is terminated anyway.  But it seems correct to dispose of the WebSocket in any case, or register it to be disposed.

So I think we should either (a) leave it as it is in this PR where the `using` statement will close the WebSocket once the websocket has finished, or (b) accept this suggestion so that the WebSocket will be disposed when the ASP.NET Core pipeline has finished.

sungam3r: > `await connection.Connect();` does not return until the connection has terminated. 

I don't understand. When does connection terminate? It's subscription, right? It could live for hours/days. If `await connection.Connect();` does not return then `GraphQLWebSocketsMiddleware.InvokeAsync` does not return too (initial HTTP request). Right? Sorry but I did not really understand the websockets yet.

Shane32: > > `await connection.Connect();` does not return until the connection has terminated.
> 
> I don't understand. When does connection terminate? It's subscription, right? It could live for hours/days. If `await connection.Connect();` does not return then `GraphQLWebSocketsMiddleware.InvokeAsync` does not return too (initial HTTP request). Right? Sorry but I did not really understand the websockets yet.

You are exactly correct.  `InvokeAsync` could be pending for "hours/days".  I tested it, and if it returns immediately, ASP.NET Core will terminate the websocket connection immediately also.  So it is required to 'await' here until the websocket is closed.

sungam3r: Worth to add comments. At least for one like me.

src/Transports.Subscriptions.WebSockets/GraphQLWebSocketsMiddleware.cs
```diff
@@ -61,7 +61,7 @@ public async Task InvokeAsync(HttpContext context)
...
 
                     await connection.Connect();
```
Shane32, line 66:
```suggestion
                    // Wait until the websocket has disconnected (and all subscriptions ended)
                    await connection.Connect();
```

sungam3r (approved):

PR #675, opened 2021-12-25
Multi-TFM approvals

sungam3r:
Waiting for #665

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,5 +1,4 @@
-test:
-  - tests/**/*
+test: ['tests/**/*', '!tests/ApiApprovalTests/**/*.txt']
 
 CI:
   - .github/workflows/**/*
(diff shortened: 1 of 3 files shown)


Shane32:
The code seems to be working fairly well in the GraphQL.NET project.  One thing: right now when one TFM fails, it won't run the other TFM check.  So it is necessary to run the api tests, rename the one file, and then run the tests again and rename the other file.  Can we have it compare/update both files before returning a failure to the test?  For example, start two tasks with Task.Run and then after a WhenAll then await one and the other or something?

sungam3r:
> One thing: right now when one TFM fails, it won't run the other TFM check. 

I know 😄 . Already thought about it.

sungam3r:
https://github.com/graphql-dotnet/graphql-dotnet/pull/2818

sungam3r:
Ported from GraphQL.NET

Shane32 (approved):

PR #677, opened 2021-12-26
Dispose all subscriptions when connection broken

Shane32:
This code ensures that all existing subscriptions are disposed when the connection is disposed.  Of course if the shutdown is clean then there should be no difference. Also it prevents additional subscriptions from connecting after the connection is disposed. Also the subscription releases all held objects upon disposal - so in case the IObserverable does not properly release the observer after it is disposed, at least there won't be any held handles to the underlying WebSocket.

### DIFF
--- a/src/Transports.Subscriptions.Abstractions/SubscriptionManager.cs
+++ b/src/Transports.Subscriptions.Abstractions/SubscriptionManager.cs
@@ -18,6 +18,7 @@ namespace GraphQL.Server.Transports.Subscriptions.Abstractions
 
         private readonly ILogger<SubscriptionManager> _logger;
         private readonly ILoggerFactory _loggerFactory;
+        private volatile bool _disposed;
 
         private readonly ConcurrentDictionary<string, Subscription> _subscriptions =
             new ConcurrentDictionary<string, Subscription>();
@@ -45,13 +46,18 @@ namespace GraphQL.Server.Transports.Subscriptions.Abstractions
                 throw new ArgumentNullException(nameof(payload));
             if (context == null)
                 throw new ArgumentNullException(nameof(context));
+            if (_disposed)
+                throw new ObjectDisposedException(nameof(SubscriptionManager));
 
             var subscription = await ExecuteAsync(id, payload, context).ConfigureAwait(false);
 
             if (subscription == null)
                 return;
 
-            _subscriptions[id] = subscription;
+            if (_disposed)
+                subscription.Dispose();
+            else
+                _subscriptions[id] = subscription;
         }
 
         /// <inheritdoc />
@@ -71,6 +77,9 @@ namespace GraphQL.Server.Transports.Subscriptions.Abstractions
             OperationMessagePayload payload,
             MessageHandlingContext context)
         {
+            if (_disposed)
+                throw new ObjectDisposedException(nameof(SubscriptionManager));
+
             var writer = context.Writer;
             _logger.LogDebug("Executing operation: {operationName} query: {query}",
                 payload.OperationName,
@@ -143,6 +152,7 @@ namespace GraphQL.Server.Transports.Subscriptions.Abstractions
 
         public void Dispose()
         {
+            _disposed = true;
             while (_subscriptions.Count > 0)
             {
                 var subscriptions = _subscriptions.ToArray();
@@ -159,7 +169,7 @@ namespace GraphQL.Server.Transports.Subscriptions.Abstractions
                             _logger.LogError($"Failed to dispose subscription '{subscriptionPair.Key}': ${ex}");
                         }
                     }
-]                }
+                }
             }
         }
     }
--- a/src/Transports.Subscriptions.Abstractions/SubscriptionManager.cs
+++ b/src/Transports.Subscriptions.Abstractions/SubscriptionManager.cs
@@ -147,7 +147,7 @@ namespace GraphQL.Server.Transports.Subscriptions.Abstractions
             return null;
         }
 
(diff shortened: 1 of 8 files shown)


sungam3r:
approvals?

Shane32:
> approvals?

Whoops!  Should be good now.

Shane32:
@sungam3r Merge if it looks good to you.

Shane32:
@sungam3r Ready for review here.

src/Transports.Subscriptions.Abstractions/SubscriptionServer.cs
```diff
@@ -136,5 +137,7 @@ private void LinkToTransportWriter()
             TransportWriter = Transport.Writer;
             _logger.LogDebug("Writer pipeline created");
         }
+
+        public void Dispose() => (Subscriptions as IDisposable)?.Dispose();
```
sungam3r, line 141: It's not a "DI-way", but OK for now.

src/Transports.Subscriptions.WebSockets/WebSocketConnection.cs
```diff
@@ -26,6 +26,7 @@ public virtual async Task Connect()
 
         public void Dispose()
         {
+            _server.Dispose();
```
sungam3r, line 29: Does `WebSocketConnection` owns `SubscriptionServer` ? Are you sure?

src/Transports.Subscriptions.Abstractions/SubscriptionManager.cs
```diff
@@ -45,13 +46,18 @@ public SubscriptionManager(IGraphQLExecuter executer, ILoggerFactory loggerFacto
...
+            if (_disposed)
+                throw new ObjectDisposedException(nameof(SubscriptionManager));
 
             var subscription = await ExecuteAsync(id, payload, context).ConfigureAwait(false);
 
             if (subscription == null)
                 return;
 
-            _subscriptions[id] = subscription;
+            if (_disposed)
```
sungam3r, line 57: Concurrent dispose?

src/Transports.Subscriptions.Abstractions/SubscriptionManager.cs
```diff
@@ -140,5 +146,28 @@ public Task UnsubscribeAsync(string id)
 
             return null;
         }
+
+        public virtual void Dispose()
```
sungam3r, line 150: Who calls this method?

src/Transports.Subscriptions.WebSockets/WebSocketConnection.cs
Shane32: Yes. Check the factory. It’s created by the factory and passed to the connection. Perhaps a poor design but that’s what it is in this version.

Shane32: (On the road now so I’m going from memory…)

src/Transports.Subscriptions.Abstractions/SubscriptionManager.cs
Shane32: The connection can disconnect during ExecuteAsync, so another check is needed

src/Transports.Subscriptions.Abstractions/SubscriptionServer.cs
Shane32: This is calling SubscriptionManager.Dispose if I remember right

Shane32:
Just checked the code:

`WebSocketConnectionFactory.CreateConnection`:
- Creates `WebSocketTransport` and `SubscriptionManager`.
  - The `WebSocketTransport` disposes the underlying WebSocket
  - The `SubscriptionManager` disposes any `Subscription`s it creates in a thread-safe manner.
    - `Subscription` instances dispose the underlying `IDisposable` returned from subscribing
- Creates `SubscriptionServer` which owns (and disposes) the `SubscriptionManager` and references the `WebSocketTransport`.
- Creates `WebSocketConnection` which owns (and disposes) `WebSocketTransport` and `SubscriptionServer`.
- The `WebSocketConnection` is disposed by the `using` statement in the middleware for ASP.Net Core (`GraphQLWebSocketsMiddleware`).

This may sound confusing, but I'm pretty sure it's all correct.

src/Transports.Subscriptions.WebSockets/WebSocketConnection.cs
Shane32: Double checked: yes it does.

src/Transports.Subscriptions.Abstractions/SubscriptionServer.cs
Shane32: Confirmed.

src/Transports.Subscriptions.Abstractions/SubscriptionManager.cs
Shane32: SubscriptionServer.Dispose

src/Transports.Subscriptions.WebSockets/WebSocketConnection.cs
sungam3r: OK

src/Transports.Subscriptions.Abstractions/SubscriptionManager.cs
sungam3r: OK. I still do not know this code.

sungam3r (approved):

PR #693, opened 2021-12-30
Remove obsolete methods

sungam3r:


sungam3r:
CodeQL doesn't work because of preview packages. I think it's OK.

sungam3r:
I would remove now, that is for v6.

Shane32 (approved):
Not sure why I never reviewed this before, but let's merge it in.

Shane32:
Looks like the code coverage isn't working on Windows.  Maybe there's some type of change to the build process that it only generates code coverage reports on Linux?

sungam3r:
As intended
изображение

sungam3r:
Prev PR:
изображение

sungam3r:
It was a transient error.

Shane32:
> It was a transient error.

Good. I ran the workflow twice and it failed both times. Glad it’s working now.

PR #694, opened 2021-12-31
[Experimental] Added support for graphql-transport-ws

fallenwood:
Add switchable support for `graphql-transport-ws` subprotocol defined at https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md.

I have tested locally with a static html client to verify the subscription flow worked, but I am not sure if there is any other scenario that not supported.

Related to #492 

<details>
<summary>The simple client</summary>

```html
<!DOCTYPE html>
<html>
<head>
  <title>ServerSentEvents</title>
  <script
    type="text/javascript"
    src="//unpkg.com/graphql-ws/umd/graphql-ws.min.js>
  </script>
</head>
<body>
  <div>Subscribe Messages</div>
  <div id="messages"></div>
  <script type="text/javascript">
    const messages = document.querySelector("#messages");
    const client = graphqlWs.createClient({
      url: 'ws://127.0.0.1:60341/graphql',
    });

    const payload = `
subscription MessageAdded {
  messageAdded {
    from { id displayName }
    content
  }
}
`;

    client.subscribe({
      query: payload,
     }, {
      next: (data) => {
        const div = document.createElement("div");
        div.innerHTML = JSON.stringify(data);
        messages.appendChild(div);
      },
      error: err => alert("Error! " + err),
      complete: () => alert("Completed"),
    });

  </script>
</body>
</html>
```
</details>

### DIFF
--- a/samples/Samples.Server/Startup.cs
+++ b/samples/Samples.Server/Startup.cs
@@ -36,6 +36,8 @@ namespace GraphQL.Samples.Server
             MicrosoftDI.GraphQLBuilderExtensions.AddGraphQL(services)
                 .AddServer(true)
                 .AddSchema<ChatSchema>()
+                // Required for subscriptions
+                .AddDocumentExecuter<SubscriptionDocumentExecuter>()
                 .ConfigureExecution(options =>
                 {
                     options.EnableMetrics = Environment.IsDevelopment();
@@ -58,7 +60,8 @@ namespace GraphQL.Samples.Server
 
             app.UseWebSockets();
 
-            app.UseGraphQLWebSockets<ChatSchema>();
+            // Note: the new GraphQLTransportWs is not compatible with Playground
+            app.UseGraphQLWebSockets<ChatSchema>(subprotocol: GraphQL.Server.Transports.WebSockets.WebSocketsSubprotocol.GraphQLWs);
             app.UseGraphQL<ChatSchema, GraphQLHttpMiddlewareWithLogs<ChatSchema>>();
 
             app.UseGraphQLPlayground(new PlaygroundOptions
--- a/src/Transports.AspNetCore.SystemTextJson/OperationMessageConverter.cs
+++ b/src/Transports.AspNetCore.SystemTextJson/OperationMessageConverter.cs
@@ -20,14 +20,20 @@ namespace GraphQL.Server
         {
             writer.WriteStartObject();
 
-            writer.WritePropertyName("id");
-            writer.WriteStringValue(message.Id);
+            if (message.Id != null)
+            {
+                writer.WritePropertyName("id");
+                writer.WriteStringValue(message.Id);
+            }
 
             writer.WritePropertyName("type");
             writer.WriteStringValue(message.Type);
 
-            writer.WritePropertyName("payload");
-            JsonSerializer.Serialize(writer, message.Payload, options);
+            if (message.Payload != null)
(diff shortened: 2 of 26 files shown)


sungam3r:
@fallenwood Thanks for PR. Honestly, I do not think this is the right time for such PR. This project requires refactoring, simplification, cleaning everything unnecessary around subscriptions. PR brings additional complication, the need to support two protocols. Also note that all PRs with any backward incompatible changes should be targeted on `develop` branch.

boulc:
Hey @sungam3r, is there any issue that could be worked on to refactor / simplify and get the code base ready for such a change?

I started also a similar [initiative](https://github.com/graphql-dotnet/server/compare/master...boulc:feat/graphql-ws?expand=1) some weeks ago and took the same approach as @fallenwood work. We could all benefit from supporting a more recent and supported subscription protocol.

sungam3r:
I am not ready to give certain advice. I only know that all code working with subscriptions requires refactoring (moving code around projects, changing dependencies, abstractions, making new projects, etc.). I was going to do this several times, but retreat. I do not use subscriptions, although I have some idea of how they work.

Shane32:
@fallenwood @boulc

I've been rewriting subscription support from scratch over the last week.  It's probably 50% complete.  It will not require a dependency on `System.Reactive` (see https://github.com/graphql-dotnet/graphql-dotnet/pull/2757 ) or `System.Threading.Tasks.Dataflow` (progress in branch `new_subscription_support`) anymore, although you can continue to use them for your own code if desired.

### NOTE: I could really use assistance testing the new code - when it's ready - since I don't use subscriptions in my own codebases.

Good news is that the proper members are marked virtual in v5.2 so that you can completely replace the subscription implementation with your own implementation.  So what I may do (we will see) is create a separate project with the new code for testing, so you can copy the new project into your own codebase and 'plug it into' your existing code by a simple DI registration.  Perhaps I would even release a beta nuget package, but I hope not to do so.

If all is successful, this would all be available for all Server v5.2+ users.  This does not cover refactoring the entire project's subscription support, which badly needs to be done, but is waiting on https://github.com/graphql-dotnet/graphql-dotnet/pull/2760 .  The refactoring changes would need to go into the next major version of GraphQL and Server (v5 and v6 respectively).

Shane32:
@fallenwood I really appreciate this PR, even if it is not ultimately merged, so I can see the changes needed for the new transport.  I think we may still consider merging this PR for v5.3 if it does not cause breaking changes to existing clients.  I would need to spend time to perform a careful review.  (I think for now I will be spending my time to continue my rewriting efforts.)  Please leave the PR open for now - thanks!

fallenwood:
Really appreciate your comments, I'll change the target to develop (although it's indeed compatible) and leave this one open.

It's not urgent for me - in my project, some other transport is needed because websockets is not supported, and this PR is a side project for me to understand the subscription workflow :) - so I really looking forward to see the new implementation.

Thanks

sungam3r:
> although it's indeed compatible

See tests/ApiApprovalTests/*.approved.txt changes in this PR. You added new arguments in public ctors and did some other things.

fallenwood:
> > although it's indeed compatible
> 
> See tests/ApiApprovalTests/*.approved.txt changes in this PR. You added new arguments in public ctors and did some other things.

Ah, yes, you are right

PR #695, opened 2022-01-01
Change labeler rules

sungam3r:


Shane32 (approved):
Whatever you want.  The labeler ALWAYS labels things incorrectly for me.  For example, why would I want a "test" label for a new feature?  Of course a new feature would have a test to go with it - a test label should be for test-specific PRs.  But it's set up so most PRs gets tagged with test regardless of the PR's purpose.  So then entire labeler should be removed in my opinion.  But if you like it, 🤷‍♂️

PR #704, opened 2022-01-04
Tune dependabot

sungam3r:


sungam3r:
@Shane32 OK for ignoring these deps ?

Shane32:
Yes

PR #716, opened 2022-02-05
Use new (de)serialization code from main project

Shane32:
Removes SystemTextJson and NewtonsoftJson libraries from this codebase.


Shane32:
Two tests fail, which should be reviewed.

tests/Samples.Server.Tests/ResponseTests.cs
```diff
@@ -103,7 +104,7 @@ public async Task Batched_Query_Should_Return_Multiple_Results()
                 HttpMethod.Post,
                 new StringContent("Oops", Encoding.UTF8, "application/json"),
                 HttpStatusCode.BadRequest,
-                "JSON body text could not be parsed. Body text should start with '{' for normal graphql query or with '[' for batched query."
+                "JSON body text could not be parsed. 'O' is an invalid start of a value. Path: $ | LineNumber: 0 | BytePositionInLine: 0."
```
Shane32, line 107: This error message (when parsing "Oops") is now just what was returned directly from the System.Text.Json deserializer instead of a custom message.

tests/Samples.Server.Tests/ResponseTests.cs
```diff
@@ -115,14 +116,23 @@ public async Task Batched_Query_Should_Return_Multiple_Results()
...
             },
 
+            // POST with JSON mime type that is null JSON should be a bad request
+            new object[]
+            {
+                HttpMethod.Post,
+                new StringContent("null", Encoding.UTF8, "application/json"),
+                HttpStatusCode.BadRequest,
+                "GraphQL query is missing."
+            },
```
Shane32, line 126: I added an additional test here in case the string `"null"` was passed in, mainly to be sure there is no null reference exception.  Note that the string `"{}"` would have the same effect.

tests/Samples.Server.Tests/Serializer.cs
```diff
@@ -29,17 +30,17 @@ internal static FormUrlEncodedContent ToFormUrlEncodedContent(GraphQLRequest req
 
             if (request.OperationName != null)
             {
-                dictionary[GraphQLRequest.OPERATION_NAME_KEY] = request.OperationName;
+                dictionary["operationName"] = request.OperationName;
```
Shane32, line 33: Here in the tests also we see another reference to the JSON keys, if you will.  As these are tests, I just embedded them directly in the source, which is probably as it should be anyway.

src/Transports.Subscriptions.WebSockets/WebSocketTransport.cs
```diff
@@ -3,26 +3,18 @@
...
-        public WebSocketTransport(WebSocket socket, IGraphQLSerializer serializer)
+        public WebSocketTransport(WebSocket socket, IGraphQLTextSerializer serializer)
         {
             _socket = socket;
-            var serializerSettings = new JsonSerializerSettings
-            {
-                DateFormatHandling = DateFormatHandling.IsoDateFormat,
-                DateFormatString = "yyyy'-'MM'-'dd'T'HH':'mm':'ss.FFFFFFF'Z'",
-                ContractResolver = new CamelCasePropertyNamesContractResolver()
-            };
```
Shane32, line 23: GraphQL.NET formats dates as strings internally now, so this code doesn't matter.

tests/Transports.AspNetCore.Tests/NewtonsoftJsonTests.cs
```diff
@@ -41,64 +41,57 @@ public async Task Decodes_BigInteger()
...
+            ret.Single().Variables["a"].ShouldBeOfType<BigInteger>().ShouldBe(bi);
         }
 
         [Fact]
         public async Task Dates_Should_Parse_As_Text()
         {
             var ret = await Deserialize(@"{""variables"":{""date"":""2015-12-22T10:10:10+03:00""}}");
-            ret.Single.Inputs["date"].ShouldBeOfType<string>().ShouldBe("2015-12-22T10:10:10+03:00");
+            ret.Single().Variables["date"].ShouldBeOfType<string>().ShouldBe("2015-12-22T10:10:10+03:00");
         }
```
Shane32, line 52: This is not the default behavior by the Newtonsoft.Json serializer.

tests/Transports.AspNetCore.Tests/NewtonsoftJsonTests.cs
```diff
@@ -41,64 +41,57 @@ public async Task Decodes_BigInteger()
...
             {
-                Body = jsonStream
-            };
-            var deserializer = new NewtonsoftJson.GraphQLRequestDeserializer(_ => { });
-            return await deserializer.DeserializeFromJsonBodyAsync(httpRequest);
+                config.DateFormatHandling = Newtonsoft.Json.DateFormatHandling.IsoDateFormat;
+                config.DateParseHandling = Newtonsoft.Json.DateParseHandling.None;
+            });
+            return await deserializer.ReadAsync<GraphQLRequest[]>(jsonStream);
         }
```
Shane32, line 95: To use date parsing as it was before, `DateFormatHandling` and `DateParseHandling` must be set.

Should we change the above test to simply verify the serializer's default configuration and remove these lines here?

tests/Transports.AspNetCore.Tests/NewtonsoftJsonTests.cs
```diff
@@ -41,64 +41,57 @@ public async Task Decodes_BigInteger()
...
-            ret.Single.Extensions.ShouldBeNull();
-        }
-
-        [Fact]
-        public async Task Name_Matching_Not_Case_Sensitive()
-        {
-            var ret = await Deserialize(@"{""VARIABLES"":{""date"":""2015-12-22T10:10:10+03:00""}}");
-            ret.Single.Inputs["date"].ShouldBeOfType<string>().ShouldBe("2015-12-22T10:10:10+03:00");
+            ret.Single().Extensions.ShouldBeNull();
         }
```
Shane32, line 59: I removed this test.  Deserializing a `GraphQLRequest` object is always case-sensitive now, even when the underlying serializer is configured to be case-insensitive.

We could test what happens when deserializing an unrecognized object type, and then we would see that it is not case sensitive.  Or we could make this match the System.Text.Json serializer which is called "Name_Matching_Is_Case_Sensitive".

tests/Transports.AspNetCore.Tests/SystemTextJsonTests.cs
```diff
@@ -41,64 +41,60 @@ public async Task Decodes_BigInteger()
...
         }
 
         [Fact]
         public async Task Name_Matching_Is_Case_Sensitive()
         {
-            var ret = await Deserialize(@"{""VARIABLES"":{""date"":""2015-12-22T10:10:10+03:00""}}");
-            ret.Single.Inputs.ShouldBeNull();
+            await Should.ThrowAsync<System.Text.Json.JsonException>(()
+                => Deserialize(@"{""VARIABLES"":{""date"":""2015-12-22T10:10:10+03:00""}}"));
         }
```
Shane32, line 66: Now our code throws a deserialization error when an unrecognized key is found.  We can change this behavior in the main project if desired.

tests/Transports.Subscriptions.Abstractions.Tests/ProtocolHandlerFacts.cs
```diff
@@ -231,5 +233,12 @@ public async Task Receive_unknown()
             Assert.Contains(_transportWriter.WrittenMessages,
                 message => message.Type == MessageType.GQL_CONNECTION_ERROR);
         }
+
+        private JObject FromObject(object value)
+        {
+            var serializer = new GraphQLSerializer();
+            var data = serializer.Serialize(value);
+            return serializer.Deserialize<JObject>(data);
+        }
```
Shane32, line 242: Because `GraphQLRequest` is not tagged with the Newtonsoft-specific attributes to indicate the proper key name, a serializer must be used.  (We do not provide a method to serialize to a `JObject`.)

tests/Transports.Subscriptions.WebSockets.Tests/WebSocketWriterPipelineTests.cs
```diff
@@ -39,7 +39,7 @@ public WebSocketWriterPipelineFacts()
                             }
                         }
                     },
-                    83
+                    95
```
Shane32, line 42: Since `type` is always serialized (a required field), the string `"type":null,` is added to most of these tests that didn't supply a type to the serialized data.  `"type":null,` is 12 characters, so I needed to add 12 here.

tests/Transports.AspNetCore.Tests/NewtonsoftJsonTests.cs
sungam3r: _not_ or _now_ ?

src/Benchmarks/Benchmarks/DeserializeFromJsonBodyBenchmark.cs
```diff
@@ -42,40 +39,37 @@ public void GlobalSetup()
...
             // Reset stream positions
-            _httpRequest.Body.Position = 0;
-            _httpRequest2.Body.Position = 0;
+            _httpRequestBody.Position = 0;
+            _httpRequestBody2.Position = 0;
         }
 
         [Benchmark(Baseline = true)]
-        public Task<GraphQLRequestDeserializationResult> NewtonsoftJson() => _nsjDeserializer.DeserializeFromJsonBodyAsync(_httpRequest);
+        public ValueTask<GraphQLRequest[]> NewtonsoftJson() => _nsjDeserializer.ReadAsync<GraphQLRequest[]>(_httpRequestBody);
```
sungam3r, line 64: I suggest to merge this file into `DeserializationBenchmark` from main repo and completely delete file from here. Nothing ASP.NET related to bench here.

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -31,12 +32,12 @@ public class GraphQLHttpMiddleware<TSchema>
...
         private readonly RequestDelegate _next;
-        private readonly IGraphQLRequestDeserializer _deserializer;
+        private readonly IGraphQLTextSerializer _serializer;
 
-        public GraphQLHttpMiddleware(RequestDelegate next, IGraphQLRequestDeserializer deserializer)
+        public GraphQLHttpMiddleware(RequestDelegate next, IGraphQLTextSerializer deserializer)
         {
             _next = next;
-            _deserializer = deserializer;
+            _serializer = deserializer;
```
sungam3r, line 40: изображение
:)

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -31,12 +32,12 @@ public class GraphQLHttpMiddleware<TSchema>
         private const string DOCS_URL = "See: http://graphql.org/learn/serving-over-http/.";
 
         private readonly RequestDelegate _next;
-        private readonly IGraphQLRequestDeserializer _deserializer;
+        private readonly IGraphQLTextSerializer _serializer;
 
-        public GraphQLHttpMiddleware(RequestDelegate next, IGraphQLRequestDeserializer deserializer)
+        public GraphQLHttpMiddleware(RequestDelegate next, IGraphQLTextSerializer deserializer)
```
sungam3r, line 37:
```suggestion
        public GraphQLHttpMiddleware(RequestDelegate next, IGraphQLTextSerializer serializer)
```

src/Transports.Subscriptions.Abstractions/ISubscriptionManager.cs
```diff
@@ -15,7 +16,7 @@ public interface ISubscriptionManager : IEnumerable<Subscription> //todo: add ID
         /// <param name="payload"></param>
         /// <param name="context"></param>
         /// <returns></returns>
-        Task SubscribeOrExecuteAsync(string id, OperationMessagePayload payload, MessageHandlingContext context);
```
sungam3r, line 18: Wow, it was GraphQLRequest.

src/Transports.Subscriptions.Abstractions/Transports.Subscriptions.Abstractions.csproj
```diff
@@ -7,7 +7,6 @@
 
   <ItemGroup>
     <PackageReference Include="GraphQL.SystemReactive" Version="5.0.0-preview-452" />
-    <PackageReference Include="GraphQL.NewtonsoftJson" Version="5.0.0-preview-452" />
```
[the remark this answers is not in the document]
sungam3r: 👍

src/Transports.Subscriptions.WebSockets/WebSocketTransport.cs
sungam3r: ok

tests/Transports.AspNetCore.Tests/NewtonsoftJsonTests.cs
sungam3r: I'm almost sure yes for `DateFormatHandling`. `DateParseHandling` is a PITA for many developers, the "main shame" of Newtonsoft.

> Should we change the above test to simply verify the serializer's default configuration and remove these lines here?

Yes, it will be better to ensure that the default configuration just works.

tests/Transports.AspNetCore.Tests/NewtonsoftJsonTests.cs
sungam3r: >  Or we could make this match the System.Text.Json serializer which is called "Name_Matching_Is_Case_Sensitive".

OK. Catch exception and verify message.

tests/Transports.AspNetCore.Tests/SystemTextJsonTests.cs
sungam3r: OK. Also check exception message.

sungam3r:
Don't forget to remove all references to removed packages from README.

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -90,17 +91,21 @@ public async Task InvokeAsync(HttpContext context)
...
-                        if (!deserializationResult.IsSuccessful)
+                        GraphQLRequest[] deserializationResult;
+                        try
                         {
-                            var message = deserializationResult.Exception is null
-                                ? "JSON body text could not be parsed."
-                                : $"JSON body text could not be parsed. {deserializationResult.Exception.Message}";
+                            deserializationResult = await _serializer.ReadAsync<GraphQLRequest[]>(httpRequest.Body, cancellationToken);
+                        }
+                        catch (Exception ex)
```
sungam3r, line 99: Exception is swallowed here. Let's add `protected virtual bool HandleDeserializationError(Exception e) => true`
And here (before return):
```csharp
if (!HandleDeserializationError(ex))
    throw;
```

sungam3r: Also `GraphQLHttpMiddlewareWithLogs` is a good example to override `HandleDeserializationError` to write log.

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -90,17 +91,21 @@ public async Task InvokeAsync(HttpContext context)
...
                             await WriteErrorResponseAsync(httpResponse, serializer, cancellationToken, message, HttpStatusCode.BadRequest);
                             return;
                         }
-                        bodyGQLRequest = deserializationResult.Single;
-                        bodyGQLBatchRequest = deserializationResult.Batch;
+                        if (deserializationResult?.Length == 1)
+                            bodyGQLRequest = deserializationResult[0];
+                        else
+                            bodyGQLBatchRequest = deserializationResult;
                         break;
```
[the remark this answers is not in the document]
sungam3r: 100% needed later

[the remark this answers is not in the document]
sungam3r: #717

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -239,20 +244,25 @@ private Task WriteResponseAsync<TResult>(HttpResponse httpResponse, IGraphQLSeri
             return serializer.WriteAsync(httpResponse.Body, result, cancellationToken);
         }
 
+        private const string QUERY_KEY = "query";
+        private const string VARIABLES_KEY = "variables";
+        private const string EXTENSIONS_KEY = "extensions";
+        private const string OPERATION_NAME_KEY = "operationName";
```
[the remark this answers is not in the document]
sungam3r: ok

src/Transports.AspNetCore/GraphQLRequestDeserializationException.cs
```diff
@@ -1,32 +0,0 @@
...
-        {
-        }
-
-        /// <summary>
-        /// Creates an instance of <see cref="GraphQLRequestDeserializationException"/> for a situations
-        /// when the first symbol of JSON body neither '{' nor '['.
-        /// </summary>
-        public static GraphQLRequestDeserializationException InvalidFirstChar()
-        {
-            return new GraphQLRequestDeserializationException("Body text should start with '{' for normal graphql query or with '[' for batched query.");
```
sungam3r, line 29: So in case of error now we have some "native" exception from serializer, right?

src/Benchmarks/Benchmarks/DeserializeFromJsonBodyBenchmark.cs
Shane32: #718

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
Shane32: Done

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
Shane32: Done

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
Shane32: I don't mind but I don't understand how that makes sense.  If it fails to parse the JSON, execution cannot continue.  There is no query to execute.  So it is not possible to "handle" a JSON exception.

I'll add a void protected method instead.

Shane32: Oh wait so handle true makes it return a failure instead of returning an error?  I'm still just as confused - how does it make sense to rethrow the exception?

Still going to add a void method.

sungam3r: 1) `HandleDeserializationError` as it is can log exception, for example.
2) Next, if `HandleDeserializationError` returns `bool` then one can decide whether to throw and return some other HTTP related result or just return _standard_ result. Imagine `UseDeveloperExceptionPage` somewhere in the request pipeline. Now any deserialization exception is swallowed, we return 400 code, the end. I want to add the possibility to bubble occurred exception up.

Shane32: #719

tests/Transports.AspNetCore.Tests/SystemTextJsonTests.cs
Shane32: Done

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
sungam3r: ok

src/Transports.AspNetCore/GraphQLRequestDeserializationException.cs
Shane32: Right.

Shane32:
> maybe I was too fast suggesting remove serializer settings

Whatever the test shows, and whether or not they use the server project, the proper way to configure Newtonsoft.Json is to disable decoding of date/time values.  Let me know if you want the test changed back.

Shane32:
> Don't forget to remove all references to removed packages from README.

Updated

tests/Transports.AspNetCore.Tests/SystemTextJsonTests.cs
sungam3r: 🤔 Actually I thought about such behaviour one more time and I'm not sure now.
изображение

Why so in both serializers? This behavior is not flexible. When I come across this in practice, it does not cause anything but bewilderment. Unknown fields should be ignored.

```json
{
"query": ...
"operationName": ...
"variables": ....,
"extensions": ....
}
```

works, but

```json
{
"query": ...
"operationName": ...
"variables": ....,
"extensions": ....,
"some_property_for_tracing_or_whatever": ...
}

does not. Why?
```

sungam3r:
> whether or not they use the server project, the proper way to configure Newtonsoft.Json is to disable decoding of date/time values. Let me know if you want the test changed back.

Yeah, agree. But why we do not set this setting in the main repo ? 
изображение

Shane32:
> But why we do not set this setting in the main repo ?

There is no way to set a default setting within the MS `IOptions` framework, unfortunately.  If we were to set the setting, there would be no way for a user to select the other behavior.  And there are most certainly people depending on the old behavior, as we have seen from issues created fairly recently pertaining to how Newtonsoft.Json decodes dates.

src/Transports.Subscriptions.Abstractions/ProtocolMessageListener.cs
```diff
@@ -62,7 +64,7 @@ private Task HandleStartAsync(MessageHandlingContext context)
         {
             var message = context.Message;
             _logger.LogDebug("Handle start: {id}", message.Id);
-            var payload = ((JObject)message.Payload).ToObject<OperationMessagePayload>();
+            var payload = _serializer.ReadNode<GraphQLRequest>(message.Payload);
             if (payload == null)
                 throw new InvalidOperationException("Could not get OperationMessagePayload from message.Payload");
```
sungam3r, line 69:
```suggestion
                throw new InvalidOperationException("Could not get GraphQLRequest from OperationMessage.Payload");
```

sungam3r:
IOptions? Maybe we are talking about different things.

1)
```
  private static JsonSerializerSettings GetDefaultSerializerSettings(bool indent, IErrorInfoProvider errorInfoProvider)
        {
            return new JsonSerializerSettings
            {
                DateParseHandling = DateParseHandling.None,                        <--------------- ADDED
                Formatting = indent ? Formatting.Indented : Formatting.None,
                ContractResolver = new GraphQLContractResolver(errorInfoProvider ?? new ErrorInfoProvider()),
            };
        }
```

Then to override if requred:
```
services.AddNewtonsoftJson(settings => settings.DateParseHandling = DateParseHandling.DateTimeOffset);
```

sungam3r:
изображение
I'm done, 3:20 AM. 👍

Shane32:
> IOptions? Maybe we are talking about different things.

It doesn't work that way.  All of the `IGraphQLBuilder` methods are set up to use the `Configure` method on the interface, which for `MicrosoftDI` configures options via the options pattern.

For instance, calling this:

```cs
services.AddNewtonsoftJson(settings => settings.DateParseHandling = DateParseHandling.DateTimeOffset);
```

Actually runs this (more or less):

```cs
Action<JsonSerializerSettings, IServiceProvider> action = (settings, services) => settings.DateParseHandling = DateParseHandling.DateTimeOffset;
services.AddSingleton<IGraphQLSerializer, NewtonsoftJson.GraphQLSerializer>();
services.AddSingleton<IGraphQLTextSerializer, NewtonsoftJson.GraphQLSerializer>();
services.TryAddSingleton(services => services.GetService<IOptions<JsonSerializerSettings>>()?.Value ?? new JsonSerializerSettings());
services.AddSingleton<IConfigureOptions<JsonSerializerSettings>>(services => new ConfigureNamedOptions<JsonSerializerSettings>(Options.DefaultName, opt => action(opt, services)));
```

Because this configuration pattern is used throughout ALL configuration settings of `IGraphQLBuilder`, ANY settings can be configured via `Configure`, such as this, perhaps:

```cs
services.Configure<JsonSerializerSettings>(Configuration.GetSection("JsonSerializer"));
```

The problem is that with the MS `IOptions` pattern, I could find no way to configure the default settings.  To be very specific, `GetDefaultSerializerSettings` WILL NOT be called.  Below is the constructor that is going to be used inside `NewtonsoftJson.GraphQLSerializer`:

```cs
public GraphQLSerializer(JsonSerializerSettings serializerSettings, IErrorInfoProvider errorInfoProvider)
```

This behavior cannot be changed unless we forfeit the current support of the `IOptions` pattern throughout the API.  I would not make a change.

Shane32:
Even if we were to make some sort of change here, if someone was to register a singleton instance of `JsonSerializerSettings` directly (which is supported with the current design), it would not have those changed defaults.  So for consistency again I would not change the default behavior here.  Yes, it is a breaking change for people that use the server project.  So be it.  It's a better design for the project long-term.

Shane32:
I could add the line of code you mentioned within `GetDefaultSerializerSettings`, but now it would only apply with certain constructors and not others.  Seemed like a bad idea to me.

Shane32:
I had looked at the microsoft source, and it may be possible to change the defaults if we register a custom `IOptionFactory<>` implementation.  This would require duplicating the Microsoft code, or pulling in a specific version of `Microsoft.Extensions.Options`, neither of which I wanted to do.  Manually registered instances of `JsonSerializerSettings` or `IOptions<JsonSerializerSettings>` would still need those 'default' options set.

Shane32:
For reference, here's the `OptionsFactory` source:

https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Options/src/OptionsFactory.cs

Shane32:
Finally, there is one other option.  Make our own settings class for serializer options.  I also did not think this was a good idea, as it would require updating every time that the underlying serializer changes their options class.

I did not think of creating a custom Json options class that inherits the normal options class, and has some code in the constructor to set default options.  We could do that.

Shane32:
> изображение I'm done, 3:20 AM. 👍

!?!?  I marked the only three files that had anything changed in them.

Shane32:
I would prefer if we finished with https://github.com/graphql-dotnet/graphql-dotnet/pull/2892 first

sungam3r:
> It doesn't work that way.

OK. I think I understand.

> a custom Json options class that inherits the normal options class, and has some code in the constructor to set default options. We could do that.

Should solve the problem.

tests/Samples.Server.Tests/ResponseTests.cs
sungam3r: ok

tests/Transports.AspNetCore.Tests/NewtonsoftJsonTests.cs
```diff
@@ -41,64 +42,61 @@ public async Task Decodes_BigInteger()
...
+            ret.Single().Variables["a"].ShouldBeOfType<BigInteger>().ShouldBe(bi);
         }
 
         [Fact]
-        public async Task Dates_Should_Parse_As_Text()
+        public async Task Dates_Should_Parse_As_Dates()
         {
             var ret = await Deserialize(@"{""variables"":{""date"":""2015-12-22T10:10:10+03:00""}}");
-            ret.Single.Inputs["date"].ShouldBeOfType<string>().ShouldBe("2015-12-22T10:10:10+03:00");
+            ret.Single().Variables["date"].ShouldBeOfType<DateTime>().ShouldBe(new DateTimeOffset(2015, 12, 22, 10, 10, 10, TimeSpan.FromHours(3)).LocalDateTime);
```
sungam3r, line 52: The question about default behavior for dates left.

tests/Transports.AspNetCore.Tests/NewtonsoftJsonTests.cs
```diff
@@ -41,64 +42,61 @@ public async Task Decodes_BigInteger()
...
 
         [Fact]
-        public async Task Name_Matching_Not_Case_Sensitive()
+        public async Task Name_Matching_Is_Case_Sensitive()
         {
-            var ret = await Deserialize(@"{""VARIABLES"":{""date"":""2015-12-22T10:10:10+03:00""}}");
-            ret.Single.Inputs["date"].ShouldBeOfType<string>().ShouldBe("2015-12-22T10:10:10+03:00");
+            var exception = await Should.ThrowAsync<Newtonsoft.Json.JsonException>(()
+                => Deserialize(@"{""VARIABLES"":{""date"":""2015-12-22T10:10:10+03:00""}}"));
+            exception.Message.ShouldBe("Exception of type 'Newtonsoft.Json.JsonException' was thrown.");
```
sungam3r, line 67: Let's change behavior of serializer to skip unknown properties.

sungam3r: This error tells about nothing.

tests/Transports.AspNetCore.Tests/SystemTextJsonTests.cs
```diff
@@ -41,64 +41,61 @@ public async Task Decodes_BigInteger()
...
-            ret.Single.Extensions.ShouldBeNull();
+            ret.Single().Extensions.ShouldBeNull();
         }
 
         [Fact]
         public async Task Name_Matching_Is_Case_Sensitive()
         {
-            var ret = await Deserialize(@"{""VARIABLES"":{""date"":""2015-12-22T10:10:10+03:00""}}");
-            ret.Single.Inputs.ShouldBeNull();
+            var exception = await Should.ThrowAsync<System.Text.Json.JsonException>(()
```
sungam3r, line 64: The same here. Imagine 
```json
{
   "query": ....
   "VARIABLES": ....   <----- BOOM
   "variables": ....       <----- even not seen
}
```

tests/Transports.Subscriptions.WebSockets.Tests/WebSocketWriterPipelineTests.cs
sungam3r: "Required" but actually is null + 
изображение
If I remember correctly I already suggested to serialize `type` only if it is non null.

sungam3r:
I looked through all files. I'm glad you got rid of serializers.

tests/Transports.AspNetCore.Tests/NewtonsoftJsonTests.cs
Shane32: That would be a change in the main project.

tests/Transports.AspNetCore.Tests/NewtonsoftJsonTests.cs
Shane32: What's the question?  Any change would need to be made in the main project.

tests/Transports.AspNetCore.Tests/SystemTextJsonTests.cs
Shane32: I'm not exactly sure what your point is.

tests/Transports.AspNetCore.Tests/NewtonsoftJsonTests.cs
sungam3r: I know. Thanks for PR, reviewed.

tests/Transports.AspNetCore.Tests/NewtonsoftJsonTests.cs
sungam3r: Yes, reviewed. Let's merge and update here.

tests/Transports.AspNetCore.Tests/SystemTextJsonTests.cs
sungam3r: Deserializer should just skip unknown properties. My example shows stupid situation when deserializer throws being from one step from the needed property. You changed behavior in main repo, thanks.

Shane32: Well, it was designed to throw an exception no matter the order.  In any case, it is changed now as you suggested.

tests/Transports.Subscriptions.WebSockets.Tests/WebSocketWriterPipelineTests.cs
Shane32: Yes, we already discussed this.

Shane32: If you want to change the behavior to skip `type`, just write a PR to do so.  Ok with me.

sungam3r (approved):

tests/Transports.Subscriptions.WebSockets.Tests/WebSocketWriterPipelineTests.cs
sungam3r: https://github.com/graphql-dotnet/graphql-dotnet/pull/2911

PR #725, opened 2022-02-14
Update examples

sungam3r:


samples/Samples.Server/Startup.cs
```diff
@@ -10,6 +10,7 @@
 using GraphQL.Server.Ui.GraphiQL;
 using GraphQL.Server.Ui.Playground;
 using GraphQL.Server.Ui.Voyager;
+using GraphQL.SystemReactive;
```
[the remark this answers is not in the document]
Shane32: I'm not sure yet.  We should discuss it.  Perhaps review a PR if it's easy to write one (looking at the api diff).  It's probably a good idea, so long as the extension methods relate to the library, like "AddSystemTextJson", which I think is already the case.

Shane32 (approved):

Shane32 (approved):

samples/Samples.Server/StartupWithRouting.cs
```diff
@@ -42,7 +42,7 @@ public void ConfigureServices(IServiceCollection services)
 
             services.AddGraphQL(builder => builder
                 .AddServer(true)
-                .AddDocumentExecuter<SubscriptionDocumentExecuter>()
+                .AddSubscriptionExecutionStrategy()
```
Shane32, line 45: This does need to be updated...

sungam3r: Why?

sungam3r: Compiler shows warning, so I changed.

Shane32: Well, that's correct.  So why didn't we merge this PR?

sungam3r: #727

Shane32: Just to be clear, I'm agreeing with this change, not against it.  Just puzzled why the PR was closed without being merged.

sungam3r: I did not want to resolve conflicts.

Shane32: I already resolved all conflicts

PR #727, opened 2022-02-14
Add serialization packages into metapackage

sungam3r:


Shane32:
I would not force references to both Newtonsoft.Json and System.Text.Json upon users of the 'All' package.  I think All is useful for "all graphql packages" but not serialization packages

Shane32:
One of the big complaints with the current version (it seemed to me) is that the websockets implementation pulled in (and used) the newtonsoft package.  This has been fixed.  We do not want to 'move backwards' by forcing users of All to maintain a reference to newtonsoft.  I think people should pull in the serialization package they need.  Or include GraphQL.SystemTextJson, becuase System.Text.Json is already referenced by ASP.Net

sungam3r:
All is for _all_ things. 

>  I think people should pull in the serialization package they need. 

Then people pull each package they need and don't use _All_.

>  I think All is useful for "all graphql packages" but not serialization packages

Serialization is also graphql-related. I do not see the reasons to consider these two packages with some kind of defective.

sungam3r:
It does. At least for those who wants to drop this package in and get started. For unit tests, examples, etc - I use it in that way.

Shane32:
I would not include the Newtonsoft serialization package in the All package.  If you want to include a serialization package, include SystemTextJson since there already is a dependency from ASP.Net.  If the stated purpose for this metapackage is for tests and samples, then there is no reason to include both anyway.  Someone can always add the newtonsoft package if they really need support for that.  Same philosophy would apply to other DI packages; only MicrosoftDI is included.

Shane32 (approved):

Issue #705, opened 2022-01-05
AddDefaultEndpointSelectorPolicy did not work with the new MicrosoftDI.GraphQLBuilderExtensions


lanwin:
Hi, I am currently upgrading our solution to the latest Graphql Version and found a problem.

Ive upgraded the solution, started it and it works. But I got the AddGraphQl obsolete warnings (by the way, it would be helpful if they had a description what todo to get rid of the obsolete).

So ive changed: 

```csharp
  services.AddGraphQL((options, provider) =>
  {
      options.EnableMetrics = true;
      var logger = provider.GetRequiredService<ILogger<Startup>>();
      options.UnhandledExceptionDelegate = ctx => logger.LogError("{Error} occurred", ctx.OriginalException.Message);
  })
  .AddDefaultEndpointSelectorPolicy()
  .AddSystemTextJson(deserializerSettings => { }, serializerSettings => { })
  .AddWebSockets()
  .AddDataLoader()
  .AddGraphTypes(typeof(MySchema));
```

to 

```csharp
    GraphQL.MicrosoftDI.GraphQLBuilderExtensions.AddGraphQL(services)
    .AddServer(true,
        (options, provider) =>
        {
            options.EnableMetrics = true;
            var logger = provider.GetRequiredService<ILogger<Startup>>();
            options.UnhandledExceptionDelegate = ctx => logger.LogError("{Error} occurred", ctx.OriginalException.Message);
        })
    .AddDefaultEndpointSelectorPolicy()
    .AddSystemTextJson(deserializerSettings => { }, serializerSettings => { })
    .AddWebSockets()
    .AddGraphTypes(typeof(MySchema).Assembly);
```

And now Ive get:
```
Microsoft.AspNetCore.Routing.Matching.AmbiguousMatchException: The request matched multiple endpoints. Matches:
GraphQL WebSockets
GraphQL
```

Using NET5 and Routing.

Ive played around with the various extensions, but did not find a working solution yet.

sungam3r:
Did you add `AddRouting`?

lanwin:
> `AddRouting`

Yes three lines above AddGraphQL.             

```
services.AddRouting();
services.AddResponseCompression();
```

sungam3r:
To investigate further I need to see `ConfigureServices` and `Configure` methods.

sungam3r:
Or you may copy source code, rename class and just debug it:
```c#
[...]

PR #728, opened 2022-02-14
Fix AddDefaultEndpointSelectorPolicy

Shane32:
Fixes:
- #705

### DIFF
--- a/src/Transports.AspNetCore/Extensions/GraphQLBuilderUserContextExtensions.cs
+++ b/src/Transports.AspNetCore/Extensions/GraphQLBuilderUserContextExtensions.cs
@@ -149,7 +149,7 @@ namespace GraphQL.Server
         /// <returns>The GraphQL builder.</returns>
         public static DI.IGraphQLBuilder AddDefaultEndpointSelectorPolicy(this DI.IGraphQLBuilder builder)
         {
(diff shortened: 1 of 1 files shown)


Shane32:
There is no logical reason to need `TryAddEnumerable` support.  Of course someone could call `AddDefaultEndpointSelectorPolicy` twice, which would cause a problem, but then they could also call `AddDocumentListener<T>` or `AddMiddleware<T>` twice also, which would also be a problem.  Perhaps there will be a time when we need support for `TryAddEnumerable`, but I do not think we need to do so now.

Shane32:
Note that Autofac very likely does not support `TryAddEnumerable`.  I know it does not support `Register` with `replace = true`, which is similar in concept, and we will simply have to ignore the `replace` parameter there (which should be fine).  Ideally we would remove the `replace` parameter and write the code such that it is not needed, for more compatibility with any DI engine.

sungam3r:
I suggest to merge #2946 and keep `TryRegister` to be more explicit.

sungam3r:
> Note that Autofac very likely does not support TryAddEnumerable.

When I worked on https://github.com/EasyNetQ/EasyNetQ/pull/1108 I was very necessary for the advice regarding Autofac. After a couple of days attempts to force all providers to work the same, I came to the conclusion that it is impossible.

sungam3r:
#730

PR #735, opened 2022-02-17
Introduce GraphQLHttpMiddleware.HandleRequest and switch to IMiddleware

sungam3r:
Allows to override response handling/writing, for example if one wants to proxy request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,8 +129,8 @@ public void ConfigureServices(IServiceCollection services)
     // Add GraphQL services and configure options
     services.AddGraphQL(builder => builder
         .AddServer(true)
-        .AddMiddleware<ChatSchema>()
-        .AddWebSocketsMiddleware<ChatSchema>()
+        .AddHttpMiddleware<ChatSchema>()
+        .AddWebSocketsHttpMiddleware<ChatSchema>()
         // For subscriptions support
         .AddDocumentExecuter<SubscriptionDocumentExecuter>()
         .AddSchema<ChatSchema>()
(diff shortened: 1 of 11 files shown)


sungam3r:
@Shane32 I would like to merge this ASAP. At work we have special GraphQL proxy called "Security Proxy" that handles all GraphQL requests to real services. In order to do so we use source code from server project with some modified bits. This PR allows us to switch to "standard" middleware and remove all copied code keeping only `HandleRequest` bits.

Shane32:
Makes sense but can we have the protected method just run the code inside the loop, which deduplicates code?

Shane32:
I guess I’m just saying, sure, but we should have a few more protected methods here, to allow overriding various bits, not just one. InvokeAsync is still too long and not virtual.  I should probably look at my code and see what functionality I would need as a virtual method in order to derive from and use the stock middleware. I’m fine if you want to merge this today though.

sungam3r:
I added `IMiddleware` as you did it GraphQL.NET example. It requires DI registrations now but middleware can be singleton.

sungam3r:
Regarding loop - both branches are rather small and differ a bit. I would not try to merge them into one virtual method, as least for now.

Shane32:
I see you made some new extension methods. Don’t we already have AddMiddleware methods? I would call these new ones AddHttpMiddleware and AddWebsocketsHttpMiddleware to eliminate confusion. Further, I think they are actually optional. We may want to see if, without these methods, the middleware gets instantiated every call or acts as a singleton. Or if they need to be registered.  Note that behavior may be different with the change to use IMiddleware.

sungam3r:
> We may want to see if, without these methods, the middleware gets instantiated every call or acts as a singleton. 

After switching for `IMiddleware` registration in DI is `required**.

Shane32:
I think we should rename the methods though. AddGraphQL.AddMiddleware implies GraphQL middleware not http middleware.

sungam3r:
renamed

sungam3r:
OK to merge?

Shane32 (approved):

PR #741, opened 2022-03-01
Support non utf-8 charsets

sungam3r:


src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -80,7 +80,21 @@ public virtual async Task InvokeAsync(HttpContext context, RequestDelegate next)
                         GraphQLRequest[] deserializationResult;
                         try
                         {
+#if NET5_0_OR_GREATER
+                            var sourceEncoding = GetEncoding(httpRequest);
```
Shane32, line 84:
```suggestion
                            var sourceEncoding = GetEncoding(mediaTypeHeader.CharSet);
```

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -288,5 +302,39 @@ private async Task<GraphQLRequest> DeserializeFromGraphBodyAsync(Stream bodyStre
...
+        {
+            if (!request.Headers.TryGetValue("Content-Type", out var value))
+                return null;
+
+            string charset = value.ToString();
+            var position = charset.IndexOf("charset=");
+            if (position == -1)
+                return null;
+
+            charset = charset.Substring(position + 8);
```
Shane32, line 317:
```suggestion
        private static System.Text.Encoding GetEncoding(string? charset)
        {
```

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -288,5 +302,39 @@ private async Task<GraphQLRequest> DeserializeFromGraphBodyAsync(Stream bodyStre
...
+            {
+                // Remove at most a single set of quotes.
+                if (charset.Length > 2 && charset[0] == '\"' && charset[charset.Length - 1] == '\"')
+                {
+                    return System.Text.Encoding.GetEncoding(charset.Substring(1, charset.Length - 2));
+                }
+                else
+                {
+                    return System.Text.Encoding.GetEncoding(charset);
+                }
```
Shane32, line 331: Verify to see if MediaTypeHeaderValue.TryParse removes these quotes automatically...

```suggestion
                return System.Text.Encoding.GetEncoding(charset);
```

Shane32 (changes requested):
@sungam3r Have you run into an issue that prompted this change?

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -288,5 +302,39 @@ private async Task<GraphQLRequest> DeserializeFromGraphBodyAsync(Stream bodyStre
...
+            }
+            catch (ArgumentException e)
+            {
+                throw new InvalidOperationException("The character set provided in ContentType is invalid.", e);
```
Shane32, line 335: This should be something like this:
```cs
await HandleContentTypeCouldNotBeParsedErrorAsync(context);
```

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
sungam3r: 🤦‍♂️

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -80,7 +80,21 @@ public virtual async Task InvokeAsync(HttpContext context, RequestDelegate next)
                         GraphQLRequest[] deserializationResult;
                         try
                         {
+#if NET5_0_OR_GREATER
+                            var sourceEncoding = GetEncoding(httpRequest);
+                            // Wrap content stream into a transcoding stream that buffers the data transcoded from the sourceEncoding to utf-8.
+                            if (sourceEncoding != null && sourceEncoding != System.Text.Encoding.UTF8)
+                            {
+                                using var tempStream = System.Text.Encoding.CreateTranscodingStream(httpRequest.Body, innerStreamEncoding: sourceEncoding, outerStreamEncoding: System.Text.Encoding.UTF8);
```
Shane32, line 88:
```suggestion
                                using var tempStream = System.Text.Encoding.CreateTranscodingStream(httpRequest.Body, innerStreamEncoding: sourceEncoding, outerStreamEncoding: System.Text.Encoding.UTF8, leaveOpen: true);
```

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -80,7 +80,21 @@ public virtual async Task InvokeAsync(HttpContext context, RequestDelegate next)
...
+                                deserializationResult = await _serializer.ReadAsync<GraphQLRequest[]>(tempStream, cancellationToken);
+                            }
+                            else
+                            {
+                                deserializationResult = await _serializer.ReadAsync<GraphQLRequest[]>(httpRequest.Body, cancellationToken);
+                            }
+#else
                             deserializationResult = await _serializer.ReadAsync<GraphQLRequest[]>(httpRequest.Body, cancellationToken);
```
Shane32, line 96: It would be nice to deduplicate this line of code which now appears 3 times... but it's okay if that's not practical.

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
sungam3r: It does not.

Shane32:
Personally, I'd just add something like this and add none of the transcoding:

```cs
if (mediaTypeHeader.CharSet != null && mediaTypeHeader.CharSet != "utf-8")
    await HandleContentTypeCouldNotBeParsedErrorAsync(context);
```

And I wouldn't bother with transcoding and stuff.  But if there's a need, sure I don't mind adding transcoding.

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
sungam3r: fixed

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
sungam3r: Counteroffer  - drop 3.1 tfm for v7 .

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
Shane32: Then are you sure it is in the spec?  I would like to see a spec that says it may have quotes before we add code parsing quotes which typically are not there and are not parsed by the .NET code.

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
Shane32: No, supporting 3.1 is essential.  I have projects that use 3.1 and cannot be upgraded at this time.  I'm sure others do as well.

sungam3r:
> Have you run into an issue that prompted this change?

Not exactly. My experience shows that our infrastructure is full of cases when old systems call new, passing data in unusual encodings. This is not very often, but it happens.

sungam3r:
I sometimes have to face with windows-1251 encoding.

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
sungam3r: 3.1 server or 3.1 GraphQL.NET ?

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
Shane32: https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.1.1

https://github.com/dotnet/runtime/issues/15947

https://github.com/dotnet/corefx/pull/33280

Unfortunately, the quote parsing has to stay (for 100% compatibility), and .NET refused to fix their parser because "it might break compatibility".

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
Shane32: .NET Core 3.1

Shane32 (approved):

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
sungam3r: Actually I copied this code from MS sources and then changed a bit.

PR #749, opened 2022-03-16
[WIP] New subscription support

Shane32:


Shane32:
@sungam3r Here is a working sample of subscription support with no dependencies on Dataflow or System.Reactive.  As such, all the websocket code can be rolled into the `Transports.AspNetCore` nuget package. I have not written xml comments or comprehensive tests.  I did add support for the newer `graphql-transport-ws` protocol (as well as the older `graphql-ws` protocol).

Here are some functional changes:
- The newer `graphql-transport-ws` protocol is supported (as well as the older `graphql-ws` protocol).  Previously only the old protocol was supported.
- The `UserContext` is built via `IUserContextBuilder` once when the HTTP connection is established.  Then it is re-used for the lifetime of the WebSockets connection.  Previously the `IUserContextBuilder` was ignored for WebSockets connections.
- The `IGraphQLExecuter` (indirectly) and `IUserContextBuilder` (directly) is injected to the middleware in the constructor.  This could easily be changed to pull from `RequestServices` like the query/mutation Asp.Net middleware does.
- Much of the logging code has been removed; only what is in the Asp.Net middleware remains.
- When a WebSockets request with an invalid protocol is received, a 400 error is returned.  Previously the connection would be accepted and then a WebSocket close message would be sent stating there was a protocol error.

I will be using this code in my own projects.  If you think I should proceed I can add comments and tests and integrate it here - but only if you think so.  These changes required a lot of effort, and there is a lot of cleanup to be done to make it ready for release in April.

Alternatively, we could rebase #694 onto `develop`, but I'm not very interested in working on it.  This project badly needs support for the current protocol one way or another.

sungam3r:
😮  I'll review this after https://github.com/graphql-dotnet/graphql-dotnet/pull/3004

Shane32:
Perhaps we should discuss some bigger-picture changes regarding the server project before a complete review of these changes.

Shane32:
The suggested changes in this PR have had much more recent updates in the https://github.com/Shane32/GraphQL.AspNetCore3 project, including a full suite of tests.

Shane32:
This PR is outdated.  Updated code is at GraphQL.AspNetCore3 repo.

PR #750, opened 2022-03-16
Remove Authorization dependency on Core

Shane32:


Shane32:
> It became possible after GraphQLRequest/OperationMessage rework, right?

Honestly I don't know.  I'm not sure that the authorization library needed those classes either.  I would have to guess that it is possible because `IGraphQLBuilder` was moved to GraphQL.NET.

sungam3r:
Right, it's about builder.

PR #752, opened 2022-03-17
Restrict acceptable GraphQL.NET versions to v4

Shane32:
This change is intended for GraphQL Server v5 only.

Tested built package with NuGet Package Explorer and it works properly.

Example of "All" nuget package:
image


sungam3r (approved):
Do you want to release v5.2.1 with new constraints just before v6?

Shane32:
Anytime will do. Now, later, whenever. It will just help ensure no mismatch of versions.

PR #753, opened 2022-03-18
Fix batched results with a single request

Shane32:
Test failed before; passes now.  Note that this actually reduces an allocation for batch requests, since previously the List would execute ToArray within the GraphQL serializer, and now it returns the List as an IList.  Since foreach is not used, there is no negative to the IList over the array.


tests/Samples.Server.Tests/ResponseTests.cs
```diff
@@ -64,6 +64,15 @@ public async Task Batched_Query_Should_Return_Multiple_Results()
             response.ShouldBeEquivalentJson(@"[{""data"":{""__schema"":{""queryType"":{""name"":""ChatQuery""}}}},{""data"":{""__schema"":{""queryType"":{""name"":""ChatQuery""}}}},{""data"":{""__schema"":{""queryType"":{""name"":""ChatQuery""}}}}]", ignoreExtensions: true);
         }
 
+        [Fact]
+        public async Task Batched_Query_Should_Return_Single_Result()
```
sungam3r, line 68:
```suggestion
        public async Task Batched_Query_Should_Return_Single_Result_As_Array()
```

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -106,7 +106,7 @@ public virtual async Task InvokeAsync(HttpContext context, RequestDelegate next)
                                 throw;
                             return;
                         }
-                        if (deserializationResult?.Length == 1)
+                        if (deserializationResult is GraphQLRequest[] array && array.Length == 1)
```
sungam3r, line 109: As I inderstand (looking through code of serializer) `&& array.Length == 1` condition here is unnecessary now but I'm fine to keep it to make things more explicit.

sungam3r, line 109:
```suggestion
                        // https://github.com/graphql-dotnet/server/issues/751
                        if (deserializationResult is GraphQLRequest[] array && array.Length == 1)
```

sungam3r (approved):

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
Shane32: Correct.

PR #754, opened 2022-03-18
Wrap subscription exceptions passed along via OnError

Shane32:
- Required by https://github.com/graphql-dotnet/graphql-dotnet/pull/3004 to retain existing behavior in respect to source-raised errors
- Even if the above PR is not merged, this change does not hurt existing implementations
- As a behavioral change, I suggest that OnError also call OnComplete -- see comments in code for details

### DIFF
--- a/src/Transports.Subscriptions.Abstractions/Subscription.cs
+++ b/src/Transports.Subscriptions.Abstractions/Subscription.cs
@@ -4,6 +4,7 @@ using System;
 using System.Linq;
 using System.Reactive.Linq;
 using System.Threading.Tasks;
+using GraphQL.Execution;
 using GraphQL.Subscription;
 using GraphQL.Transport;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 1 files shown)


src/Transports.Subscriptions.Abstractions/Subscription.cs
```diff
@@ -56,7 +57,31 @@ public void OnCompleted()
...
+            {
+                // but in the unlikely event that an unhandled exception delegate throws an exception,
+                // or for any other reason it is not an ExecutionError instance, wrap the error
+                executionError = new UnhandledError($"Unhandled error of type {error?.GetType().Name}", error!);
+            }
+
+            // pass along the error as an execution result instance
+            OnNext(new ExecutionResult { Errors = new ExecutionErrors { executionError } });
+
+            // Optionally we can disconnect the client by calling OnComplete() here
```
sungam3r, line 79: Is it a loss of backward compatibility?

sungam3r (approved):

src/Transports.Subscriptions.Abstractions/Subscription.cs
```diff
@@ -56,7 +57,31 @@ public void OnCompleted()
             _unsubscribe = null;
         }
 
-        public void OnError(Exception error) => throw new NotImplementedException();
```
sungam3r, line 59: How did it work before?

src/Transports.Subscriptions.Abstractions/Subscription.cs
Shane32: Previously it did not terminate the connection after an error. That is why the code here is commented out. I believe, after reviewing MS recommendations on IObservable, that it should. It will can result in different behavior depending on user code.

If OnError is not called by the observable then there is no difference. If it is, and if they are using System.Reactive which will discard packets after OnError, then this change actually improves support because it informs the client that no more packets will be received. 

And keep in mind that this library used System.Reactive previously so it would not have been possible to send additional packets after OnError regardless of what the user code did.

src/Transports.Subscriptions.Abstractions/Subscription.cs
Shane32: The subscription execution strategy converted OnError events to OnData events with an ExecutionResult wrapping the error. The PR over there still wraps the error in an ExecutionError but passes it along OnError for it to get wrapped here by this PR.

src/Transports.Subscriptions.Abstractions/Subscription.cs
sungam3r: > after reviewing MS recommendations on IObservable

OK, but keep in mind that all that MS stuff around observables is an implementation details. They can give any recommendations, but these recommendations have a lower priority compared with the GraphQL design/protocols. The behavior of a specific implementation should not dictate the behavior of the top-level design.

Shane32: See:
- #755

PR #755, opened 2022-03-20
Source errors terminate subscription within WebSockets stream

Shane32:
This change causes errors raised by the source stream to inform the client that there will be no more events and disconnects the subscription (although it does not disconnect the WebSockets connection).  This does not affect resolver execution errors; they will be passed with the data responses uninterrupted.

This change request is based on the following "Best Practices Guide":

https://docs.microsoft.com/en-us/dotnet/standard/events/observer-design-pattern-best-practices

> Once the provider calls the OnError or IObserver<T>.OnCompleted method, there should be
> no further notifications, and the provider can unsubscribe its observers.

It is important to note also that in accordance with the 'best practices' documentation above, the System.Reactive package's `Subject<T>` class will stop sending packets after an `OnError` event is raised.  As such, if the user was using the `Subject<T>` class to broadcast notifications, it would be appropriate to inform connected clients that the stream has ended.

Also, if the user code never calls `OnError`, then there will be no behavioral change, as this code would never come into play.  (Any other errors, such as document listener exceptions, are distributed through `OnNext` as resolver errors.)

After reviewing MS recommendations on IObservable, I believe this change is correct.


sungam3r:
I'm not sure if this change is dangerous for clients but let's merge and go on.

PR #759, opened 2022-03-28
Fix some exception handling bugs

Shane32:
Some bugs I found while writing tests


src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -129,6 +129,7 @@ public virtual async Task InvokeAsync(HttpContext context, RequestDelegate next)
                             {
                                 if (!await HandleDeserializationErrorAsync(context, ex))
                                     throw;
+                                return;
```
sungam3r, line 132: oops, indeed

sungam3r (approved):

Issue #132, opened 2018-07-11
Add Automatic Persisted Query Support


RehanSaeed:
[Automatic Persisted Queries](https://blog.apollographql.com/automatic-persisted-queries-and-cdn-caching-with-apollo-server-2-0-bf42b3a313de) is something implemented by the Appollo GraphQL Server. The server caches queries with their hash and the client can send a hash of the query instead of the full multi-kilobyte payload. To illustrate:

![1 mmsphtd3-9p7dqcb5b1d7g](https://user-images.githubusercontent.com/2270364/42558379-f0e44dbc-84e8-11e8-99ef-6d051102f7da.png)

If no hash of the query is found on the server:

![1 1qc1q6atkn1k7x8zduc9ja](https://user-images.githubusercontent.com/2270364/42558497-5538e516-84e9-11e8-84f2-07ed16aca13c.png)

I think this can be implemented by using the built in `IDistributedCache` in ASP.NET Core.

RehanSaeed:
Forgot to mention that there is a `cacheControl` directive that can be used alongside this. For example, this schema indicates that all fields that return an Author should be cached for 60 seconds, and that the posts field should itself be cached for 180 seconds:

```
type Author @cacheControl(maxAge: 60) {
  id: Int
  firstName: String
  lastName: String
  posts: [Post] @cacheControl(maxAge: 180)
}
```

More info in the [docs](https://www.apollographql.com/docs/guides/performance.html#automatic-persisted-queries).

aleGuardiola:
I am working in a project that already support this two features maybe you can check it out. I did it based on the apollo server. https://github.com/aleGuardiola/Graphql-Controller

miladkakouli:
you can use this package for graphql persisted queries 

https://www.nuget.org/packages/GraphQL.PersistedQueries/
https://github.com/miladkakouli/graphql-dotnet-persistedqueries

sungam3r:
Regarding `cacheControl` directives - it is another task to do compared to APQ iitself.

sungam3r:
Done in https://github.com/graphql-dotnet/graphql-dotnet/pull/3134.

PR #782, opened 2022-05-05
Automatic Persisted Queries

SlavaUtesinov:
Implementation of [Automatic Persisted Queries](https://www.apollographql.com/docs/react/api/link/persisted-queries) (APQ) based on an APQ protocol and in memory cache located in a server's middleware layer as a possible solution of the issue #132. The feature was implemented in accordance with [Apollo server approach](https://github.com/apollographql/apollo-server/blob/main/packages/apollo-server-core/src/requestPipeline.ts#L170,L208), corresponding [entry point](https://www.apollographql.com/docs/apollo-server/performance/apq/#memcached):
```
const server = new ApolloServer({
  typeDefs,
  resolvers,
  csrfPrevention: true,
  persistedQueries: {
    cache: new MemcachedCache(),
  },
});
```
and [general architecture](https://www.apollographql.com/blog/apollo-client/persisted-graphql-queries/):
image


Shane32:
This is a great idea, but I think this should exist as a separate nuget package in the main project.  It need not have any tight integration with the server project.  We spent a notable amount of time for GraphQL Server 6 to eliminate interdependency on other parts of GraphQL.

In this case, all that is necessary is a custom document executer which can wrap another document executer.  The only change in necessary in the server project is to allow empty queries to be 'executed'.  Such a change has been implemented and released in GraphQL.AspNetCore3 version 3.0.0 (see https://github.com/Shane32/GraphQL.AspNetCore3/pull/28).

Ideally we should modify `IConfigureExecution` to allow modification of the request and response, allowing for APQ and Apollo Tracing extension methods, without having to stack document executer implementations (which works poorly with MSDI since decorators are not supported).  But this would need to take place in GraphQL v6 as it would be a breaking change.

On a side note, as it stands now this PR currently does not utilize the APQ for requests over websocket connections -- but I would not spend any time on any implementations within this library until all code from GraphQL.AspNetCore3 has been migrated over.  Besides, as a `IDocumentExecuter` implementation, all requests would flow through the APQ handler regardless of server implementation or the specific transport in use.

@sungam3r comments?

Shane32:
As for implementation, we should consider supplying an implementation based on `IDistributedCache` so that a Redis server can be used to pull cache entries.  A `MemoryCache`-based implementation is of course also useful and would be slightly more optimized than a `MemoryDistributedCache` implementation.

SlavaUtesinov:
Actually, I was partially inspired by this [article](https://www.apollographql.com/blog/apollo-client/persisted-graphql-queries/). There it was written that persisted queries should be implemented even in front of a server but in terms of asp.net core, I think, it would be fair enough to interpret/translate it as a well-known servers' middleware class. 
Some extract from "Server" paragraph:
_"This means that after you drop in some really simple middleware, your server code doesn’t have to change at all since it can just pretend that it received the GraphQL document from the client, per usual."_

image

Shane32:
> This means that after you drop in some really simple middleware, your server code doesn’t have to change at all since it can just pretend that it received the GraphQL document from the client, per usual

I think I'm agreeing, although perhaps interpreting it differently.  I am trying to say that I think this feature should be "middleware", such that it does not need to interact directly with this particular server project, or any particular graphql schema.  Rather anyone with any GraphQL configuration can just write:

```cs
services.AddGraphQL(b => b
    // configuration of schema
    .AddAutomaticPersistedQueries()    // <-- "middleware" is injected
    // whatever else
);
```

or similar.  Many people still use MVC controller actions to initiate the GraphQL pipeline.  Others, like myself, have written custom ASP.NET middleware based on various unique requirements.

We have many injection points already in the system that behave like middleware.  There are document listeners, "field middleware", the `IConfigureSchema` interface, and the `IConfigureExecutionOptions` interface, not to mention the various interfaces for resolving graph types and similar.  Unfortunately none of them is _exactly_ what we need, and we struggled with the same issue for Apollo Tracing results.  For now we can replace the document listener (as we did with Apollo Tracing results), but that doesn't allow us to easily build an execution and stack behaviors.

I think we really need to enhance the `IConfigureExecutionOptions` interface to act as middleware, allowing the ability to replace or stack behaviors, which will be useful in this scenario and also for Apollo Tracing results.  I would use that feature in my own libraries.  Then we can inject this behavior into the pipeline "automatically", without it being a part of the server project proper.

Until then, I suggest we write a document listener to utilize the `IAutomaticPersistedQueryCache` interface, but not build anything directly into the server project.  Then when `AddAutomaticPersistedQueries()` or similar is called, it will just configure a custom document executer, which can be used with this server project, or any other.  The document executer should have a constructor that allows wrapping behavior around another document executer in case the need arises for two behaviors to stack.  That way it is relatively easy to (by hand) stack multiple behaviors together, while retaining separation of this new functionality from the server project or the main project.

sungam3r:
>  A MemoryCache-based implementation is of course also useful and would be **slightly** more optimized than a MemoryDistributedCache implementation.

Not slightly. `MemoryDistributedCache` (as well as `IDistributedCache`) works with `byte[]` values. NET provides `DistributedCacheExtensions` utility class with a bunch methods converting `byte[]` into `string:
```cs
  public static void SetString(this IDistributedCache cache, string key, string value, DistributedCacheEntryOptions options)
        {
            if (key == null)
            {
                throw new ArgumentNullException(nameof(key));
            }
            if (value == null)
            {
                throw new ArgumentNullException(nameof(value));
            }
            cache.Set(key, Encoding.UTF8.GetBytes(value), options);
        }
  public static string GetString(this IDistributedCache cache, string key)
        {
            byte[] data = cache.Get(key);
            if (data == null)
            {
                return null;
            }
            return Encoding.UTF8.GetString(data, 0, data.Length);
        }
```

So allocations... Constant allocations will only harm. Caching in memory is done intentionally.

sungam3r:
@Shane32 @SlavaUtesinov I will look more carefully tomorrow. I agree with @Shane32 words that this can be done "more beautifully". Although the current implementation has its own value. It does not require breaking changes in design, so it does not require a major version bump and can work now.

> This is a great idea, but I think this should exist as a separate nuget package in the main project.

😕 Perhaps, but I doubt it is worth it.

Shane32:
We already have a caching nuget package. It can exist in there.

SlavaUtesinov:
I think, a main point of the article is that we should distance APQ layer from a other staff as much as possible to make them totally independent i.e. the last should know nothing about the former (it is just a way to supply a request i.e. like transport details), because exchange from hash to query or any similar things could be implemented even at a separate server located in front of a main graphql server (that is why at the article they didn't unite terms: server and middleware(it is not necessary a part of the server)). Also in case of Apollo they located ["my code"](https://github.com/apollographql/apollo-server/blob/444c403011209023b5d3b5162b8fb81991046b23/packages/apollo-server-core/src/requestPipeline.ts#L172,L208) at a server's project pipeline separately from [graphql execution part](https://github.com/apollographql/apollo-server/blob/444c403011209023b5d3b5162b8fb81991046b23/packages/apollo-server-core/src/requestPipeline.ts#L403).

sungam3r:
It makes sense.

Shane32:
See https://github.com/graphql-dotnet/graphql-dotnet/pull/3121

Shane32:
> I think, the main point of the article is that we should distance APQ layer from other staff as much as possible to make them totally independent

This statement is just as true for being independent from the server project as the execution.  Take a look at Apollo Link here: https://www.apollographql.com/docs/react/api/link/introduction/ - you can see that any behavior such as APQ support can be added as a "link" so that, as it says in the blog, it "works with any client or server"

With https://github.com/graphql-dotnet/graphql-dotnet/pull/3121 we can add the equivalent of "Apollo Links" which can be entirely independent from either the server implementation or the execution.

Keep in mind that this project is not intended to be a port of Apollo's code.  It is an independent project designed to be as well-designed as we can make it, while meeting the GraphQL specification and related specifications.  Just because Apollo implemented Automatic Persisted Queries within directly within their main codebase does not mean we intend to do so.  APQ would be a great feature to add.  But it should not be directly in the server project, nor directly in the main GraphQL.NET project.  We already have a GraphQL.MemoryCache nuget package, and this feature could easily be added there.

Shane32:
FYI, although it may not be entirely clear from the article, Apollo implemented APQ as a Link.  See https://www.apollographql.com/docs/react/api/link/persisted-queries

SlavaUtesinov:
I am sure, you've made a great work in your PR which will lead to new benefits. I think, these PR-s don't contradict each other. It is worth mentioning that in both articles they are talking about "links" with respect to Apollo Client's data flow, not server nor actual graphql execution stage. I want to say, they didn't adhere the same approach on the other/backend side for some reason.

Shane32:
> these PR-s don't contradict each other

That's not the point.  The point is to allow this feature to work with any server project or custom implementation.  As stated before, many people have custom server implementations.  I certainly do in some of my own projects.  In addition, there are multiple server projects that exist _and are maintained_:
- https://github.com/graphql-dotnet/server
- https://github.com/JuergenGutsch/graphql-aspnetcore/tree/develop/GraphQl.AspNetCore
- https://github.com/Shane32/GraphQL.AspNetCore3

Why add this middleware here?  Why not add it as a feature that can be applied to any implementation?  The entire design of GraphQL.NET allows for such customizations.

> they are talking about "links" with respect to Apollo Client's data flow

True but the same principle applies here.  Take a look at ASP.NET Core's design.  Anyone can add middleware to the pipeline to make changes to the HttpContext and/or replace behavior.  Exactly the same as Apollo Links do in a client context.  GraphQL.NET should learn from those designs and allow for a customizable server pipeline.  Which, in fact, we already have 90% of the functionality necessary; my PR is a relatively minor change to allow for the necessary functionality to support APQ.

@sungam3r and I spent much time over the last 6-12 months to take all the dependencies and common code out of the server project so to isolate the server project to simply providing the HTTP-request transition.  This PR is a great feature, but there is no reason to integrate it directly into this project.  It does not have anything to do with ASP.NET Core, nor does it have anything to do with the transport in use.  Let's add a PR to the GraphQL.NET project for this feature.

sungam3r:
https://github.com/graphql-dotnet/graphql-dotnet/pull/3121 reviewed.

tests/Samples.Server.Tests/Serializer.cs
```diff
@@ -39,6 +39,11 @@ internal static FormUrlEncodedContent ToFormUrlEncodedContent(GraphQLRequest req
             dictionary["variables"] = ToJson(request.Variables);
         }
 
+        if (request.Extensions != null)
```
sungam3r, line 42: Extracted into #785

Shane32:
@SlavaUtesinov PR https://github.com/graphql-dotnet/graphql-dotnet/pull/3121 has been merged and all necessary changes have been made in the main project to support APQ as an independent feature/middleware.  If you would like to add this feature to the `GraphQL.MemoryCache` nuget package, please make a PR against the main GraphQL.NET repo, and we can probably get this feature into the 5.3.0 release.  Thanks!

SlavaUtesinov:
Ok, thanks.

sungam3r:
👍

PR #799, opened 2022-05-20
Add WebSocket support to IUserContextBuilder

Shane32:
This changes `IUserContextBuilder` to support WebSocket connections via the graphql-ws or graphql-transport-ws subprotocol. 

This is provided by:
- Adding `object? payload` to the interface method signature
- Changing the websocket support (not yet posted as a PR) to create the user context while the initialization message is being processed, after authentication has occurred.  (Now it runs before WebSocket authentication.)

So with this change:
1. You can access `HttpContext.User` to pull properties of the user
2. You can examine any other properties of the WebSocket connection initialization message while creating the user context

Breaking changes:
- Direct implementations of `IUserContextBuilder` need to be modified

Non-breaking:
- The user context builder methods still work without changes.  New methods were added if the code needs the payload.

### DIFF
--- a/src/Transports.AspNetCore/Extensions/GraphQLBuilderUserContextExtensions.cs
+++ b/src/Transports.AspNetCore/Extensions/GraphQLBuilderUserContextExtensions.cs
@@ -92,7 +92,8 @@ public static class GraphQLBuilderUserContextExtensions
         where TUserContext : class, IDictionary<string, object?>
     {
         builder.Services.Register<IUserContextBuilder>(new UserContextBuilder<TUserContext>(creator ?? throw new ArgumentNullException(nameof(creator))));
-        builder.ConfigureExecutionOptions(options => {
+        builder.ConfigureExecutionOptions(options =>
+        {
             if (options.UserContext == null || options.UserContext.Count == 0 && options.UserContext.GetType() == typeof(Dictionary<string, object>))
             {
                 var requestServices = options.RequestServices ?? throw new MissingRequestServicesException();
@@ -111,7 +112,8 @@ public static class GraphQLBuilderUserContextExtensions
         if (creator == null)
             throw new ArgumentNullException(nameof(creator));
         builder.Services.Register<IUserContextBuilder>(new UserContextBuilder<TUserContext>((context, payload) => new(creator(context, payload))));
-        builder.ConfigureExecutionOptions(async options => {
+        builder.ConfigureExecutionOptions(async options =>
+        {
             if (options.UserContext == null || options.UserContext.Count == 0 && options.UserContext.GetType() == typeof(Dictionary<string, object>))
             {
                 var requestServices = options.RequestServices ?? throw new MissingRequestServicesException();
--- a/src/Transports.AspNetCore/Extensions/GraphQLBuilderUserContextExtensions.cs
+++ b/src/Transports.AspNetCore/Extensions/GraphQLBuilderUserContextExtensions.cs
@@ -27,7 +27,7 @@ public static class GraphQLBuilderUserContextExtensions
                 var requestServices = options.RequestServices ?? throw new MissingRequestServicesException();
                 var httpContext = requestServices.GetRequiredService<IHttpContextAccessor>().HttpContext!;
                 var contextBuilder = requestServices.GetRequiredService<IUserContextBuilder>();
-                options.UserContext = await contextBuilder.BuildUserContextAsync(httpContext);
+                options.UserContext = await contextBuilder.BuildUserContextAsync(httpContext, null);
             }
         });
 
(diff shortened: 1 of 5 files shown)


Shane32:
This needs to be merged to allow my WebSockets PR to merge in.

src/Transports.AspNetCore/UserContextBuilder.cs
```diff
@@ -8,28 +8,54 @@ namespace GraphQL.Server.Transports.AspNetCore;
...
     /// <summary>
     /// Initializes a new instance with the specified delegate.
     /// </summary>
     public UserContextBuilder(Func<HttpContext, ValueTask<TUserContext>> func)
     {
-        _func = func ?? throw new ArgumentNullException(nameof(func));
+        if (func == null)
+            throw new ArgumentNullException(nameof(func));
+
+        _func = async (context, payload) => await func(context);
```
sungam3r, line 21:
```suggestion
        _func = async (context, _) => await func(context);
```

src/Transports.AspNetCore/UserContextBuilder.cs
```diff
@@ -8,28 +8,54 @@ namespace GraphQL.Server.Transports.AspNetCore;
...
-    /// Initializes a new instance with the specified delegate.
-    /// </summary>
+    /// <inheritdoc cref="UserContextBuilder(Func{HttpContext, ValueTask{TUserContext}})"/>
     public UserContextBuilder(Func<HttpContext, TUserContext> func)
     {
         if (func == null)
             throw new ArgumentNullException(nameof(func));
 
-        _func = x => new(func(x));
+        _func = (context, payload) => new(func(context));
```
sungam3r, line 30:
```suggestion
        _func = (context, _) => new(func(context));
```

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -483,12 +485,12 @@ protected virtual ValueTask<bool> HandleAuthorizeWebSocketConnectionAsync(HttpCo
     /// In this manner, both scoped and singleton <see cref="IUserContextBuilder"/>
     /// instances are supported, although singleton instances are recommended.
     /// </summary>
-    protected virtual async ValueTask<IDictionary<string, object?>> BuildUserContextAsync(HttpContext context)
+    protected virtual async ValueTask<IDictionary<string, object?>> BuildUserContextAsync(HttpContext context, object? payload)
```
sungam3r, line 488: 🤔 Is `payload` always `null` here?

src/Transports.AspNetCore/Extensions/GraphQLBuilderUserContextExtensions.cs
```diff
@@ -80,6 +80,49 @@ public static IGraphQLBuilder AddUserContextBuilder<TUserContext>(this IGraphQLB
...
+    /// <inheritdoc cref="AddUserContextBuilder{TUserContext}(IGraphQLBuilder, Func{HttpContext, TUserContext})"/>
+    public static IGraphQLBuilder AddUserContextBuilder<TUserContext>(this IGraphQLBuilder builder, Func<HttpContext, object?, Task<TUserContext>> creator)
+        where TUserContext : class, IDictionary<string, object?>
+    {
+        if (creator == null)
+            throw new ArgumentNullException(nameof(creator));
+        builder.Services.Register<IUserContextBuilder>(new UserContextBuilder<TUserContext>((context, payload) => new(creator(context, payload))));
+        builder.ConfigureExecutionOptions(async options =>
+        {
+            if (options.UserContext == null || options.UserContext.Count == 0 && options.UserContext.GetType() == typeof(Dictionary<string, object>))
```
sungam3r, line 115: Why do we use `|| options.UserContext.Count == 0 && options.UserContext.GetType() == typeof(Dictionary<string, object>)` condition?

src/Transports.AspNetCore/Extensions/GraphQLBuilderUserContextExtensions.cs
```diff
@@ -80,6 +80,49 @@ public static IGraphQLBuilder AddUserContextBuilder<TUserContext>(this IGraphQLB
...
+        where TUserContext : class, IDictionary<string, object?>
+    {
+        builder.Services.Register<IUserContextBuilder>(new UserContextBuilder<TUserContext>(creator ?? throw new ArgumentNullException(nameof(creator))));
+        builder.ConfigureExecutionOptions(options =>
+        {
+            if (options.UserContext == null || options.UserContext.Count == 0 && options.UserContext.GetType() == typeof(Dictionary<string, object>))
+            {
+                var requestServices = options.RequestServices ?? throw new MissingRequestServicesException();
+                var httpContext = requestServices.GetRequiredService<IHttpContextAccessor>().HttpContext!;
+                options.UserContext = creator(httpContext, null);
```
sungam3r, line 99: I would pull context builder here and in another method below from DI as `  public static IGraphQLBuilder AddUserContextBuilder<TUserContextBuilder>(this IGraphQLBuilder builder)` does. I have a suspicion that in the case of subsequent registrations of the context builder a situation may arise when options are still configured with the "old" `creator` delegate from closure.

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
Shane32: No. The payload comes in from the WebSocket handler into this protected method so as to make it easy to override this method to change behavior. The WebSocket code hasn’t been merged yet of course. It’s probably a couple thousand lines of code so I’m trying to split it up into smaller pieces for review purposes. It just makes it a little more difficult to see the big picture.

src/Transports.AspNetCore/Extensions/GraphQLBuilderUserContextExtensions.cs
Shane32: Because the execution options are preinitialized with an empty dictionary by default. I think.

src/Transports.AspNetCore/Extensions/GraphQLBuilderUserContextExtensions.cs
Shane32: I rewrote the methods to conform to your suggestion. Please review.

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
sungam3r: Ok

src/Transports.AspNetCore/Extensions/GraphQLBuilderUserContextExtensions.cs
sungam3r: Ah, and `IProvideUserContext` has only getter. Looks a bit awkward but OK.

src/Transports.AspNetCore/Extensions/GraphQLBuilderUserContextExtensions.cs
```diff
@@ -9,75 +9,96 @@ namespace GraphQL.Server;
...
     public static IGraphQLBuilder AddUserContextBuilder<TUserContext>(this IGraphQLBuilder builder, Func<HttpContext, Task<TUserContext>> creator)
         where TUserContext : class, IDictionary<string, object?>
     {
         builder.Services.Register<IUserContextBuilder>(new UserContextBuilder<TUserContext>(context => new(creator(context))));
-        builder.ConfigureExecutionOptions(async options =>
+        builder.Services.TryRegister<IConfigureExecution, UserContextConfigurator>(DI.ServiceLifetime.Singleton, RegistrationCompareMode.ServiceTypeAndImplementationType);
+
+        return builder;
+    }
+
```
sungam3r, line 51:
```suggestion
```

sungam3r (approved):

PR #800, opened 2022-05-20
Enable NRT annotations for Transports.AspNetCore

Shane32:


Shane32:
As there are no code or API changes, I will merge this if all tests pass.  All files except one already was marked with `#nullable enable` anyway - and that one only contained an `internal` class.

Shane32:
@sungam3r If you have any comments we can change them in another PR.  I assume at some point we will want to enable NRT for the entire solution of course.

sungam3r:
> we will want to enable NRT for the entire solution of course.

sooner or later

PR #803, opened 2022-05-21
Replace authorization project with new rule

Shane32:
New features:
- Supports role-based policy rules
- Supports `AllowAnonymous` to allow querying an anonymous field of a protected type
- Introspection queries are treated as anonymous fields; the schema can have authorization rules applied if introspection queries are also to be protected
- Supports `@skip` and `@include` directives properly
- Returns more helpful 'access denied' messages as it pertains to the field or type that failed the check
- Does not reveal the reason that the authentication check failed, as this leaks security information.
- Provides virtual methods which can be overridden to provide more detailed messages.
- `IAuthorizationService` is pulled from the scoped DI, just in case it is a scoped service (although it probably isn't).

Removed features:
- Does not support authorization of input types.  This is because the only practical way to validate input object types is to validate all possible child types that an input object type might contain, due to the fact that input objects are deserialized via `ParseDictionary` within the input object, not to mention they may contain default values.  As such, it is just as effective to protect the field or query argument that references the input type that would trigger the validation check.  This could be done, however, in another PR.
- Does not support `IClaimsPrincipalAccessor`.  This project is for ASP.NET Core and need not support custom authentication schemes.  The GraphQL.Authorization project would be better suited for such a scenario.  The `HttpContext.User` property may be set in application pipeline to change the `ClaimsPrincipal` that validation takes place against.
- Does not support


Shane32:
@sungam3r This PR should be next to concentrate on.

Shane32:
Bump @sungam3r - status?

sungam3r:
I remember this, I will look tomorrow. Is this the last PR of the large changes in the server?

Shane32:
> I remember this, I will look tomorrow. Is this the last PR of the large changes in the server?

Yes. Just samples, tests and documentation after that, which while being a lot of code/text, is not substantive in terms of functionality.

sungam3r:
> Supports @skip and @include directives properly

Were they used incorrectly before? GetRecursivelyReferencedFragments  ?

> Works with any DI provider.

As you said _This project is for ASP.NET Core_ so I think it does not matter.

sungam3r:
I doubt a bit about removed `IAuthorizationErrorMessageBuilder` and `IClaimsPrincipalAccessor` though I understand your intent.

Shane32:
> Were they used incorrectly before? GetRecursivelyReferencedFragments ?

`@skip` and `@include` were ignored previously.  Note that those directives can be applied directly to a field, so it does not matter if `GetRecursivelyReferencedFragments` supports `@skip` and `@include`.

Below is a sample of a failing test that should pass if `@skip` were handled properly.

```cs
[Fact]
public void ignores_skipped_fields()
{
    ConfigureAuthorizationOptions(options =>
    {
        options.AddPolicy("ClassPolicy", x => x.RequireClaim("admin"));
        options.AddPolicy("SchemaPolicy", x => x.RequireClaim("some"));
    });

    ShouldPassRule(config =>
    {
        config.Query = @"query { post @skip(if: true) __typename }";
        config.Schema = BasicSchema<BasicQueryWithAttributesAndClassPolicy>().AuthorizeWithPolicy("SchemaPolicy");
        config.User = CreatePrincipal(claims: new Dictionary<string, string>
        {
            { "some", "true" },
        });
    });
}
```

Shane32:
> I doubt a bit about removed `IAuthorizationErrorMessageBuilder` and `IClaimsPrincipalAccessor` though I understand your intent.

We could probably add them back as optional registrations if you think we should.

## `IClaimsPrincipalAccessor`

There is no easy workaround for `IClaimsPrincipalAccessor`, as for proper support it should be used both in the authorization rule and by the middleware.  I feel that there is two primary use cases right now:

### 1. Authorization is different than the rest of the application.

For example, if JWT authentication is used for the GraphQL endpoint, whereas cookie authentication is used for the web app.  I myself have this scenario which resulted in an ugly hack:

```cs
        // GraphQL requests will ignore ZbDb authentication and use JWT authentication only
        {
            var path = new PathString("/graphql");
            app.MapWhen(
                context => context.Request.Path.Equals(path) &&
                    (context.WebSockets.IsWebSocketRequest || HttpMethods.IsPost(context.Request.Method)),
                b => {
                    // use JWT authentication, and ignore cookie authentication
                    b.Use(next => (RequestDelegate)(async context => {
                        //-- you'd think this would work: --
                        //var user = context.User.Identities.FirstOrDefault(x => x.AuthenticationType == JwtBearerDefaults.AuthenticationScheme);
                        //context.User = new ClaimsPrincipal(user ?? new ClaimsIdentity());

                        //-- and you'd think this would work: --
                        //var auth = context.RequestServices.GetRequiredService<IAuthenticationService>();
                        //await auth.AuthenticateAsync(context, JwtBearerDefaults.AuthenticationScheme);

                        //-- but they don't, so we do this --
                        context.User = new ClaimsPrincipal(new ClaimsIdentity());
                        var authHeader = context.Request.Headers["Authorization"].FirstOrDefault();
                        var authToken = authHeader?.StartsWith("Bearer ") ?? false ? authHeader[7..] : null;
                        if (authToken != null) {
                            var validationParameters = context.RequestServices.GetRequiredService<Microsoft.Extensions.Options.IOptionsMonitor<JwtBearerOptions>>()
                                .Get(JwtBearerDefaults.AuthenticationScheme)
                                .TokenValidationParameters;
                            var handler = new JwtSecurityTokenHandler();
                            try {
                                context.User = handler.ValidateToken(authToken, validationParameters, out _);
                            } catch { }
                        }

                        await next(context).ConfigureAwait(false);
                    }));
                    // apply GraphQL CORS policy
                    b.UseCors("GraphQL");
                    // call graphql
                    b.UseGraphQL(path, opts => {
                        opts.AuthorizationRequired = true;
                    });
                });
        }
```

So I certainly would not be opposed to adding `IClaimsPrincipalAccessor` back.  It's just a bit "anti" ASP.NET Core design principles, as nothing else I know of in ASP.NET Core allows for changing where `IClaimsPrinicpal` is pulled from.

There also is an issue with compatibility in that `IClaimsPrincipalAccessor` currently has a signature as follows:

```cs
/// <summary>
/// Provides access to the <see cref="ClaimsPrincipal"/> used for GraphQL operation authorization.
/// </summary>
public interface IClaimsPrincipalAccessor
{
    /// <summary>
    /// Provides the <see cref="ClaimsPrincipal"/> for the current <see cref="ValidationContext"/>
    /// </summary>
    /// <param name="context">The <see cref="ValidationContext"/> of the current operation</param>
    /// <returns></returns>
    ClaimsPrincipal GetClaimsPrincipal(ValidationContext context);
}
```

Since it requires `ValidationContext`, this presents a problem when used from the middleware.  I would suggest something like this:

```cs
/// <summary>
/// Provides access to the <see cref="ClaimsPrincipal"/> used for GraphQL operation authorization.
/// </summary>
public interface IClaimsPrincipalAccessor
{
    /// <summary>
    /// Provides the <see cref="ClaimsPrincipal"/> for the current <see cref="HttpContext"/>
    /// </summary>
    ValueTask<ClaimsPrincipal> GetClaimsPrincipalAsync(HttpContext context);
}
```

However, the change prevents pulling a `ClaimsPrincipal` from within the actual message.

### 2. WebSocket requests need to be authenticated by the initialization (e.g. GQL_INIT) message.

This situation is now handled by the `IWebSocketAuthenticationService` interface.

If adding back `IClaimsPrincipalAccessor`, we need to be sure it will not interfere.  We could consider adding the new method from the former service to the latter service and remove `IWebSocketAuthenticationService`.  So `IClaimsPrincipalAccessor` would be defined like this:

```cs
/// <summary>
/// Provides access to the <see cref="ClaimsPrincipal"/> used for GraphQL operation authorization.
/// </summary>
public interface IClaimsPrincipalAccessor
{
    /// <summary>
    /// Provides the <see cref="ClaimsPrincipal"/> for the current <see cref="HttpContext"/>.
    /// <br/><br/>
    /// If <see cref="IClaimsPrincipalAccessor"/> is not registered, the value of <see cref="HttpContext.User"/> will be used.
    /// </summary>
    ValueTask<ClaimsPrincipal> GetClaimsPrincipalAsync(HttpContext context);

    /// <summary>
    /// Authenticates an incoming GraphQL over WebSockets request with the connection initialization message.
    /// <br/><br/>
    /// If <see cref="IClaimsPrincipalAccessor"/> is not registered, the value of <see cref="HttpContext.User"/> will be used.
    /// <br/><br/>
    /// After calling this method to authenticate the request, the infrastructure will authorize the incoming request via the
    /// <see cref="GraphQLHttpMiddlewareOptions.AuthorizationRequired"/>, <see cref="GraphQLHttpMiddlewareOptions.AuthorizedRoles"/> and
    /// <see cref="GraphQLHttpMiddlewareOptions.AuthorizedPolicy"/> properties.
    /// </summary>
    ValueTask<ClaimsPrincipal> GetClaimsPrincipalAsync(IWebSocketConnection connection, string subProtocol, OperationMessage operationMessage);
}
```

## `IAuthorizationErrorMessageBuilder`

As for `IAuthorizationErrorMessageBuilder`, the workaround would currently be to use `IErrorInfoProvider` to replace the message.  If we add this interface back, we may want to determine if we also want to add back the old implementation in case users want to use the old messages rather than the new ones.  I would not have the old implementation as default.

## Notes

If we add back either interface, I do want to be sure they do not need to be registered for default operation.

src/Transports.AspNetCore/Extensions/GraphQLBuilderExtensions.cs
```diff
@@ -127,4 +127,15 @@ public static IGraphQLBuilder AddWebSocketAuthentication(this IGraphQLBuilder bu
...
     }
+
+    /// <summary>
+    /// Registers <see cref="AuthorizationValidationRule"/> with the dependency injection framework
+    /// and configures it to be used when executing a request.
+    /// </summary>
+    public static IGraphQLBuilder AddAuthorization(this IGraphQLBuilder builder)
+    {
+        builder.AddValidationRule<AuthorizationValidationRule>(true);
+        builder.Services.TryRegister<IHttpContextAccessor, HttpContextAccessor>(DI.ServiceLifetime.Singleton);
```
sungam3r, line 138: Why not `AddHttpContextAccessor`? Some dependency?

sungam3r: Ah, it's our GraphQL builder.

Shane32:
> Known issues:
> 
> * `ValidationContext.GetRecursivelyReferencedFragments` does not consider `@skip` and `@include` directives, and so authorization checks may be performed on fragments that are not part of the operation due to `@skip` or `@include`.  This should be fixed in GraphQL.NET.

This will be fixed in a subsequent PR.  See https://github.com/Shane32/GraphQL.AspNetCore3/pull/50

src/Transports.AspNetCore/AuthorizationValidationRule.cs
```diff
@@ -0,0 +1,32 @@
...
+        _contextAccessor = httpContextAccessor ?? throw new ArgumentNullException(nameof(httpContextAccessor));
+    }
+
+    /// <inheritdoc/>
+    public ValueTask<INodeVisitor?> ValidateAsync(ValidationContext context)
+    {
+        var httpContext = _contextAccessor.HttpContext
+            ?? throw new InvalidOperationException("HttpContext could not be retrieved from IHttpContextAccessor.");
+        var user = httpContext.User
+            ?? throw new InvalidOperationException("ClaimsPrincipal could not be retrieved from HttpContext.User.");
```
sungam3r, line 22:
```suggestion
            ?? throw new InvalidOperationException("ClaimsPrincipal could not be retrieved from HttpContext.");
```

tests/Authorization.AspNetCore.Tests/ValidationTestBase.cs
```diff
@@ -92,7 +93,8 @@ private IValidationResult Validate(ValidationTestConfig config)
             Document = document,
             Operation = document.Definitions.OfType<GraphQLOperationDefinition>().First(),
             Rules = config.Rules,
-            Variables = config.Variables
+            Variables = config.Variables,
+            RequestServices = ServiceProvider,
```
sungam3r, line 97: `ServiceProvider` is root service provider. Maybe create and set service provider from scope ?

src/Transports.AspNetCore/AuthorizationVisitorBase.Validation.cs
```diff
@@ -0,0 +1,214 @@
...
+            parentFieldType = context.TypeInfo.GetFieldDef();
+            parentGraphType = context.TypeInfo.GetLastType(1)?.GetNamedType();
+        }
+        return new(obj, node, parentFieldType, parentGraphType, context);
+    }
+
+    /// <summary>Provides contextual information to the schema, graph, field, or query argument being validated.</summary>
+    /// <param name="Obj">The schema, graph type, field type, or query argument being validated. May be an interface type if fragments are in use.</param>
+    /// <param name="Node">Null for a schema validation; otherwise the <see cref="GraphQLOperationDefinition"/>, <see cref="GraphQLField"/>, or <see cref="GraphQLArgument"/> being validated.</param>
+    /// <param name="Context">The validaion context; but <see cref="ValidationContext.TypeInfo"/> may not be applicable for node being validated.</param>
```
sungam3r, line 45:
```suggestion
    /// <param name="Context">The validation context; but <see cref="ValidationContext.TypeInfo"/> may not be applicable for node being validated.</param>
```

src/Transports.AspNetCore/AuthorizationVisitorBase.Validation.cs
```diff
@@ -0,0 +1,214 @@
...
+    {
+        var resource = GenerateResourceDescription(info);
+        var err = info.Node == null ? new AccessDeniedError(resource) : new AccessDeniedError(resource, info.Context.Document.Source, info.Node);
+        err.RolesRequired = roles;
+        info.Context.ReportError(err);
+    }
+
+    /// <summary>
+    /// Adds a error to the validation context indicating that the user is not a member of any of
+    /// the roles required by this graph, field or query argument.
```
sungam3r, line 163: copy-pasted from HandleNodeNotInRoles

src/Transports.AspNetCore/AuthorizationVisitorBase.Validation.cs
```diff
@@ -0,0 +1,214 @@
...
+                return $"type '{graphType.Name}' for {op.Operation.ToString().ToLower(System.Globalization.CultureInfo.InvariantCulture)} operation{(!string.IsNullOrEmpty(op.Name?.StringValue) ? $" '{op.Name}'" : null)}";
+            }
+            else
+            {
+                return $"type '{graphType.Name}'";
+            }
+        }
+        else if (info.Obj is IFieldType fieldType)
+        {
+            return $"field '{fieldType.Name}' on type '{info.ParentGraphType?.Name}'";
```
sungam3r, line 203: Consider to use more precise schema coordinates syntax https://github.com/graphql/graphql-spec/pull/794

src/Transports.AspNetCore/AuthorizationVisitor.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// </summary>
+    protected ClaimsPrincipal ClaimsPrincipal { get; }
+
+    /// <summary>
+    /// Gets the authorization service that is used to authorize policy requests.
+    /// </summary>
+    protected IAuthorizationService AuthorizationService { get; }
+
+    /// <inheritdoc/>
+    protected override bool Authorize()
```
sungam3r, line 27: Method looks like an action to perform (verb). The same for AuthorizeRole. Do they definitely have such names?

sungam3r (commented):
I looked through all changes. What can I say? Well, the previous validation rule takes about 200 lines of code and is quite understandable. The current implementation is greatly complicated and overloaded. Honestly, I almost immediately lost my desire to understand the code. I watched it not in detail and I'm sure that this code will be more difficult to maintain. In addition, I still doubt the correctness of the selected approach with building error messages (removed message builder) though I'm fine with any way to assemble those messages. Howewer, in this situation, I am pleased that despite such complexity, it is limited by only one rule of validation and does not go beyond it. See my other comments about schema coordinates and other and fell free to merge since I can't help here more reviewing/tracking control flow.

sungam3r:
Of course, as a user of this rule, I would need comprehensive documentation.

src/Transports.AspNetCore/AuthorizationVisitorBase.Validation.cs
Shane32: Fixed

src/Transports.AspNetCore/AuthorizationVisitorBase.Validation.cs
Shane32: #835

src/Transports.AspNetCore/AuthorizationVisitor.cs
Shane32: I named them so they all start with `Authorize`, and so they are similar to the extension methods we have on defined within GraphQL.NET, which are `Authorize`, `AuthorizeWithRoles` and `AuthorizeWithPolicy`.  We could name them similar to the members they map to:

- `bool IsAuthenticated { get; }`,
- `bool IsInRole(string role)`, and
- `AuthorizationResult Authorize(string policy)`

What do you think?  I'm open to suggestions and would like to do whatever makes the most sense.

For reference here is the code:

```cs
protected override bool Authorize()
    => ClaimsPrincipal.Identity?.IsAuthenticated ?? false;

protected override bool AuthorizeRole(string role)
    => ClaimsPrincipal.IsInRole(role);

protected override AuthorizationResult AuthorizePolicy(string policy)
    => AuthorizationService.AuthorizeAsync(ClaimsPrincipal, policy).GetAwaiter().GetResult();
```

Shane32:
> Well, the previous validation rule takes about 200 lines of code and is quite understandable. The current implementation is greatly complicated and overloaded.

I agree.  The nature of properly applying the `AllowAnonymous` attribute is complex to implement.  I've added many comments but still it is not easy to follow.  The only other thing I could do is to break it up into additional methods, but I don't think it will make it much easier to understand.

> See my other comments about schema coordinates

I looked at the spec you referenced.  I certainly see the value in using the spec here.  I added an issue so we can come back to this.  As this is a implementation detail (being an error message description) changing it after release is not a breaking change.  I will review again when I have time; maybe soon.

> feel free to merge

I will await your comments on the `Authorize` messages before merging.  We should certainly choose the best names we can.

src/Transports.AspNetCore/AuthorizationVisitor.cs
sungam3r: It is bad that for all of them you can not use the name `IsXXX`. The last one method is verb. There is no 100% symmetry. I don't know how to do better.


    bool IsAuthenticated { get; },
    bool IsInRole(string role), and
    AuthorizationResult Authorize(string policy)

This variant at least looks reasonable.

Shane32: Will do.  At least it matches the methods they call (by default).

Shane32:
With the vast number of changes made in this project as a whole (basically a 100% rewrite), maybe it would be better to retain `IClaimsPrincipal` and the authorization message builder code as `[Obsolete]` services.  At least it gives everyone a chance to open an issue if they feel there is a need or value in the old code before it is completely removed.  Then we can remove in v8 if there are no issues -- or we can remove the `[Obsolete]` tag if there is value in those interfaces.

tests/Authorization.AspNetCore.Tests/ValidationTestBase.cs
sungam3r, line 97:
```suggestion
            RequestServices = ServiceProvider, // root provider here instead of scoped one, but since nothing registered is scoped, it makes a little difference so we just use rooot provider directly
```

Shane32:
> With the vast number of changes made in this project as a whole (basically a 100% rewrite), maybe it would be better to retain `IClaimsPrincipal` and the authorization message builder code as `[Obsolete]` services. At least it gives everyone a chance to open an issue if they feel there is a need or value in the old code before it is completely removed. Then we can remove in v8 if there are no issues -- or we can remove the `[Obsolete]` tag if there is value in those interfaces.

Another option would be to update the old nuget package `GraphQL.Server.Authorization.AspNetCore` to use the new code here, inheriting all of the new features, but still retaining (within that package) the `IClaimsPrincipal` and message builder code.  The entire package can be marked as deprecated, but for the immediate transition to v7 it will be a drop-in replacement.  I anticipate no changes to the code within this `GraphQL.Server.Transports.AspNetCore` package at all.  Note that since GraphQL.NET's `INodeVisitor` is changing for v7, any old validation rules will not work anymore.  So users cannot use GraphQL.Server v6 authorization rule with GraphQL.NET v7.

sungam3r:
Do you mean something like `[Obsolete("This type is not used anymore now and will be removed in v8. Instead use XXX. If you think you really need it, then feel free to open an issue to return it back")]` ?

Shane32:
> Do you mean something like [Obsolete("This type is not used anymore now and will be removed in v8. Instead use XXX. If you think you really need it, then feel free to open an issue to return it back")] ?

Sure; something like that.

sungam3r (approved):

PR #804, opened 2022-05-22
Add WebSocketWriterStream class

Shane32:
Required for writing JSON to a WebSocket as a message.  It is set up so that the JSON serializer will write to this class, which will write to the underlying WebSocket connection, and then a call to FlushAsync will inform the WebSocket that the message has completed.

This is largely a copy of the existing `GraphQL.Server.Transports.WebSockets.WebsocketWriterStream` class.


Shane32:

sungam3r (approved):

PR #805, opened 2022-05-22
Add ReusableMemoryReaderStream class

Shane32:
This is an optimization for deserializing JSON messages for WebSocket support.

In the message loop, WebSocket messages are read into a buffer (byte array), and then can be read by the installed GraphQL.NET serialization engine.  However, the serializer only accepts streams, not byte arrays with offset & length.  So while we could create a memory stream for each message, this would cause an allocation for each message (the MemoryStream itself), even if the actual memory data was not copied (which it might be; can't remember).

Note that we cannot use MemoryStream with GetBuffer because the JSON deserializer will throw an exception saying that data exists after the end of the JSON data.  It must encounter the end of the stream to properly deserialize.

So this class maps to a buffer and can be reset to Position 0 with a settable length depending on the length of the message received.

### DIFF
--- a/src/Transports.AspNetCore/WebSockets/ReusableMemoryReaderStream.cs
+++ b/src/Transports.AspNetCore/WebSockets/ReusableMemoryReaderStream.cs
@@ -1,0 +1,106 @@
+namespace GraphQL.Server.Transports.AspNetCore.WebSockets;
+
+/// <summary>
+/// A readable memory stream based on a buffer, with a resettable maximum length.
+/// </summary>
+internal class ReusableMemoryReaderStream : Stream
+{
+    private readonly byte[] _buffer;
+    private int _position;
+    private int _length;
+
+    /// <summary>
+    /// Initializes a new instance based on the specified buffer.
+    /// </summary>
+    public ReusableMemoryReaderStream(byte[] buffer)
+    {
+        _buffer = buffer ?? throw new ArgumentNullException(nameof(buffer));
+    }
+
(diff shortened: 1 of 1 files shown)


Shane32:

sungam3r:
I'll review this one and other PRs that evening.

sungam3r (approved):
Did you consider https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream ?

sungam3r:
Regarding my promise https://github.com/graphql-dotnet/server/pull/805#issuecomment-1139421420 - unfortunately, I didn’t have time. I will have a few hours tomorrow morning (since 6 AM) to carefully review everything left.

Shane32:
> Regarding my promise [#805 (comment)](https://github.com/graphql-dotnet/server/pull/805#issuecomment-1139421420) - unfortunately, I didn’t have time. I will have a few hours tomorrow morning (since 6 AM) to carefully review everything left.

No problem. I was really just asking for an ETA on the other outstanding PRs to help manage expectations.

Shane32:
> Did you consider https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream ?

No, but thanks for the link - I’ll review. It may be useful when the WebSocket message does not fit in the preallocated buffer.

sungam3r:
I use `RecyclableMemoryStream` in some services in production. This is a fairly simple concept and detailed documentation is written for the project.

PR #806, opened 2022-05-22
Add SubscriptionList class for WebSocket support

Shane32:
This class manages active subscriptions for a given GraphQL WebSocket connection.  The class is fully thread-safe and all methods are necessary.

It functions somewhat like a `ConcurrentDictionary`, except that it also disposes of removed/replaced `IDisposable` instances.

Note:
- `TryAdd` is needed as one of the two current protocols must reject duplicate IDs
- `this[string id] { set; }` is needed the other current protocol must override duplicate IDs
- `CompareExchange` is needed as there is a time during which the id is reserved before it has been completely initialized

This class is somewhat similar to the old `GraphQL.Server.Transports.Subscriptions.Abstractions.SubscriptionManager` class, except it simply manages a list of subscriptions.


Shane32:

sungam3r:
I reviewed two other small PRs this morning and will return to this one today some time after.

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
```diff
@@ -0,0 +1,228 @@
...
+    {
+        List<IDisposable> subscriptionsToDispose;
+        lock (_lock)
+        {
+            subscriptionsToDispose = _subscriptions.Values.ToList();
+            _subscriptions.Clear();
+        }
+        foreach (var disposer in subscriptionsToDispose)
+        {
+            disposer.Dispose();
```
sungam3r, line 40: 🤔 Should each ` disposer.Dispose();` be under its own try/catch?

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
```diff
@@ -0,0 +1,228 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Adds a subscription to the internal list if none is already in the list with the same id.
+    /// </summary>
+    /// <exception cref="ArgumentNullException"/>
+    /// <exception cref="OperationCanceledException"/>
+    public bool TryAdd(string id, IDisposable subscription)
```
sungam3r, line 49: I would write a note about _id_ argument mentioning that this is the subscription identifier or something.

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
```diff
@@ -0,0 +1,228 @@
...
+        lock (_lock)
+        {
+            _cancellationToken.ThrowIfCancellationRequested();
+            return _subscriptions.TryAdd(id, subscription);
+        }
+    }
+
+    /// <summary>
+    /// Adds a subscription to the internal list, overwriting an existing registration if any.
+    /// When overwriting an existing registration, the old registration is disposed.
```
sungam3r, line 65: Will it happen in practice?

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
```diff
@@ -0,0 +1,228 @@
...
+            }
+        }
+    }
+
+    /// <summary>
+    /// Validates that the specified subscription is still active.
+    /// </summary>
+    /// <exception cref="ArgumentNullException"/>
+    /// <exception cref="OperationCanceledException"/>
+    public bool Contains(string id, IDisposable subscription)
```
sungam3r, line 100: Do you really need _subscription_ parameter here?

sungam3r: I see second overload without _subscription_ parameter. Does this one needed?

sungam3r: The same question about `TryRemove`.

sungam3r:
изображение

I would add these notes into xml comments. Recently, I have been more and more inclined to add all significant thoughts in the code, so that later it is easier to understand and fix.

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
```diff
@@ -0,0 +1,228 @@
...
+    /// <summary>
+    /// Initializes a new instance.
+    /// </summary>
+    public SubscriptionList(CancellationToken cancellationToken)
+    {
+        _cancellationToken = cancellationToken;
+    }
+
+    /// <summary>
+    /// Disposes of all active subscriptions.
```
sungam3r, line 28:
```suggestion
    /// Disposes all subscriptions in the list.
```

Formally, they do not have to be active because they can be disposed from the outside.

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
```diff
@@ -0,0 +1,228 @@
...
+    /// </summary>
+    public void Dispose()
+    {
```
sungam3r, line 31:
```suggestion
    {
        if (_cancellationToken.IsCancellationRequested)
            return;
```

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
```diff
@@ -0,0 +1,228 @@
...
+    public bool TryAdd(string id, IDisposable subscription)
+    {
+        if (id == null)
+            throw new ArgumentNullException(nameof(id));
+        if (subscription == null)
+            throw new ArgumentNullException(nameof(subscription));
+
+        lock (_lock)
+        {
+            _cancellationToken.ThrowIfCancellationRequested();
```
sungam3r, line 58: Should this line be under lock statement? The same question for all other methods.

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
```diff
@@ -0,0 +1,228 @@
...
+            }
+            finally
+            {
+                oldDisposable?.Dispose();
+            }
+        }
+    }
+
+    /// <summary>
+    /// Validates that the specified subscription is still active.
```
sungam3r, line 96:
```suggestion
    /// Validates that the specified subscription is still in the list.
```

I would avoid term _active_ .

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
```diff
@@ -0,0 +1,228 @@
...
+
+        lock (_lock)
+        {
+            _cancellationToken.ThrowIfCancellationRequested();
+            return _subscriptions.TryGetValue(id, out var value) && value == subscription;
+        }
+    }
+
+    /// <summary>
+    /// Validates that the specified subscription is still active.
```
sungam3r, line 115:
```suggestion
    /// Validates that the specified subscription is still in the list.
```

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
```diff
@@ -0,0 +1,228 @@
...
+    /// </summary>
+    public SubscriptionList(CancellationToken cancellationToken)
+    {
+        _cancellationToken = cancellationToken;
+    }
+
+    /// <summary>
+    /// Disposes of all active subscriptions.
+    /// </summary>
+    public void Dispose()
```
sungam3r, line 30: If one calls `Dispose` (without signalling CancellationToken outside) then this class is still functioning.

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
Shane32: That is determined by the client.

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
Shane32: I am using these methods in my implementation, so that if a long running operation is cancelled and another started with the same ID, the initial operation does not send a response when it finally completes. It may be a rather unusual situation, and perhaps may never occur, but a valid one.

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
Shane32: It is EXCEEDINGLY important that all subscriptions are disposed of, since any not disposed of are GUARANTEED memory leaks. The GC will not collect lost objects. As such, most likely I wrote it this way to be sure the list was not shut down when adding or modifying entries. You can be sure I considered the question.

Typically a check like this might go on both sides of the lock, but here the redundant lock is unlikely to occur, so I implemented only the required check.

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
Shane32: Yup. This class is a bit weird having both a cancellation token and dispose to shut it down, and certainly abnormal. I was thinking maybe I should implement a linked cancellation token so that dispose can signal it. Another option is to register a call to dispose from the cancellation token. I’ll think about it some more today.

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
Shane32: Good idea. Done

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
Shane32: I reviewed all these conditions again, and added disposed checks.

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
Shane32: I worked on this for a while and ended up simply adding dispose checks.

Shane32: Note that I do not want to remove the cancellation token because it is important for an `OperationCanceledException` thrown under normal circumstances.  If you want to re-review this decision, we should do so during the review of #810

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
Shane32: I altered the class comment.

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
Shane32: I spent a few more hours on this, trying to determine if the cancellation token was necessary at all within this class.  I determined that it was not.  I made adjustments within GraphQL.AspNetCore3 including changes to tests to verify that everything works properly without the cancellation token.  Then I made the changes here.  So now this class has a very typical disposal pattern.

See PR:
- https://github.com/Shane32/GraphQL.AspNetCore3/pull/34

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
Shane32: As noted above, I have determined that the cancellation token is not necessary, and as such was removed from the class.

sungam3r:
I'll continue tomorrow morning.

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
```diff
@@ -0,0 +1,232 @@
+namespace GraphQL.Server.Transports.AspNetCore.WebSockets;
+
+/// <summary>
+/// A list of subscriptions represented by a dictonary of string keys and <see cref="IDisposable"/> values.
+/// All members are thread-safe.
+/// <br/><br/>
+/// Upon signalling the cancellation token, all future calls to any method except
+/// <see cref="Dispose"/> will throw an <see cref="OperationCanceledException"/>.
```
sungam3r, line 8: Is not true anymore.

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
```diff
@@ -0,0 +1,232 @@
+namespace GraphQL.Server.Transports.AspNetCore.WebSockets;
+
+/// <summary>
+/// A list of subscriptions represented by a dictonary of string keys and <see cref="IDisposable"/> values.
```
sungam3r, line 4:
```suggestion
/// A list of subscriptions represented by a dictionary of string keys (subscription ids) and <see cref="IDisposable"/> values.
```

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
```diff
@@ -0,0 +1,232 @@
...
+/// All members are thread-safe.
+/// <br/><br/>
+/// Upon signalling the cancellation token, all future calls to any method except
+/// <see cref="Dispose"/> will throw an <see cref="OperationCanceledException"/>.
+/// <br/><br/>
+/// It is recommended to call <see cref="Dispose"/> immediately after signalling
+/// the cancellation token.
```
sungam3r, line 11: Is not true anymore.

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
```diff
@@ -0,0 +1,232 @@
...
+            subscriptionsToDispose = _subscriptions.Values.ToList();
+            _subscriptions = null;
+        }
+        foreach (var disposer in subscriptionsToDispose)
+        {
+            try
+            {
+                disposer.Dispose();
+            }
+            catch { }
```
sungam3r, line 37: I would aggregate failed disposables and eventually throw to give the caller side to know about problem instead of silently swallow all exceptions.

sungam3r (approved):
3 final suggestions left

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
Shane32: Deleted inapplicable comments.

src/Transports.AspNetCore/WebSockets/SubscriptionList.cs
Shane32: Done

Shane32:
> 3 final suggestions left

Fixed all 3

PR #802, opened 2022-05-20
WebSocket options prep

Shane32:
Requires #799

PR #807, opened 2022-05-22
Add interfaces for WebSocket connections

Shane32:
Adds the basic interfaces for GraphQL WebSocket connection support:

This consists of three interfaces:
- `IWebSocketConnection` is responsible for reading from and writing to the underlying `WebSocket`, including serialization and deserialization, maintaining thread safety for outbound messages, and disposing of the `IOperationMessageProcessor` when the connection terminates.  Based on the assumption that all incoming and outgoing packets are serializable to/from the `OperationMessage` class.
- `IOperationMessageProcessor` performs all logic associated with a connection.
- `IWebSocketAuthenticationService` performs authentication in response to a ConnectionInit message.  This is required since browsers cannot send HTTP `Authorize` headers when opening a WebSocket connection, and so the two current GraphQL over WebSockets protocols in use today allow for authentication via the payload of the initialization message, which is required before any other communication occurs on the WebSocket channel.

Basic connection initialization is performed as follows:
1. Create `IWebSocketConnection` implementation with handle to the specific ASP.NET `WebSocket`.  This implementation now has `SendMessageAsync` and `CloseConnectionAsync` methods available for sending messages to the client.
2. Create `IOperationMessageProcessor` implementation with handle to the `IWebSocketConnection`.
3. Call `IWebSocketConnection.ExecuteAsync` is called with the `IOperationMessageProcessor` instance.

`ExecuteAsync` first calls `IOperationMessageProcessor.InitializeConnectionAsync` so that any connection timers can be initialized.  `ExecuteAsync` also contains the "message loop", reading messages from the underlying `WebSocket` and passing them to `IOperationMessageProcessor.OnMessageReceivedAsync`.

Note that in PR #802 I added `IWebSocketHandler`.  The handler is responsible for setting up the appropriate `IWebSocketConnection` and `IOperationMessageProcessor` class, depending on the requested sub-protocols, allowing for selecting a sub-protocol based on which one is most preferred by the server.  Any protocol that does not use `OperationMessage`s for communication can use a different implementation.  The GraphQL HTTP middleware supports multiple installed `IWebSocketHandler` instances, so a sub-protocol can be added or overridden without affecting any of the existing code.

### DIFF
--- a/src/Transports.AspNetCore/WebSockets/IOperationMessageProcessor.cs
+++ b/src/Transports.AspNetCore/WebSockets/IOperationMessageProcessor.cs
@@ -1,0 +1,21 @@
+using GraphQL.Transport;
+
+namespace GraphQL.Server.Transports.AspNetCore.WebSockets;
+
+/// <summary>
+/// Processes a stream of messages received from a WebSockets client.
+/// <see cref="IDisposable.Dispose"/> must be called when the connection terminates.
+/// Methods defined within this interface need not be thread-safe.
+/// </summary>
+public interface IOperationMessageProcessor : IDisposable
+{
+    /// <summary>
+    /// Starts the connection initialization timer, if configured.
+    /// </summary>
+    Task InitializeConnectionAsync();
+
+    /// <summary>
+    /// Called when a message is received from the client.
+    /// </summary>
+    Task OnMessageReceivedAsync(OperationMessage message);
+}
--- a/src/Transports.AspNetCore/WebSockets/IOperationMessageProcessor.cs
+++ b/src/Transports.AspNetCore/WebSockets/IOperationMessageProcessor.cs
@@ -4,7 +4,8 @@ namespace GraphQL.Server.Transports.AspNetCore.WebSockets;
 
 /// <summary>
 /// Processes a stream of messages received from a WebSockets client.
-/// <see cref="IDisposable.Dispose"/> must be called when the connection terminates.
+/// <see cref="IDisposable.Dispose"/> must be called when the WebSocket connection
+/// has received a close message, and/or when the connection terminates.
 /// Methods defined within this interface need not be thread-safe.
 /// </summary>
 public interface IOperationMessageProcessor : IDisposable
--- a/src/Transports.AspNetCore/WebSockets/IWebSocketAuthenticationService.cs
+++ b/src/Transports.AspNetCore/WebSockets/IWebSocketAuthenticationService.cs
@@ -1,0 +1,25 @@
+using GraphQL.Transport;
+using Microsoft.AspNetCore.Http;
+
+namespace GraphQL.Server.Transports.AspNetCore.WebSockets;
+
+/// <summary>
+/// Authenticates an incoming GraphQL over WebSockets request with the
+/// connection initialization message.  A typical implementation will
+/// set the <see cref="HttpContext.User"/> property after reading the
+/// authorization token.  This service must be registered as a singleton
+/// in the dependency injection framework.
+/// </summary>
+public interface IWebSocketAuthenticationService
+{
+    /// <summary>
+    /// Authenticates an incoming GraphQL over WebSockets request with the connection initialization message.  The implementation should
+    /// set the <paramref name="connection"/>.<see cref="IWebSocketConnection.HttpContext">HttpContext</see>.<see cref="HttpContext.User">User</see>
+    /// property after validating the provided credentials.
+    /// <br/><br/>
+    /// After calling this method to authenticate the request, the infrastructure will authorize the incoming request via the
+    /// <see cref="GraphQLHttpMiddlewareOptions.AuthorizationRequired"/>, <see cref="GraphQLHttpMiddlewareOptions.AuthorizedRoles"/> and
+    /// <see cref="GraphQLHttpMiddlewareOptions.AuthorizedPolicy"/> properties.
+    /// </summary>
+    Task AuthenticateAsync(IWebSocketConnection connection, string subProtocol, OperationMessage operationMessage);
+}
--- a/src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
+++ b/src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
@@ -1,0 +1,50 @@
+using GraphQL.Transport;
+using Microsoft.AspNetCore.Http;
+
+namespace GraphQL.Server.Transports.AspNetCore.WebSockets;
+
+/// <summary>
+/// Represents a WebSocket connection, dispatching received messages over
+/// the connection to the specified <see cref="IOperationMessageProcessor"/>,
+/// and sending requested messages out the connection when requested.
+/// <br/><br/>
+/// <see cref="ExecuteAsync(IOperationMessageProcessor)"/> may be only called once.
+/// All members must be thread-safe.
+/// </summary>
+public interface IWebSocketConnection
+{
+    /// <summary>
+    /// Listens to incoming messages over the WebSocket connection,
+    /// dispatching the messages to the specified <paramref name="operationMessageProcessor"/>.
+    /// Returns or throws <see cref="OperationCanceledException"/> when the WebSocket connection is closed.
+    /// </summary>
+    Task ExecuteAsync(IOperationMessageProcessor operationMessageProcessor);
+
+    /// <summary>
+    /// Sends a message.
+    /// </summary>
+    Task SendMessageAsync(OperationMessage message);
+
+    /// <summary>
+    /// Closes the WebSocket connection.
+    /// </summary>
+    Task CloseConnectionAsync();
+
+    /// <summary>
+    /// Closes the WebSocket connection with the specified error information.
+    /// </summary>
+    Task CloseConnectionAsync(int eventId, string? description);
+
+    /// <summary>
+    /// Returns the last UTC time that a message was sent.
+    /// </summary>
+    DateTime LastMessageSentAt { get; }
+
+    /// <inheritdoc cref="HttpContext.RequestAborted"/>
+    CancellationToken RequestAborted { get; }
+
+    /// <summary>
+    /// Returns the <see cref="Microsoft.AspNetCore.Http.HttpContext"/> associated with this request.
+    /// </summary>
+    HttpContext HttpContext { get; }
+}
--- a/src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
+++ b/src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
@@ -11,7 +11,7 @@ namespace GraphQL.Server.Transports.AspNetCore.WebSockets;
 /// <see cref="ExecuteAsync(IOperationMessageProcessor)"/> may be only called once.
 /// All members must be thread-safe.
 /// </summary>
-public interface IWebSocketConnection
+public interface IWebSocketConnection : IDisposable
 {
     /// <summary>
     /// Listens to incoming messages over the WebSocket connection,
@@ -21,17 +21,17 @@ public interface IWebSocketConnection
     Task ExecuteAsync(IOperationMessageProcessor operationMessageProcessor);
 
     /// <summary>
-    /// Sends a message.
+    /// Sends a message to the client.
     /// </summary>
     Task SendMessageAsync(OperationMessage message);
 
     /// <summary>
-    /// Closes the WebSocket connection.
+    /// Performs a graceful shutdown of the WebSocket connection with event ID 1000.
     /// </summary>
     Task CloseConnectionAsync();
 
     /// <summary>
-    /// Closes the WebSocket connection with the specified error information.
+    /// Performs a graceful shutdown of the WebSocket connection with the specified error information.
     /// </summary>
     Task CloseConnectionAsync(int eventId, string? description);
 
--- a/src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
+++ b/src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
@@ -36,7 +36,7 @@ public interface IWebSocketConnection : IDisposable
     Task CloseAsync(int eventId, string? description);
 
     /// <summary>
-    /// Returns the last UTC time that a message was sent.
+    /// Returns the last UTC time that a message was sent to the client. This property is needed to implement the idle timeout to only send a keep-alive message after ## seconds of "no traffic". 
     /// </summary>
     DateTime LastMessageSentAt { get; }
 
--- a/src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
+++ b/src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
@@ -36,7 +36,8 @@ public interface IWebSocketConnection : IDisposable
     Task CloseAsync(int eventId, string? description);
 
     /// <summary>
-    /// Returns the last UTC time that a message was sent to the client. This property is needed to implement the idle timeout to only send a keep-alive message after ## seconds of "no traffic". 
+    /// Returns the last UTC time that a message was sent to the client.
+    /// This property is needed to implement the idle timeout to only send a keep-alive message after ## seconds of "no traffic". 
     /// </summary>
     DateTime LastMessageSentAt { get; }
 
--- a/src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
+++ b/src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
@@ -28,7 +28,7 @@ public interface IWebSocketConnection : IDisposable
     /// <summary>
     /// Performs a graceful shutdown of the WebSocket connection with event ID 1000.
     /// </summary>
-    Task CloseConnectionAsync();
+    Task CloseAsync();
 
     /// <summary>
     /// Performs a graceful shutdown of the WebSocket connection with the specified error information.
--- a/src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
+++ b/src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
@@ -33,7 +33,7 @@ public interface IWebSocketConnection : IDisposable
     /// <summary>
     /// Performs a graceful shutdown of the WebSocket connection with the specified error information.
     /// </summary>
-    Task CloseConnectionAsync(int eventId, string? description);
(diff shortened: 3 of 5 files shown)


Shane32:

src/Transports.AspNetCore/WebSockets/IOperationMessageProcessor.cs
```diff
@@ -0,0 +1,21 @@
...
+
+/// <summary>
+/// Processes a stream of messages received from a WebSockets client.
+/// <see cref="IDisposable.Dispose"/> must be called when the connection terminates.
+/// Methods defined within this interface need not be thread-safe.
+/// </summary>
+public interface IOperationMessageProcessor : IDisposable
+{
+    /// <summary>
+    /// Starts the connection initialization timer, if configured.
```
sungam3r, line 13: connection initialization timer?

src/Transports.AspNetCore/WebSockets/IOperationMessageProcessor.cs
```diff
@@ -0,0 +1,21 @@
+using GraphQL.Transport;
+
+namespace GraphQL.Server.Transports.AspNetCore.WebSockets;
+
+/// <summary>
+/// Processes a stream of messages received from a WebSockets client.
+/// <see cref="IDisposable.Dispose"/> must be called when the connection terminates.
```
sungam3r, line 7: Do you mean IWebSocketConnection? If so I would add comment that _usually IWebSocketConnection implementation disposes this object_.

src/Transports.AspNetCore/WebSockets/IOperationMessageProcessor.cs
sungam3r: Also I suspect there is some initialization context (at least connection object itself). How are you going to make a implementation?

src/Transports.AspNetCore/WebSockets/IOperationMessageProcessor.cs
```diff
@@ -0,0 +1,21 @@
...
+/// Methods defined within this interface need not be thread-safe.
+/// </summary>
+public interface IOperationMessageProcessor : IDisposable
```
sungam3r, line 10: Will `IAsyncDisposable` be useful here?

src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
```diff
@@ -0,0 +1,50 @@
...
+{
+    /// <summary>
+    /// Listens to incoming messages over the WebSocket connection,
+    /// dispatching the messages to the specified <paramref name="operationMessageProcessor"/>.
+    /// Returns or throws <see cref="OperationCanceledException"/> when the WebSocket connection is closed.
+    /// </summary>
+    Task ExecuteAsync(IOperationMessageProcessor operationMessageProcessor);
+
+    /// <summary>
+    /// Sends a message.
```
sungam3r, line 24:
```suggestion
    /// Sends a message to the client.
```

src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// </summary>
+    Task SendMessageAsync(OperationMessage message);
+
+    /// <summary>
+    /// Closes the WebSocket connection.
+    /// </summary>
+    Task CloseConnectionAsync();
```
sungam3r, line 31: `IOperationMessageProcessor` implements `IDisposable` but here you use dedicated method to close connection. What about `Dispose` here?

sungam3r, line 31:
```suggestion
    Task CloseAsync();
```

since the context is clear

src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
```diff
@@ -0,0 +1,50 @@
...
+
+    /// <summary>
+    /// Closes the WebSocket connection with the specified error information.
+    /// </summary>
+    Task CloseConnectionAsync(int eventId, string? description);
```
sungam3r, line 36:
```suggestion
    Task CloseAsync(int eventId, string? description);
```

src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
sungam3r: Also `CloseAsync` in interface looks excessive - there is always overload with two parameters. I would implement parameterless `CloseAsync()` method as extension method.

src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
```diff
@@ -0,0 +1,50 @@
...
+
+    /// <summary>
+    /// Returns the last UTC time that a message was sent.
```
sungam3r, line 39: From server to client? Why do you need it?

src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// </summary>
+    DateTime LastMessageSentAt { get; }
+
+    /// <inheritdoc cref="HttpContext.RequestAborted"/>
+    CancellationToken RequestAborted { get; }
```
sungam3r, line 44: Why do you need it? `HttpContext` is already here.

sungam3r: Or is that `HttpContext` Initial one where connection was created?

src/Transports.AspNetCore/WebSockets/IWebSocketAuthenticationService.cs
```diff
@@ -0,0 +1,25 @@
+using GraphQL.Transport;
+using Microsoft.AspNetCore.Http;
+
+namespace GraphQL.Server.Transports.AspNetCore.WebSockets;
+
+/// <summary>
+/// Authenticates an incoming GraphQL over WebSockets request with the
+/// connection initialization message.  A typical implementation will
```
sungam3r, line 8: What is _ connection initialization message_ ?

src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
sungam3r: I guess this method should send special message for client. So what about lifetime of connection after this method being called?

src/Transports.AspNetCore/WebSockets/IOperationMessageProcessor.cs
Shane32: > connection initialization timer

Starts the timer that closes the connection if no ConnectionInit message is received after ConnectionInitWaitTimeout 

See: https://github.com/graphql-dotnet/server/blob/develop/src/Transports.AspNetCore/WebSockets/GraphQLWebSocketOptions.cs#L8..L12

> some initialization context

The initialization message does not need to create context at all -- its only purpose is to authenticate the connection (since browsers cannot send a HTTP Authorize header with the connection).  Per the protocol, any other message is refused until this "ConnectionInit" message is received and acknowledged.

> connection object itself

Yes, the `IOperationMessageProcessor` holds state.  Note that it is not possible for two simultaneous connections over a single WebSocket connection.  So it's basically just a state machine along with a list of active subscribers (see #806 ) and a few background timers.  See #810 for implementation.

src/Transports.AspNetCore/WebSockets/IOperationMessageProcessor.cs
Shane32: No the owner of the `IOperationMessageProcessor`, which will be the `IWebSocketHandler` implementation, will dispose of the objects it creates.

See https://github.com/graphql-dotnet/server/pull/812/files#diff-3c1de1415b9d3b97edb7c74eed76976a357d80372e264ef41721a0c4fb297fd5R102-R104

src/Transports.AspNetCore/WebSockets/IOperationMessageProcessor.cs
Shane32: Not at this time.  We could add the interface anyway, but the only thing that happens during disposal is signaling cancellation tokens and disconnecting active subscriptions (meaning, signaling the event source that no more notifications should be provided).

src/Transports.AspNetCore/WebSockets/IOperationMessageProcessor.cs
Shane32: Actually, I do also call `Dispose` proactively from the `IWebSocketConnection` implementation in order to prevent additional subscription notifications arriving after the WebSocket is negotiating a graceful shutdown.  It is still necessary in `WebSocketHandler` in case a graceful shutdown does not occur, of course.

https://github.com/graphql-dotnet/server/pull/809/files#diff-9bfc1e96fa77f9fb55180ef1c03bd235d0d755c0154382b67f8436daf4422acbR102-R103

Shane32:
```suggestion
/// <see cref="IDisposable.Dispose"/> must be called when the WebSocket connection
/// has received a close message, and/or when the connection terminates.
```

src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
Shane32: `CloseConnectionAsync` posts a WebSocket message through the message pump to the client that informs it that the WebSocket connection is shutting down.  Per HTTP WebSocket protocol, a graceful shutdown occurs only after both endpoints have sent/acknowledged a close message.  Per GraphQL over WebSocket protocols, the close event id needs to be certain values for certain events.

`CloseConnectionAsync` does not dispose of the `IWebSocketConnection` implementation, as it must stay alive until any remaining messages in its queue have sent and the counterparty has responded with a close message.  In fact, there is no `Dispose` method at all because all of the held objects (e.g. `WebSocket`) originate from `HttpContext` and are disposed by ASP.NET Core.

I did just now add `IDisposable` inheritance to `IWebSocketConnection` anyway, in case of code changes in the future or other implementations, although it does not contain any code at this time.

Within the implementation (#809 ), `ExecuteAsync` will return once the graceful shutdown has completed.  Once the middleware returns, ASP.NET Core will close the HTTP connection.

Shane32: > CloseAsync

I don't know.  Please re-review the updated comments and then confirm if you think we should go with `CloseAsync`.  This is a `IWebSocketConnection` after all, so having `Connection` in the name seems excessive, but I don't want to imply that disposal occurs (like `Close` for a `DbConnection`), which it does not.  On the other hand, it does (eventually) close the connection, so `CloseAsync` may be fine.

> extension method

Seems to me the normal closure message and/or behavior is dependent on the implementation.  I would not eliminate the method.

src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
Shane32: Please re-review xml comments and my comments above and confirm.

src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
Shane32: To implement the idle timeout, it seemed to me to make sense to only send a keep-alive message after ## seconds of "no traffic".  But sent messages are posted to a queue, and there is no way for `IOperationMessageProcessor` to know when the last message was actually sent.  Multiple subscriptions may be active on one channel, not to mention queries or mutations being executed, and some `ExecutionResult` messages may be large, or connection quality may be poor -- any of those conditions may delay sending of messages from when it was posted.  So this property allows for a more accurate idle timeout implementation.

src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
Shane32: Because shutting down the host typically waits for active HTTP connections to complete, up to 30 seconds.  GraphQL subscriptions would prevent prompt shutdown, typically preventing shutdown for the full 30 seconds, but by monitoring `IHostApplicationLifetime`, we can shut down WebSocket connections promptly when the host is shutting down.

Note that while it is possible to attempt a graceful shutdown of the WebSocket connection upon a host shutdown notification, I did not attempt it.  Rather, if the host is shutting down, all subscriptions are immediately disconnected, and the HTTP connection is immediately shut down.  This is accomplished by creating a linked cancellation token source and using it throughout the WebSocket pipeline rather than `HttpContext.RequestAborted`.

See also:
- https://github.com/ChilliCream/hotchocolate/pull/4775

Shane32: > Or is that `HttpContext` initial one where connection was created?

Yes... there is only a single `HttpContext` (and a single `WebSocket` instance) for the entire lifetime of a WebSocket connection over HTTP.

src/Transports.AspNetCore/WebSockets/IWebSocketAuthenticationService.cs
Shane32: The ConnectionInit message.  Again, the only purpose of the ConnectionInit message is to allow authentication/authorization for the connection, since browsers cannot send the HTTP 'Authorization' header to a WebSocket endpoint.

src/Transports.AspNetCore/WebSockets/IOperationMessageProcessor.cs
Shane32: Please review again, and provide a suggestion if you think further xml comment changes are necessary.

src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
Shane32: After thinking about it for the last hour or so, I'll go with `CloseAsync` here; but please lmk if you have any other comments.

src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
Shane32: See above

src/Transports.AspNetCore/WebSockets/IOperationMessageProcessor.cs
sungam3r: > Starts the timer that closes the connection if no ConnectionInit message is received after ConnectionInitWaitTimeout

I would write exactly that in xml comments.

>  Per the protocol, any other message is refused until this "ConnectionInit" message is received and acknowledged.

And may be that too.

src/Transports.AspNetCore/WebSockets/IOperationMessageProcessor.cs
sungam3r: Now I understand.

src/Transports.AspNetCore/WebSockets/IOperationMessageProcessor.cs
sungam3r: OK to keep as is.

src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
sungam3r: Thanks. And again I note that I would happy to see these explanations in the xml comments.

src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// </summary>
+    Task CloseAsync();
+
+    /// <summary>
+    /// Performs a graceful shutdown of the WebSocket connection with the specified error information.
+    /// </summary>
+    Task CloseAsync(int eventId, string? description);
+
+    /// <summary>
+    /// Returns the last UTC time that a message was sent.
```
sungam3r, line 39:
```suggestion
    /// Returns the last UTC time that a message was sent to the client. This property is needed to implement the idle timeout to only send a keep-alive message after ## seconds of "no traffic". 
```

src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
sungam3r: Thanks for explanations. So actually `RequestAborted` does not equal to `HttpContext.RequestAborted`. It may be a some linked token, right?. If so maybe to give some other name to this property to not interfere with `HttpContext.RequestAborted` ?

src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// </summary>
+    Task ExecuteAsync(IOperationMessageProcessor operationMessageProcessor);
+
+    /// <summary>
+    /// Sends a message to the client.
+    /// </summary>
+    Task SendMessageAsync(OperationMessage message);
+
+    /// <summary>
+    /// Performs a graceful shutdown of the WebSocket connection with event ID 1000.
```
sungam3r, line 29: > ID 1000

Required by protocol spec?

sungam3r (approved):
I approve this PR so not to slow you down. My remaining suggestions/questions relate to comments.

src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
Shane32: I’m open to suggestions. It’s not like the name is “ConnectionClosed” - the application shutdown is aborting the request so the terminology is still correct - but we could change it if you think we should.  (I’d leave it.)

src/Transports.AspNetCore/WebSockets/IWebSocketConnection.cs
Shane32: The WebSocket spec, yes. See https://docs.microsoft.com/en-us/dotnet/api/system.net.websockets.websocketclosestatus?view=net-6.0

I did not use the enum since the number is arbitrary and the GraphQL over WebSocket spec defines other event id numbers to use for specific situations which are not in the enum.

That may be a good use of an extension method - an “overload” with the status enum instead of the int.

Shane32:
> I approve this PR so not to slow you down. My remaining suggestions/questions relate to comments.

Then I'll merge it.  We can always issue a follow-up PR with additional changes.

PR #808, opened 2022-05-22
Add AsyncMessagePump class for WebSocket support

Shane32:
This facilitates a WebSocketConnection implementation keeping the outbound messages in order while remaining thread-safe.

PR #809, opened 2022-05-22
Add WebSocketConnection implementation

Shane32:
Implementation of `IWebSocketConnection` interface.

Depends on:
- [x] #802 
- [x] #804 
- [x] #805 
- [x] #807 
- [x] #808

### DIFF
--- a/src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
+++ b/src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
@@ -15,7 +15,7 @@ namespace GraphQL.Server.Transports.AspNetCore.WebSockets;
 /// Calls to <see cref="IOperationMessageProcessor.OnMessageReceivedAsync(OperationMessage)"/> be awaited
 /// before dispatching subsequent messages.
 /// <br/><br/>
-/// Calls to <see cref="CloseConnectionAsync()"/> and <see cref="SendMessageAsync(OperationMessage)"/> may be
+/// Calls to <see cref="CloseAsync()"/> and <see cref="SendMessageAsync(OperationMessage)"/> may be
 /// called on multiple threads simultaneously. They are queued for delivery and sent in the order posted.
 /// Messages posted after requesting the connection be closed will be discarded.
 /// </summary>
@@ -105,7 +105,7 @@ public class WebSocketConnection : IWebSocketConnection
                     if (!_outputClosed.Task.IsCompleted)
                     {
                         // queue the closure
-                        _ = CloseConnectionAsync();
+                        _ = CloseAsync();
                         // wait until the close has been sent
                         var completedTask = Task.WhenAny(
                             _outputClosed.Task,
@@ -166,11 +166,11 @@ public class WebSocketConnection : IWebSocketConnection
     }
 
     /// <inheritdoc/>
-    public Task CloseConnectionAsync()
-        => CloseConnectionAsync(1000, null);
+    public Task CloseAsync()
+        => CloseAsync(1000, null);
 
     /// <inheritdoc/>
-    public Task CloseConnectionAsync(int eventId, string? description)
+    public Task CloseAsync(int eventId, string? description)
     {
         _pump.Post(new Message { CloseStatus = (WebSocketCloseStatus)eventId, CloseDescription = description });
         return Task.CompletedTask;
@@ -188,8 +188,8 @@ public class WebSocketConnection : IWebSocketConnection
     /// or <see cref="WebSocketCloseStatus"/> with description, passing to either
     /// <see cref="OnSendMessageAsync(OperationMessage)"/> or <see cref="OnCloseOutputAsync(WebSocketCloseStatus, string?)"/>.
     /// <br/><br/>
-    /// The methods <see cref="SendMessageAsync(OperationMessage)"/>, <see cref="CloseConnectionAsync()"/>
-    /// and <see cref="CloseConnectionAsync(int, string?)"/> add <see cref="Message"/> instances to the queue.
+    /// The methods <see cref="SendMessageAsync(OperationMessage)"/>, <see cref="CloseAsync()"/>
+    /// and <see cref="CloseAsync(int, string?)"/> add <see cref="Message"/> instances to the queue.
     /// </summary>
     private async Task HandleMessageAsync(Message message)
     {
--- a/src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
+++ b/src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
@@ -26,7 +26,7 @@ public class WebSocketConnection : IWebSocketConnection
     private readonly IGraphQLSerializer _serializer;
     private readonly WebSocketWriterStream _stream;
     private readonly TaskCompletionSource<bool> _outputClosed = new();
-    private readonly int _closeTimeoutMs;
+    private readonly TimeSpan _closeTimeout;
     private int _executed;
 
     /// <inheritdoc/>
@@ -47,7 +47,7 @@ public class WebSocketConnection : IWebSocketConnection
     /// <summary>
     /// Initializes an instance with the specified parameters.
     /// </summary>
-    public WebSocketConnection(HttpContext httpContext, WebSocket webSocket, IGraphQLSerializer serializer, GraphQLHttpMiddlewareOptions options, CancellationToken cancellationToken)
+    public WebSocketConnection(HttpContext httpContext, WebSocket webSocket, IGraphQLSerializer serializer, GraphQLHttpMiddlewareOptions options, CancellationToken requestAborted)
     {
         HttpContext = httpContext ?? throw new ArgumentNullException(nameof(httpContext));
         if (options == null)
@@ -59,12 +59,12 @@ public class WebSocketConnection : IWebSocketConnection
                 throw new ArgumentOutOfRangeException(nameof(options) + "." + nameof(GraphQLWebSocketOptions.DisconnectionTimeout));
 #pragma warning restore CA2208 // Instantiate argument exceptions correctly
         }
-        _closeTimeoutMs = (int)(options.WebSockets.DisconnectionTimeout ?? DefaultDisconnectionTimeout).TotalMilliseconds;
+        _closeTimeout = options.WebSockets.DisconnectionTimeout ?? DefaultDisconnectionTimeout;
         _webSocket = webSocket ?? throw new ArgumentNullException(nameof(webSocket));
         _stream = new(webSocket);
         _serializer = serializer ?? throw new ArgumentNullException(nameof(serializer));
-        _pump = new(HandleMessageAsync);
-        RequestAborted = cancellationToken;
+        _pump = new(HandleOutgoingMessageAsync);
+        RequestAborted = requestAborted;
     }
 
     /// <summary>
@@ -101,17 +101,17 @@ public class WebSocketConnection : IWebSocketConnection
                 {
                     // prevent any more messages from being queued
                     operationMessageProcessor.Dispose();
-                    // send a close request if none was sent yet
-                    if (!_outputClosed.Task.IsCompleted)
+                    // send a close request if none was sent yet and the socket has not yet been closed
+                    if (!_outputClosed.Task.IsCompleted && !RequestAborted.IsCancellationRequested)
                     {
                         // queue the closure
                         _ = CloseAsync();
                         // wait until the close has been sent
-                        var completedTask = Task.WhenAny(
+                        await Task.WhenAny(
                             _outputClosed.Task,
-                            Task.Delay(_closeTimeoutMs, RequestAborted));
+                            Task.Delay(_closeTimeout, RequestAborted));
                     }
-                    // quit
+                    // quit after the close request was fulfilled
                     return;
                 }
                 // if this is the last block terminating a message
@@ -153,7 +153,7 @@ public class WebSocketConnection : IWebSocketConnection
                 }
             }
         }
-        catch (WebSocketException)
+        catch (WebSocketException) when (RequestAborted.IsCancellationRequested)
         {
         }
         finally
@@ -167,10 +167,16 @@ public class WebSocketConnection : IWebSocketConnection
     }
 
     /// <inheritdoc/>
+    /// <remarks>
+    /// The close message is posted to a queue and execution returns immediately.
+    /// </remarks>
     public Task CloseAsync()
         => CloseAsync(1000, null);
 
     /// <inheritdoc/>
+    /// <remarks>
+    /// The close message is posted to a queue and execution returns immediately.
+    /// </remarks>
     public Task CloseAsync(int eventId, string? description)
     {
         _pump.Post(new Message { CloseStatus = (WebSocketCloseStatus)eventId, CloseDescription = description });
@@ -178,6 +184,9 @@ public class WebSocketConnection : IWebSocketConnection
     }
 
     /// <inheritdoc/>
+    /// <remarks>
+    /// The message is posted to a queue and execution returns immediately.
+    /// </remarks>
     public Task SendMessageAsync(OperationMessage message)
     {
         _pump.Post(new Message { OperationMessage = message });
@@ -192,8 +201,9 @@ public class WebSocketConnection : IWebSocketConnection
     /// The methods <see cref="SendMessageAsync(OperationMessage)"/>, <see cref="CloseAsync()"/>
     /// and <see cref="CloseAsync(int, string?)"/> add <see cref="Message"/> instances to the queue.
     /// </summary>
-    private async Task HandleMessageAsync(Message message)
+    private async Task HandleOutgoingMessageAsync(Message message)
     {
+        // Messages posted after requesting the connection be closed will be discarded.
         if (_outputClosed.Task.IsCompleted)
             return;
         LastMessageSentAt = DateTime.UtcNow;
@@ -203,8 +213,8 @@ public class WebSocketConnection : IWebSocketConnection
         }
         else
         {
-            _outputClosed.TrySetResult(true);
             await OnCloseOutputAsync(message.CloseStatus, message.CloseDescription);
+            _outputClosed.TrySetResult(true);
         }
     }
 
@@ -239,9 +249,9 @@ public class WebSocketConnection : IWebSocketConnection
         => _webSocket.CloseOutputAsync(closeStatus, closeDescription, RequestAborted);
 
     /// <summary>
-    /// A queue entry; see <see cref="HandleMessageAsync(Message)"/>.
+    /// A queue entry; see <see cref="HandleOutgoingMessageAsync(Message)"/>.
     /// </summary>
-    /// <param name="OperationMessage">The message to send, if set; if it is null then this is a closure message.</param>
+    /// <param name="OperationMessage">The message to send, if set; if it is <see langword="null"/> then this is a closure message.</param>
     /// <param name="CloseStatus">The close status.</param>
     /// <param name="CloseDescription">The close description.</param>
     private record struct Message(OperationMessage? OperationMessage, WebSocketCloseStatus CloseStatus, string? CloseDescription);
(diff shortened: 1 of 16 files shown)


sungam3r:
I would like to review this one after #807 and #808 being merged.

Shane32:
@sungam3r This PR is ready for review.  I did look at the RecyclableMemoryStream and it seems that it would be useful here.  However, I suggest we review it as a separate PR / optimization after this is merged, as all the code here is already done and tested.

Another optimization I had thought of would be to have the reads from the serializer get passed through directly to the WebSocket class -- so the JSON result can be deserialized 'live' as the data is read from the socket, and it eliminates having to copy the data from the stream buffer to the serializer buffer.  That would not need a 3rd party dependency, but it requires a bunch of code.  Of course only System.Text.Json is async, whereas Newtonsoft is sync, and the WebSocket class only supports async reads.

And finally, the code that exists now is fairly optimized for requests < 16kb.  I would expect that most requests will fall into this category; only requests with very large variables will trigger an allocation.  So maybe we should leave it as-is.

sungam3r:
>  So maybe we should leave it as-is.

I'm 80% sure to leave as is (at least for some time) without making changes to read directly from WebSocket.

sungam3r:
I'll try to review this one today. Every morning (more) and evening (less) I spend a few hours to review, support of other projects and reading email. The rest of the time I am engaged in the building/preparation of the house, so my resources are limited even though I'm on vacation.

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
```diff
@@ -0,0 +1,253 @@
...
+
+/// <summary>
+/// Manages a WebSocket connection, dispatching messages to the specified <see cref="IOperationMessageProcessor"/>,
+/// and sending messages requested by the <see cref="IOperationMessageProcessor"/> implementation.
+/// <br/><br/>
+/// The <see cref="ExecuteAsync(IOperationMessageProcessor)"/> method may only be executed once for each
+/// instance. Awaiting the result will return once the WebSocket connection has been properly closed from both
+/// ends, after all messages have been sent.
+/// <br/><br/>
+/// Calls to <see cref="IOperationMessageProcessor.OnMessageReceivedAsync(OperationMessage)"/> be awaited
```
sungam3r, line 15:
```suggestion
/// Calls to <see cref="IOperationMessageProcessor.OnMessageReceivedAsync(OperationMessage)"/> will be awaited
```

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
```diff
@@ -0,0 +1,253 @@
...
+/// before dispatching subsequent messages.
+/// <br/><br/>
+/// Calls to <see cref="CloseAsync()"/> and <see cref="SendMessageAsync(OperationMessage)"/> may be
+/// called on multiple threads simultaneously. They are queued for delivery and sent in the order posted.
+/// Messages posted after requesting the connection be closed will be discarded.
```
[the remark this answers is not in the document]
sungam3r: Ah, I see condition in `HandleMessageAsync`.

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
```diff
@@ -0,0 +1,253 @@
...
+/// </summary>
+public class WebSocketConnection : IWebSocketConnection
+{
+    private readonly WebSocket _webSocket;
+    private readonly AsyncMessagePump<Message> _pump;
+    private readonly IGraphQLSerializer _serializer;
+    private readonly WebSocketWriterStream _stream;
+    private readonly TaskCompletionSource<bool> _outputClosed = new();
+    private readonly int _closeTimeoutMs;
```
sungam3r, line 29:
```suggestion
    private readonly TimeSpan _closeTimeout;
```

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
```diff
@@ -0,0 +1,253 @@
...
+        if (options == null)
+            throw new ArgumentNullException(nameof(options));
+        if (options.WebSockets.DisconnectionTimeout.HasValue)
+        {
+            if ((options.WebSockets.DisconnectionTimeout.Value != Timeout.InfiniteTimeSpan && options.WebSockets.DisconnectionTimeout.Value.TotalMilliseconds < 0) || options.WebSockets.DisconnectionTimeout.Value.TotalMilliseconds > int.MaxValue)
+#pragma warning disable CA2208 // Instantiate argument exceptions correctly
+                throw new ArgumentOutOfRangeException(nameof(options) + "." + nameof(GraphQLWebSocketOptions.DisconnectionTimeout));
+#pragma warning restore CA2208 // Instantiate argument exceptions correctly
+        }
+        _closeTimeoutMs = (int)(options.WebSockets.DisconnectionTimeout ?? DefaultDisconnectionTimeout).TotalMilliseconds;
```
sungam3r, line 62:
```suggestion
        _closeTimeout = options.WebSockets.DisconnectionTimeout ?? DefaultDisconnectionTimeout;
```

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
```diff
@@ -0,0 +1,253 @@
...
+    /// <inheritdoc/>
+    public DateTime LastMessageSentAt { get; private set; } = DateTime.UtcNow;
+
+    /// <inheritdoc/>
+    public HttpContext HttpContext { get; }
+
+    /// <summary>
+    /// Initializes an instance with the specified parameters.
+    /// </summary>
+    public WebSocketConnection(HttpContext httpContext, WebSocket webSocket, IGraphQLSerializer serializer, GraphQLHttpMiddlewareOptions options, CancellationToken cancellationToken)
```
sungam3r, line 50: I would append a note that `cancellationToken` is for `RequestAborted` property or rename it into `requestAborted`.

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
```diff
@@ -0,0 +1,253 @@
...
+    /// Listens to incoming messages on the WebSocket specified in the constructor,
+    /// dispatching the messages to the specified <paramref name="operationMessageProcessor"/>.
+    /// Returns or throws <see cref="OperationCanceledException"/> when the WebSocket connection is closed.
+    /// </summary>
+    public virtual async Task ExecuteAsync(IOperationMessageProcessor operationMessageProcessor)
+    {
+        if (operationMessageProcessor == null)
+            throw new ArgumentNullException(nameof(operationMessageProcessor));
+        if (Interlocked.Exchange(ref _executed, 1) == 1)
+            throw new InvalidOperationException($"{nameof(ExecuteAsync)} may only be called once per instance.");
```
sungam3r, line 80: 👍

sungam3r, line 80:
```suggestion
            throw new InvalidOperationException($"{nameof(ExecuteAsync)} may only be called only once per instance.");
```

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
```diff
@@ -0,0 +1,253 @@
...
+            operationMessageProcessor.Dispose();
+        }
+    }
+
+    /// <inheritdoc/>
+    public Task CloseAsync()
+        => CloseAsync(1000, null);
+
+    /// <inheritdoc/>
+    public Task CloseAsync(int eventId, string? description)
```
sungam3r, line 174: Worth to add a comment to `IWebSocketConnection` that close methods do not really close connection immediately.

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
```diff
@@ -0,0 +1,253 @@
...
+                {
+                    // prevent any more messages from being queued
+                    operationMessageProcessor.Dispose();
+                    // send a close request if none was sent yet
+                    if (!_outputClosed.Task.IsCompleted)
+                    {
+                        // queue the closure
+                        _ = CloseAsync();
+                        // wait until the close has been sent
+                        var completedTask = Task.WhenAny(
```
sungam3r, line 110: You do not use `completedTask`. Let's throw `TimeoutException` here.

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
```diff
@@ -0,0 +1,253 @@
...
+                            _outputClosed.Task,
+                            Task.Delay(_closeTimeoutMs, RequestAborted));
+                    }
+                    // quit
```
sungam3r, line 114:
```suggestion
                    // quit after close request was fulfilled
```

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
```diff
@@ -0,0 +1,253 @@
...
+                else
+                {
+                    // if there is any data in this block, add it to the buffer
+                    if (result.Count > 0)
+                        receiveStream.Write(buffer, 0, result.Count);
+                }
+            }
+        }
+        catch (WebSocketException)
+        {
```
sungam3r, line 157: Why empty cach?

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
```diff
@@ -0,0 +1,253 @@
...
+        else
+        {
+            _outputClosed.TrySetResult(true);
+            await OnCloseOutputAsync(message.CloseStatus, message.CloseDescription);
+        }
+    }
+
+    /// <summary>
+    /// Dispatches a received message to an <see cref="IOperationMessageProcessor"/> instance.
+    /// Override if logging is desired.
```
sungam3r, line 213: 👍

sungam3r:
Reviewed.

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
```diff
@@ -0,0 +1,253 @@
...
+    /// Initializes an instance with the specified parameters.
+    /// </summary>
+    public WebSocketConnection(HttpContext httpContext, WebSocket webSocket, IGraphQLSerializer serializer, GraphQLHttpMiddlewareOptions options, CancellationToken cancellationToken)
+    {
+        HttpContext = httpContext ?? throw new ArgumentNullException(nameof(httpContext));
+        if (options == null)
+            throw new ArgumentNullException(nameof(options));
+        if (options.WebSockets.DisconnectionTimeout.HasValue)
+        {
+            if ((options.WebSockets.DisconnectionTimeout.Value != Timeout.InfiniteTimeSpan && options.WebSockets.DisconnectionTimeout.Value.TotalMilliseconds < 0) || options.WebSockets.DisconnectionTimeout.Value.TotalMilliseconds > int.MaxValue)
```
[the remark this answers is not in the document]
Shane32: I’m not sure what your comment is. TimeSpans can contain negative values, which is illegal here, except for -1 which means no timeout. (Infinite timeout)

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
Shane32: It often happens during normal closure and I did not find another way to prevent it.

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
Shane32: It certainly should await the result.  But why return a timeout exception?  On this end, the closure has taken place normally.  The only reason the delay exists is to facilitate a graceful shutdown, but a non-graceful shutdown does not indicate a problem on this end.

Shane32: I added tests in my project for this scenario, which will be copied over when we copy the rest of the tests.  (Ugh; it took like an hour or more...)

Shane32: https://github.com/Shane32/GraphQL.AspNetCore3/pull/35

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
Shane32: I'm unable to reproduce the problem now.  But practically, the only situation when a WebSocketException should occur is when the client disconnects in an incorrect manner or sends an invalid WebSocket packet, in which case it's not a server problem (and hence should not be logged).  Still, let's provide a protected method so that the exception could be handled if desired.

Shane32: Actually, inherited classes can wrap ExecuteAsync if they want to catch that exception.  They would also need to wrap HandleMessageAsync so as to catch exceptions while sending.

Shane32: I'm just changing this line to

```cs
catch (WebSocketException) when (RequestAborted.IsCancellationRequested)
```

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
```diff
@@ -0,0 +1,253 @@
...
+    {
+        if (operationMessageProcessor == null)
+            throw new ArgumentNullException(nameof(operationMessageProcessor));
+        if (Interlocked.Exchange(ref _executed, 1) == 1)
+            throw new InvalidOperationException($"{nameof(ExecuteAsync)} may only be called once per instance.");
+        try
+        {
+            await operationMessageProcessor.InitializeConnectionAsync();
+            // set up a buffer in case a message is longer than one block
+            var receiveStream = new MemoryStream();
```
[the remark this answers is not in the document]
Shane32: Do you have any questions here?  Yes it is complicated to read.  Yes it would certainly be more optimized with the MS package you previously referenced.  But everything should work.

Note that I do not know how likely it is for a message to come through as a single packet (and so retain in the 16kb buffer).  It is possible that messages often come through as multiple packets, especially if TCP breaks apart packets over 1500 bytes.  This could be further optimized by reusing the 16kb buffer until it was full before switching to the stream.  

It's also possible that the message comes through as a packet and then the end-of-message packet comes through separately (just as we send messages).  This would currently hit the stream instead of staying within the 16kb buffer (without further optimization).

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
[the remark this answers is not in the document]
sungam3r: > TimeSpans can contain negative values

WOW, did not notice before. OK.

> Also of note, -0.5 milliseconds is invalid, which this test checks for.

How?
```
 TimeSpan f = new TimeSpan(5000);
 var t = f.TotalMilliseconds;  <--- 0.5 (double), 0.5 < 0 is false
 ```

File: src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs (hunk 10 above)
Comment (sungam3r) [reply src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs:157]: Yes, they can wrap but if we will keep exception swallowing here then no exception bubbles up.

File: src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs (hunk 13 above)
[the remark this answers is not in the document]
Comment (sungam3r) [reply src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs:85]: No suggestions here. I'm fine with current implementation.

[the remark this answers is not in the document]
Comment (sungam3r) [reply src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs:85]: #813

File: src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs (hunk 10 above)
Comment (Shane32) [reply src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs:157]: Ok…So now it’s only swallowed when it’s logical to do so - ok??

File: src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs (hunk 12 above)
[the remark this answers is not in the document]
Comment (Shane32) [reply src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs:57]: Just saying rather than check “if less than -1 throw”, which would not catch the -0.5 error condition, the code checks “if less than zero except -1” which is more accurate. And since Task.Delay converts to int milliseconds, anything over int.MaxValue would be an overflow, so that’s also invalid. My code checks for all those values.

File: src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs (hunk 8 above)
Comment (Shane32) [reply src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs]: I don’t think we should add OnNonGracefulShutdown at all. I have seen implementations that never perform a graceful shutdown. But if it is added, then we should determine more precisely what situations it is triggered in. 

1. Client disconnects with no warning
2. Client disconnects after server requests closure, but without client sending close message
3. Client sends close request and immediately disconnects without waiting for server close message 
4. Server sends close message, and there is a timeout waiting for client disconnect message to arrive

`OnNonGracefulShutdown(bool receivedCloseMessage, bool sentCloseMessage)`?

File: src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs (hunk 10 above)
Comment (Shane32) [reply src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs:157]: My thought is that middleware and code in general should handle expected errors and only unexpected errors should fall through. In this manner user exception handling will catch unexpected errors. 

For example, when I write any TCP listener code, I always catch socket exceptions, because it is normal for a socket exception to occur when there is a disconnection. And that’s not a abnormal error. It’s typical. 

I’m not sure how the WebSocket code is written, but it certainly should not be throwing exceptions when the socket is disconnected, and probably no other WebSocketExceptions either.

File: src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs (hunk 12 above)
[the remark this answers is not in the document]
Comment (sungam3r) [reply src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs:57]: Got it.

File: src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs (hunk 10 above)
Comment (sungam3r) [reply src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs:157]: > So now it’s only swallowed when it’s logical to do so

Seems so. Most likely such an implementation is good.

File: src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs (hunk 8 above)
Comment (Shane32) [reply src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs]: Please confirm:

```cs
public virtual Task OnNonGracefulShutdownAsync(bool receivedCloseMessage, bool sentCloseMessage)
    => Task.CompletedTask;
```

sungam3r: OK if this handles all cases. I did not read WebSocket protocol spec yet.

Shane32: Done.  Note that this relates to WebSockets in general, and is not specific to GraphQL over WebSockets.

sungam3r: I understand it. Thanks.

Shane32:
> I see you added _typically_, OK, but the wording is not clear for me. You said _This should be left up to the implementation_. Implementation of what? Options? I think you mean an implementation of other object that uses and **interprets** values from options (`IOperationMessageProcessor`). For some time I could not understand what confuses me in all these comments about default values and now I understood. These comments are about the behavior of **other** objects in relation to the `null`-values from this object. Now reading comments, this is not obvious for person not familiar with internals.

We could rewrite it again, something like this:

> A value of null indicates to use the default value defined by the implementation.  The included implementations in this library have a default value of 10 seconds.

I'm open to suggestions.

Shane32:
If the xml comments on the options class are the only outstanding issue, let's open a new issue/PR to clarify them and merge this one.

sungam3r:
> A value of null indicates to use the default value defined by the implementation. The included implementations in this library have a default value of 10 seconds.

Good.

sungam3r:
Let's make this last fix for docs and merge.

Shane32:
> Let's make this last fix for docs and merge.

Ok take a look now... ??

sungam3r (approved):

sungam3r:
Moving to #810?

PR #810, opened 2022-05-22
Add subscription servers for GraphQL and SubscriptionsTransportWs protocols

Shane32:
Provides an implementation (logic) for the GraphQL and SubscriptionsTransportWs protocols.

See:
- https://github.com/apollographql/subscriptions-transport-ws
- https://github.com/enisdenjo/graphql-ws

Depends on:
- [x] #802 
- [x] #806 
- [x] #807

### DIFF
--- a/src/Transports.AspNetCore/AuthorizationHelper.cs
+++ b/src/Transports.AspNetCore/AuthorizationHelper.cs
@@ -27,7 +27,7 @@ public static class AuthorizationHelper
             }
         }
 
-        if (options.AuthorizedRoles?.Count() > 0)
+        if (options.AuthorizedRoles?.Any() == true)
         {
             var user = options.HttpContext.User ?? NoUser();
             foreach (var role in options.AuthorizedRoles)
--- a/src/Transports.AspNetCore/AuthorizationHelper.cs
+++ b/src/Transports.AspNetCore/AuthorizationHelper.cs
@@ -27,7 +27,7 @@ public static class AuthorizationHelper
             }
         }
 
-        if (options.AuthorizedRoles?.Count > 0)
+        if (options.AuthorizedRoles?.Count() > 0)
         {
             var user = options.HttpContext.User ?? NoUser();
             foreach (var role in options.AuthorizedRoles)
--- a/src/Transports.AspNetCore/IAuthorizationOptions.cs
+++ b/src/Transports.AspNetCore/IAuthorizationOptions.cs
@@ -15,7 +15,7 @@ public interface IAuthorizationOptions
     /// for the user within <see cref="HttpContext.User"/>
     /// prior to executing the GraphQL request or accepting the WebSocket connection.
     /// Technically this property should be named as AuthenticationRequired but for
-    /// ASP.NET Core / GraphQL.NET naming and design decisions it was called so. 
+    /// ASP.NET Core / GraphQL.NET naming and design decisions it was called so.
     /// </summary>
     bool AuthorizationRequired { get; }
 
--- a/src/Transports.AspNetCore/IAuthorizationOptions.cs
+++ b/src/Transports.AspNetCore/IAuthorizationOptions.cs
@@ -1,0 +1,37 @@
+using System.Security.Claims;
+using System.Security.Principal;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Http;
+
+namespace GraphQL.Server.Transports.AspNetCore;
+
+/// <summary>
+/// Provides authorization options.
+/// </summary>
+public interface IAuthorizationOptions
+{
+    /// <summary>
+    /// If set, requires that <see cref="IIdentity.IsAuthenticated"/> return <see langword="true"/>
+    /// for the user within <see cref="HttpContext.User"/>
+    /// prior to executing the GraphQL request or accepting the WebSocket connection.
+    /// Technically this property should be named as AuthenticationRequired but for
+    /// ASP.NET Core / GraphQL.NET naming and design decisions it was called so. 
+    /// </summary>
+    bool AuthorizationRequired { get; }
+
+    /// <summary>
+    /// Requires that <see cref="ClaimsPrincipal.IsInRole(string)"/> return <see langword="true"/>
+    /// for the user within <see cref="HttpContext.User"/>
+    /// for at least one role in the list prior to executing the GraphQL request or accepting
+    /// the WebSocket connection.  If no roles are specified, authorization is not checked.
+    /// </summary>
+    IEnumerable<string> AuthorizedRoles { get; }
+
+    /// <summary>
+    /// If set, requires that <see cref="IAuthorizationService.AuthorizeAsync(ClaimsPrincipal, object, string)"/>
+    /// return a successful result for the user within <see cref="HttpContext.User"/>
+    /// for the specified policy before executing the GraphQL
+    /// request or accepting the WebSocket connection.
+    /// </summary>
+    string? AuthorizedPolicy { get; }
+}
--- a/src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.Observer.cs
+++ b/src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.Observer.cs
@@ -7,15 +7,15 @@ public abstract partial class BaseSubscriptionServer
     /// </summary>
     private class Observer : IObserver<ExecutionResult>
     {
-        private readonly BaseSubscriptionServer _handler;
+        private readonly BaseSubscriptionServer _server;
         private readonly string _id;
         private readonly bool _closeAfterOnError;
         private readonly bool _closeAfterAnyError;
         private int _done;
 
-        public Observer(BaseSubscriptionServer handler, string id, bool closeAfterOnError, bool closeAfterAnyError)
+        public Observer(BaseSubscriptionServer server, string id, bool closeAfterOnError, bool closeAfterAnyError)
         {
-            _handler = handler;
+            _server = server;
             _id = id;
             _closeAfterOnError = closeAfterOnError;
             _closeAfterAnyError = closeAfterAnyError;
@@ -27,35 +27,29 @@ public abstract partial class BaseSubscriptionServer
                 return;
             try
             {
-                _ = _handler.SendCompletedAsync(_id);
+                _ = _server.SendCompletedAsync(_id);
             }
             catch { }
         }
 
         public async void OnError(Exception error)
         {
-            if (_closeAfterOnError)
-            {
-                if (Interlocked.Exchange(ref _done, 1) == 1)
-                    return;
-            }
-            else
-            {
-                if (Interlocked.CompareExchange(ref _done, 0, 0) == 1)
-                    return;
-            }
+            if (Thread.VolatileRead(ref _done) == 1)
+                return;
+            if (_closeAfterOnError && Interlocked.Exchange(ref _done, 1) == 1)
+                return;
             try
             {
                 if (error != null)
                 {
-                    var executionError = error is ExecutionError ee ? ee : await _handler.HandleErrorFromSourceAsync(error);
+                    var executionError = error is ExecutionError ee ? ee : await _server.HandleErrorFromSourceAsync(error);
                     if (executionError != null)
                     {
                         var result = new ExecutionResult
                         {
                             Errors = new ExecutionErrors { executionError },
                         };
-                        await _handler.SendDataAsync(_id, result);
+                        await _server.SendDataAsync(_id, result);
                     }
                 }
             }
@@ -63,23 +57,21 @@ public abstract partial class BaseSubscriptionServer
             try
             {
                 if (_closeAfterOnError)
-                    await _handler.SendCompletedAsync(_id);
+                    await _server.SendCompletedAsync(_id);
             }
             catch { }
         }
 
         public async void OnNext(ExecutionResult value)
         {
-            if (Interlocked.CompareExchange(ref _done, 0, 0) == 1)
-                return;
-            if (value == null)
+            if (value == null || Thread.VolatileRead(ref _done) == 1)
                 return;
             try
             {
-                await _handler.SendDataAsync(_id, value);
-                if (_closeAfterAnyError && value.Errors != null && value.Errors.Count > 0)
+                await _server.SendDataAsync(_id, value);
+                if (_closeAfterAnyError && value.Errors?.Count > 0)
                 {
-                    await _handler.SendCompletedAsync(_id);
+                    await _server.SendCompletedAsync(_id);
                 }
             }
             catch { }
--- a/src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.Observer.cs
+++ b/src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.Observer.cs
@@ -1,0 +1,88 @@
+namespace GraphQL.Server.Transports.AspNetCore.WebSockets;
+
+public abstract partial class BaseSubscriptionServer
+{
+    /// <summary>
+    /// Handles messages from the event source.
+    /// </summary>
+    private class Observer : IObserver<ExecutionResult>
+    {
+        private readonly BaseSubscriptionServer _handler;
+        private readonly string _id;
+        private readonly bool _closeAfterOnError;
+        private readonly bool _closeAfterAnyError;
+        private int _done;
+
+        public Observer(BaseSubscriptionServer handler, string id, bool closeAfterOnError, bool closeAfterAnyError)
+        {
+            _handler = handler;
+            _id = id;
+            _closeAfterOnError = closeAfterOnError;
+            _closeAfterAnyError = closeAfterAnyError;
+        }
+
+        public void OnCompleted()
+        {
+            if (Interlocked.Exchange(ref _done, 1) == 1)
+                return;
+            try
+            {
+                _ = _handler.SendCompletedAsync(_id);
+            }
+            catch { }
+        }
+
+        public async void OnError(Exception error)
+        {
+            if (_closeAfterOnError)
+            {
+                if (Interlocked.Exchange(ref _done, 1) == 1)
+                    return;
+            }
+            else
+            {
+                if (Interlocked.CompareExchange(ref _done, 0, 0) == 1)
+                    return;
+            }
+            try
+            {
+                if (error != null)
+                {
+                    var executionError = error is ExecutionError ee ? ee : await _handler.HandleErrorFromSourceAsync(error);
+                    if (executionError != null)
+                    {
+                        var result = new ExecutionResult
+                        {
+                            Errors = new ExecutionErrors { executionError },
+                        };
+                        await _handler.SendDataAsync(_id, result);
+                    }
+                }
+            }
+            catch { }
+            try
+            {
+                if (_closeAfterOnError)
+                    await _handler.SendCompletedAsync(_id);
+            }
+            catch { }
+        }
+
+        public async void OnNext(ExecutionResult value)
+        {
+            if (Interlocked.CompareExchange(ref _done, 0, 0) == 1)
+                return;
+            if (value == null)
+                return;
+            try
+            {
+                await _handler.SendDataAsync(_id, value);
+                if (_closeAfterAnyError && value.Errors != null && value.Errors.Count > 0)
+                {
+                    await _handler.SendCompletedAsync(_id);
+                }
+            }
+            catch { }
+        }
+    }
+}
--- a/src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.Observer.cs
+++ b/src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.Observer.cs
@@ -40,6 +40,8 @@ public abstract partial class BaseSubscriptionServer
                 return;
             try
             {
+                // although error should never be null, if the event source does call OnError(null!),
+                // skip sending an error packet/message (allowed by spec)
                 if (error != null)
                 {
                     var executionError = error is ExecutionError ee ? ee : await _server.HandleErrorFromSourceAsync(error);
--- a/src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
+++ b/src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
@@ -45,12 +45,12 @@ public abstract partial class BaseSubscriptionServer : IOperationMessageProcesso
     protected virtual TimeSpan DefaultConnectionTimeout { get; } = TimeSpan.FromSeconds(10);
 
     /// <summary>
-    /// Initailizes a new instance with the specified parameters.
+    /// Initializes a new instance with the specified parameters.
     /// </summary>
-    /// <param name="sendStream">The WebSockets stream used to send data packets or close the connection.</param>
+    /// <param name="connection">The WebSockets stream used to send data packets or close the connection.</param>
     /// <param name="options">Configuration options for this instance.</param>
-    public BaseSubscriptionServer(
-        IWebSocketConnection sendStream,
+    protected BaseSubscriptionServer(
+        IWebSocketConnection connection,
         GraphQLHttpMiddlewareOptions options)
     {
         _options = options ?? throw new ArgumentNullException(nameof(options));
@@ -68,8 +68,8 @@ public abstract partial class BaseSubscriptionServer : IOperationMessageProcesso
                 throw new ArgumentOutOfRangeException($"{nameof(options)}.{nameof(GraphQLHttpMiddlewareOptions.WebSockets)}.{nameof(GraphQLWebSocketOptions.KeepAliveTimeout)}");
 #pragma warning restore CA2208 // Instantiate argument exceptions correctly
         }
-        Connection = sendStream ?? throw new ArgumentNullException(nameof(sendStream));
-        _cancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(sendStream.RequestAborted);
+        Connection = connection ?? throw new ArgumentNullException(nameof(connection));
+        _cancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(connection.RequestAborted);
         CancellationToken = _cancellationTokenSource.Token;
     }
 
@@ -114,7 +114,7 @@ public abstract partial class BaseSubscriptionServer : IOperationMessageProcesso
     }
 
     /// <summary>
-    /// Indicates if the connection has been initialized yet.
+    /// Indicates if the connection has been already initialized.
     /// </summary>
     protected bool Initialized
         => _initialized == 1;
@@ -248,14 +248,14 @@ public abstract partial class BaseSubscriptionServer : IOperationMessageProcesso
         => ErrorAccessDeniedAsync();
 
     /// <summary>
-    /// Executes when the client is attempting to initalize the connection.
+    /// Executes when the client is attempting to initialize the connection.
     /// <br/><br/>
     /// By default, this first checks <see cref="AuthorizeAsync(OperationMessage)"/> to validate that the
     /// request has passed authentication.  If validation fails, the connection is closed with an Access
     /// Denied message.
     /// <br/><br/>
     /// Otherwise, the connection is acknowledged via <see cref="OnConnectionAcknowledgeAsync(OperationMessage)"/>,
-    /// <see cref="TryInitialize"/> is called to indicate that this WebSocket connection is ready to accept requests, 
+    /// <see cref="TryInitialize"/> is called to indicate that this WebSocket connection is ready to accept requests,
     /// and keep-alive messages are sent via <see cref="OnSendKeepAliveAsync"/> if configured to do so.
     /// Keep-alive messages are only sent if no messages have been sent over the WebSockets connection for the
     /// length of time configured in <see cref="GraphQLWebSocketOptions.KeepAliveTimeout"/>.
@@ -280,7 +280,7 @@ public abstract partial class BaseSubscriptionServer : IOperationMessageProcesso
 
         async Task StartKeepAliveLoopAsync()
         {
-            while (true)
+            while (!CancellationToken.IsCancellationRequested)
             {
                 await Task.Delay(keepAliveTimeout, CancellationToken);
                 await OnSendKeepAliveAsync();
@@ -291,18 +291,18 @@ public abstract partial class BaseSubscriptionServer : IOperationMessageProcesso
          * This 'smart' keep-alive logic doesn't work with subscription-transport-ws because the client will
          * terminate a connection 20 seconds after the last keep-alive was received (or it will never
          * terminate if no keep-alive was ever received).
-         * 
+         *
          * See: https://github.com/graphql/graphql-playground/issues/1247
          */
         async Task StartSmartKeepAliveLoopAsync()
         {
             var lastKeepAliveSent = Connection.LastMessageSentAt;
-            while (true)
+            while (!CancellationToken.IsCancellationRequested)
             {
                 var lastSent = Connection.LastMessageSentAt;
-                var lastComm = lastKeepAliveSent > lastSent ? lastKeepAliveSent : lastSent;
+                var lastCommunication = lastKeepAliveSent > lastSent ? lastKeepAliveSent : lastSent;
                 var now = DateTime.UtcNow;
-                var timePassed = now.Subtract(lastComm);
+                var timePassed = now.Subtract(lastCommunication);
                 if (timePassed >= keepAliveTimeout)
                 {
                     await OnSendKeepAliveAsync();
@@ -330,7 +330,7 @@ public abstract partial class BaseSubscriptionServer : IOperationMessageProcesso
     protected abstract Task OnConnectionAcknowledgeAsync(OperationMessage message);
 
     /// <summary>
-    /// Executes when a new subscription request has occurred.
+    /// Executes when a new GraphQL request (typically subscription) has occurred.
     /// Optionally disconnects any existing subscription associated with the same id.
     /// </summary>
     protected virtual async Task SubscribeAsync(OperationMessage message, bool overwrite)
@@ -416,7 +416,7 @@ public abstract partial class BaseSubscriptionServer : IOperationMessageProcesso
         => Task.FromResult<ExecutionError>(new UnhandledError("Unable to set up subscription for the requested field.", ex));
 
     /// <summary>
-    /// Sends a single result to the client for a subscription or request, along with a notice
+    /// Sends a single result to the client for a query or mutation request, along with a notice
     /// that it was the last result in the event stream.
     /// </summary>
     protected virtual async Task SendSingleResultAsync(OperationMessage message, ExecutionResult result)
@@ -426,35 +426,29 @@ public abstract partial class BaseSubscriptionServer : IOperationMessageProcesso
     }
 
     /// <summary>
-    /// Sends an execution error to the client during set-up of a subscription.
+    /// Sends an execution error to the client during set-up of a GraphQL request (typically subscription).
     /// </summary>
     protected virtual Task SendErrorResultAsync(OperationMessage message, ExecutionError executionError)
         => SendErrorResultAsync(message.Id!, executionError);
 
-    /// <summary>
-    /// Sends an execution error to the client during set-up of a subscription.
-    /// </summary>
+    /// <inheritdoc cref="SendErrorResultAsync(OperationMessage, ExecutionError)"/>
     protected virtual Task SendErrorResultAsync(string id, ExecutionError executionError)
         => SendErrorResultAsync(id, new ExecutionResult { Errors = new ExecutionErrors { executionError } });
 
-    /// <summary>
-    /// Sends an error result to the client during set-up of a subscription.
-    /// </summary>
+    /// <inheritdoc cref="SendErrorResultAsync(OperationMessage, ExecutionError)"/>
     protected virtual Task SendErrorResultAsync(OperationMessage message, ExecutionResult result)
         => SendErrorResultAsync(message.Id!, result);
 
-    /// <summary>
-    /// Sends an error result to the client during set-up of a subscription.
-    /// </summary>
+    /// <inheritdoc cref="SendErrorResultAsync(OperationMessage, ExecutionError)"/>
     protected abstract Task SendErrorResultAsync(string id, ExecutionResult result);
 
     /// <summary>
-    /// Sends a data packet to the client for a subscription event.
+    /// Sends a data packet to the client for a GraphQL request (typically a subscription event).
     /// </summary>
     protected abstract Task SendDataAsync(string id, ExecutionResult result);
 
     /// <summary>
-    /// Sends a notice that a subscription has completed and no more data packets will be sent.
+    /// Sends a notice that a GraphQL request has completed (typically a subscription) and no more data packets will be sent.
     /// </summary>
     protected abstract Task SendCompletedAsync(string id);
 
--- a/src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
+++ b/src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
@@ -438,19 +438,28 @@ public abstract partial class BaseSubscriptionServer : IOperationMessageProcesso
 
     /// <summary>
     /// Sends an execution error to the client during set-up of a GraphQL request (typically subscription).
+    /// The default implementation calls <see cref="SendErrorResultAsync(string, ExecutionError)"/>.
     /// </summary>
     protected virtual Task SendErrorResultAsync(OperationMessage message, ExecutionError executionError)
         => SendErrorResultAsync(message.Id!, executionError);
 
-    /// <inheritdoc cref="SendErrorResultAsync(OperationMessage, ExecutionError)"/>
+    /// <summary>
+    /// Sends an execution error to the client during set-up of a GraphQL request (typically subscription).
+    /// The default implementation calls <see cref="SendErrorResultAsync(string, ExecutionResult)"/>.
+    /// </summary>
     protected virtual Task SendErrorResultAsync(string id, ExecutionError executionError)
         => SendErrorResultAsync(id, new ExecutionResult { Errors = new ExecutionErrors { executionError } });
 
-    /// <inheritdoc cref="SendErrorResultAsync(OperationMessage, ExecutionError)"/>
+    /// <summary>
+    /// Sends an execution error to the client during set-up of a GraphQL request (typically subscription).
+    /// The default implementation calls <see cref="SendErrorResultAsync(string, ExecutionResult)"/>.
+    /// </summary>
     protected virtual Task SendErrorResultAsync(OperationMessage message, ExecutionResult result)
         => SendErrorResultAsync(message.Id!, result);
 
-    /// <inheritdoc cref="SendErrorResultAsync(OperationMessage, ExecutionError)"/>
+    /// <summary>
+    /// Sends an execution error to the client during set-up of a GraphQL request (typically subscription).
+    /// </summary>
     protected abstract Task SendErrorResultAsync(string id, ExecutionResult result);
 
     /// <summary>
--- a/src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
+++ b/src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
@@ -20,7 +20,7 @@ public abstract partial class BaseSubscriptionServer : IOperationMessageProcesso
     /// Returns a <see cref="IWebSocketConnection"/> instance that can be used
     /// to send messages to the client.
     /// </summary>
-    protected IWebSocketConnection Client { get; }
+    protected IWebSocketConnection Connection { get; }
 
     /// <summary>
     /// Returns a <see cref="System.Threading.CancellationToken"/> that is signaled
@@ -68,7 +68,7 @@ public abstract partial class BaseSubscriptionServer : IOperationMessageProcesso
                 throw new ArgumentOutOfRangeException($"{nameof(options)}.{nameof(GraphQLHttpMiddlewareOptions.WebSockets)}.{nameof(GraphQLWebSocketOptions.KeepAliveTimeout)}");
 #pragma warning restore CA2208 // Instantiate argument exceptions correctly
         }
-        Client = sendStream ?? throw new ArgumentNullException(nameof(sendStream));
+        Connection = sendStream ?? throw new ArgumentNullException(nameof(sendStream));
         _cancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(sendStream.RequestAborted);
         CancellationToken = _cancellationTokenSource.Token;
     }
@@ -136,55 +136,55 @@ public abstract partial class BaseSubscriptionServer : IOperationMessageProcesso
     /// Executes upon a request to close the connection from the client.
     /// </summary>
     protected virtual Task OnCloseConnectionAsync()
-        => Client.CloseAsync();
+        => Connection.CloseAsync();
 
     /// <summary>
     /// Executes upon a request that has failed authorization.
     /// </summary>
     protected virtual Task ErrorAccessDeniedAsync()
-        => Client.CloseAsync(4401, "Access denied");
+        => Connection.CloseAsync(4401, "Access denied");
 
     /// <summary>
     /// Sends a fatal error message indicating that the initialization timeout has expired
     /// without the connection being initialized.
     /// </summary>
     protected virtual Task ErrorConnectionInitializationTimeoutAsync()
-        => Client.CloseAsync(4408, "Connection initialization timeout");
+        => Connection.CloseAsync(4408, "Connection initialization timeout");
 
     /// <summary>
     /// Sends a fatal error message indicating that the client attempted to initialize
     /// the connection more than one time.
     /// </summary>
     protected virtual Task ErrorTooManyInitializationRequestsAsync(OperationMessage message)
-        => Client.CloseAsync(4429, "Too many initialization requests");
+        => Connection.CloseAsync(4429, "Too many initialization requests");
 
     /// <summary>
     /// Sends a fatal error message indicating that the client attempted to subscribe
     /// to an event stream before initialization was complete.
     /// </summary>
     protected virtual Task ErrorNotInitializedAsync(OperationMessage message)
-        => Client.CloseAsync(4401, "Unauthorized");
+        => Connection.CloseAsync(4401, "Unauthorized");
 
     /// <summary>
     /// Sends a fatal error message indicating that the client attempted to use an
     /// unrecognized message type.
     /// </summary>
     protected virtual Task ErrorUnrecognizedMessageAsync(OperationMessage message)
-        => Client.CloseAsync(4400, "Unrecognized message");
+        => Connection.CloseAsync(4400, "Unrecognized message");
 
     /// <summary>
     /// Sends a fatal error message indicating that the client attempted to subscribe
     /// to an event stream with an empty id.
     /// </summary>
     protected virtual Task ErrorIdCannotBeBlankAsync(OperationMessage message)
-        => Client.CloseAsync(4400, "Id cannot be blank");
+        => Connection.CloseAsync(4400, "Id cannot be blank");
 
     /// <summary>
     /// Sends a fatal error message indicating that the client attempted to subscribe
     /// to an event stream with an id that was already in use.
     /// </summary>
     protected virtual Task ErrorIdAlreadyExistsAsync(OperationMessage message)
-        => Client.CloseAsync(4409, $"Subscriber for {message.Id} already exists");
+        => Connection.CloseAsync(4409, $"Subscriber for {message.Id} already exists");
 
     /// <summary>
     /// Authorizes an incoming GraphQL over WebSockets request with the connection initialization message.
@@ -206,7 +206,7 @@ public abstract partial class BaseSubscriptionServer : IOperationMessageProcesso
         // allocation-free authorization here
         var success = await AuthorizationHelper.AuthorizeAsync(
             new AuthorizationParameters<(BaseSubscriptionServer Server, OperationMessage Message)>(
-                Client.HttpContext,
+                Connection.HttpContext,
                 _options,
                 static info => info.Server.OnNotAuthenticatedAsync(info.Message),
                 static info => info.Server.OnNotAuthorizedRoleAsync(info.Message),
@@ -296,10 +296,10 @@ public abstract partial class BaseSubscriptionServer : IOperationMessageProcesso
          */
         async Task StartSmartKeepAliveLoopAsync()
         {
-            var lastKeepAliveSent = Client.LastMessageSentAt;
+            var lastKeepAliveSent = Connection.LastMessageSentAt;
             while (true)
             {
-                var lastSent = Client.LastMessageSentAt;
+                var lastSent = Connection.LastMessageSentAt;
                 var lastComm = lastKeepAliveSent > lastSent ? lastKeepAliveSent : lastSent;
                 var now = DateTime.UtcNow;
                 var timePassed = now.Subtract(lastComm);
--- a/src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
+++ b/src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
@@ -363,6 +363,13 @@ public abstract partial class BaseSubscriptionServer : IOperationMessageProcesso
             CancellationToken.ThrowIfCancellationRequested();
             if (!Subscriptions.Contains(messageId, dummyDisposer))
                 return;
+            // There is no support within the GraphQL spec for multiple streams to be returned.
+            // GraphQL.NET has validation rules to prevent this, but technically the execution
+            // strategy still allows for it (if the validation rule were not used).  Regardless, the
+            // protocol does not support such a scenario as there would need to be separate
+            // IDs for each stream.  So, if there were to be more than one stream returned,
+            // execution would fall through to SendErrorResultAsync -- but so long as the
+            // validation rules are in place, that should not be possible.
             if (result.Streams?.Count == 1)
             {
                 // do not return a result, but set up a subscription
--- a/src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
+++ b/src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
@@ -133,55 +133,55 @@ public abstract partial class BaseSubscriptionServer : IOperationMessageProcesso
     /// Executes upon a request to close the connection from the client.
     /// </summary>
     protected virtual Task OnCloseConnectionAsync()
-        => Client.CloseConnectionAsync();
+        => Client.CloseAsync();
 
     /// <summary>
     /// Executes upon a request that has failed authorization.
     /// </summary>
     protected virtual Task ErrorAccessDeniedAsync()
-        => Client.CloseConnectionAsync(4401, "Access denied");
+        => Client.CloseAsync(4401, "Access denied");
 
     /// <summary>
     /// Sends a fatal error message indicating that the initialization timeout has expired
     /// without the connection being initialized.
     /// </summary>
     protected virtual Task ErrorConnectionInitializationTimeoutAsync()
-        => Client.CloseConnectionAsync(4408, "Connection initialization timeout");
+        => Client.CloseAsync(4408, "Connection initialization timeout");
 
     /// <summary>
     /// Sends a fatal error message indicating that the client attempted to initialize
     /// the connection more than one time.
     /// </summary>
     protected virtual Task ErrorTooManyInitializationRequestsAsync(OperationMessage message)
-        => Client.CloseConnectionAsync(4429, "Too many initialization requests");
+        => Client.CloseAsync(4429, "Too many initialization requests");
 
     /// <summary>
     /// Sends a fatal error message indicating that the client attempted to subscribe
     /// to an event stream before initialization was complete.
     /// </summary>
     protected virtual Task ErrorNotInitializedAsync(OperationMessage message)
-        => Client.CloseConnectionAsync(4401, "Unauthorized");
+        => Client.CloseAsync(4401, "Unauthorized");
 
     /// <summary>
     /// Sends a fatal error message indicating that the client attempted to use an
     /// unrecognized message type.
     /// </summary>
     protected virtual Task ErrorUnrecognizedMessageAsync(OperationMessage message)
-        => Client.CloseConnectionAsync(4400, "Unrecognized message");
+        => Client.CloseAsync(4400, "Unrecognized message");
 
     /// <summary>
     /// Sends a fatal error message indicating that the client attempted to subscribe
     /// to an event stream with an empty id.
     /// </summary>
     protected virtual Task ErrorIdCannotBeBlankAsync(OperationMessage message)
-        => Client.CloseConnectionAsync(4400, "Id cannot be blank");
+        => Client.CloseAsync(4400, "Id cannot be blank");
 
     /// <summary>
     /// Sends a fatal error message indicating that the client attempted to subscribe
     /// to an event stream with an id that was already in use.
     /// </summary>
     protected virtual Task ErrorIdAlreadyExistsAsync(OperationMessage message)
-        => Client.CloseConnectionAsync(4409, $"Subscriber for {message.Id} already exists");
+        => Client.CloseAsync(4409, $"Subscriber for {message.Id} already exists");
 
     /// <summary>
     /// Authorizes an incoming GraphQL over WebSockets request with the connection initialization message.
--- a/src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
+++ b/src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
@@ -12,7 +12,7 @@ namespace GraphQL.Server.Transports.AspNetCore.WebSockets;
 /// </summary>
 public abstract partial class BaseSubscriptionServer : IOperationMessageProcessor
 {
-    private volatile int _initialized;
+    private int _initialized;
     private CancellationTokenSource? _cancellationTokenSource;
     private readonly GraphQLWebSocketOptions _options;
     private readonly IAuthorizationOptions _authorizationOptions;
@@ -86,7 +86,7 @@ public abstract partial class BaseSubscriptionServer : IOperationMessageProcesso
             _ = Task.Run(async () =>
             {
                 await Task.Delay(connectInitWaitTimeout, CancellationToken); // CancellationToken is set when this class is disposed
-                if (_initialized == 0)
+                if (!Initialized)
                     await OnConnectionInitWaitTimeoutAsync();
             });
         }
@@ -121,7 +121,7 @@ public abstract partial class BaseSubscriptionServer : IOperationMessageProcesso
     /// Indicates if the connection has been already initialized.
     /// </summary>
     protected bool Initialized
-        => _initialized == 1;
+        => Thread.VolatileRead(ref _initialized) == 1;
 
     /// <summary>
     /// Sets the initialized flag if it has not already been set.
@@ -132,6 +132,7 @@ public abstract partial class BaseSubscriptionServer : IOperationMessageProcesso
 
     /// <summary>
     /// Executes when a message has been received from the client.
+    /// This method should essentially implement a state machine for the implemented protocol.
     /// </summary>
     /// <exception cref="OperationCanceledException"/>
     public abstract Task OnMessageReceivedAsync(OperationMessage message);
@@ -271,7 +272,8 @@ public abstract partial class BaseSubscriptionServer : IOperationMessageProcesso
             return;
         }
         await OnConnectionAcknowledgeAsync(message);
-        TryInitialize();
+        if (TryInitialize() == false)
+            return;
 
         var keepAliveTimeout = _options.KeepAliveTimeout ?? DefaultKeepAliveTimeout;
         if (keepAliveTimeout > TimeSpan.Zero)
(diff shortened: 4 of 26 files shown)


sungam3r:
I will review this next after #806/#807 being merged.

Shane32:
> I will review this next after #806/#807 being merged.

@sungam3r If you approve #806 , merge it in to develop, and merge develop in here.  This PR should then compile and is ready for review.

This PR is larger than the rest as it contains somewhat dependent logic; you will likely want to read one of the `SubscriptionServer` implementations to understand how the it works with the `BaseSubscriptionServer` class.  I have split them into separate files as much as makes sense.  Also, any changes to the `BaseSubscriptionServer` will likely impact the other two `SubscriptionServer` classes.

I suggest approaching the review as follows:
1. Review the graphql-ws specification here: [src/Transports.AspNetCore/WebSockets/GraphQLWs/MessageType.cs](https://github.com/graphql-dotnet/server/pull/810/files#diff-e90e3db14131339e94eb8fa9c2bc8f30adcb6187be44f1c8e3014559eed8d433)
2. Review the `InitializeConnectionAsync` method of [src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs](https://github.com/graphql-dotnet/server/pull/810/files#diff-b861614bf0bcea9ab9253f8f38cd7db98fea89ab2c7e3398a1453a72e9a6aaa9)
3. Review the `OnMessageReceivedAsync` method of [src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs](https://github.com/graphql-dotnet/server/pull/810/files#diff-93750d4f2bb82a42a1b696bc4a564e75f75f316dfc5b4b62d3630e1f589e3c44)

That, and the methods they reference, contain most of the logic.  Then you can continue to follow the logic for data events from the event source here:

[src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.Observer.cs](https://github.com/graphql-dotnet/server/pull/810/files#diff-fc2abda965d4a9bd737fa0bc957fb310a4ceaca8e5226a14c347931961824d6c)

At this point you should be able to perform a full review of all of those files, and subsequently review the SubscriptionsTransportWs protocol and server.

Shane32:
@sungam3r What's the ETA for review?

sungam3r:
Was busy at home and other projects. I took only simple PRs to review last 2-3 days. I'll start to review this one tomorrow morning.

sungam3r:
@Shane32 I'm reviewing this one now. Was busy yesterday.

src/Transports.AspNetCore/WebSockets/SubscriptionsTransportWs/MessageType.cs
```diff
@@ -0,0 +1,92 @@
...
+/// </summary>
+public static class MessageType
+{
+    /// <summary>
+    ///     Client sends this message after plain websocket connection to start the communication with the server
+    ///     The server will response only with GQL_CONNECTION_ACK + GQL_CONNECTION_KEEP_ALIVE(if used) or GQL_CONNECTION_ERROR
+    ///     to this message.
+    ///     payload: Object : optional parameters that the client specifies in connectionParams
+    /// </summary>
+    public const string GQL_CONNECTION_INIT = "connection_init";
```
sungam3r, line 17:
```suggestion
    public const string GQL_CONNECTION_INIT = "connection_init"; // Client -> Server
```

src/Transports.AspNetCore/WebSockets/SubscriptionsTransportWs/MessageType.cs
```diff
@@ -0,0 +1,92 @@
...
+
+    /// <summary>
+    ///     Client sends this message to execute GraphQL operation
+    ///     id: string : The id of the GraphQL operation to start
+    ///     payload: Object:
+    ///     query: string : GraphQL operation as string or parsed GraphQL document node
+    ///     variables?: Object : Object with GraphQL variables
+    ///     operationName?: string : GraphQL operation name
+    /// </summary>
+    public const string GQL_START = "start";
```
sungam3r, line 58:
```suggestion
    public const string GQL_START = "start"; // Client -> Server
```

src/Transports.AspNetCore/WebSockets/GraphQLWs/MessageType.cs
```diff
@@ -0,0 +1,155 @@
...
+    /// Indicates that the client wants to establish a connection within the existing socket.
+    /// This connection is not the actual WebSocket communication channel, but is rather a
+    /// frame within it asking the server to allow future operation requests.
+    /// </para>
+    /// <para>
+    /// The server must receive the connection initialization message within the allowed waiting
+    /// time specified in the <see cref="GraphQLWebSocketOptions.ConnectionInitWaitTimeout"/>
+    /// parameter during the server setup.  If the client does not request a connection within
+    /// the allowed timeout, the server will close the socket with the event:
+    /// <i>4408: Connection initialization timeout</i>.
```
sungam3r, line 24: initiali**s**ation in the original spec 😕

src/Transports.AspNetCore/WebSockets/GraphQLWs/MessageType.cs
```diff
@@ -0,0 +1,155 @@
...
+    /// </para>
+    /// <para>
+    /// If the server receives more than one <see cref="ConnectionInit"/> message at any
+    /// given time, the server will close the socket with the event
+    /// <i>4429: Too many initialization requests</i>.
```
sungam3r, line 29: the same here

src/Transports.AspNetCore/WebSockets/GraphQLWs/MessageType.cs
```diff
@@ -0,0 +1,155 @@
...
+    /// <summary>
+    /// <para>
+    /// Expected response to the <see cref="ConnectionInit"/> message from the client acknowledging
+    /// a successful connection with the server.
+    /// </para>
+    /// <para>
+    /// The server can use the optional <see cref="OperationMessage.Payload"/> field to transfer additional details about the connection.
+    /// </para>
+    /// <para>
+    /// The client is now ready to request subscription operations.
```
sungam3r, line 47: Just a note - not only subscriptions operations.

src/Transports.AspNetCore/WebSockets/GraphQLWs/MessageType.cs
```diff
@@ -0,0 +1,155 @@
...
+    /// </para>
+    /// <para>
+    /// The optional <see cref="OperationMessage.Payload"/> field can be used to transfer additional details about the ping.
+    /// </para>
+    /// </summary>
+    public const string Ping = "ping"; // Bidirectional
+
+    /// <summary>
+    /// <para>
+    /// The response to the <see cref="Ping"/> message. Must be sent as soon as the Ping message is received.
```
sungam3r, line 144:
```suggestion
    /// The response to the <see cref="Ping"/> message. Must be sent as soon as the <see cref="Ping"/> message is received.
```

sungam3r:
I fixed noisy CodeQL annotations that doubled screen space to scroll.

sungam3r:
Did I understand correctly that the server would support both protocols at the same time?

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,488 @@
...
+{
+    private volatile int _initialized;
+    private CancellationTokenSource? _cancellationTokenSource;
+    private readonly GraphQLHttpMiddlewareOptions _options;
+
+    /// <summary>
+    /// Returns a <see cref="IWebSocketConnection"/> instance that can be used
+    /// to send messages to the client.
+    /// </summary>
+    protected IWebSocketConnection Client { get; }
```
sungam3r, line 23:
```suggestion
    protected IWebSocketConnection Connection { get; }
```

I would rename. It looks more familiar.

sungam3r:
I read the specification of both protocols. I will return today with additional comments. In our company we have a goal to start using complexity analysis, so I have to do several things at once.

Shane32:
> Did I understand correctly that the server would support both protocols at the same time?

Yes, depending on the protocol negotiated when the connection is opened (see #812 ). You cannot use commands from each protocol within the same connection.

src/Transports.AspNetCore/WebSockets/GraphQLWs/MessageType.cs
Shane32: I probably fixed any other obvious language flaws in the text but I think I left most of the text the same.  I can't remember.

src/Transports.AspNetCore/WebSockets/GraphQLWs/MessageType.cs
Shane32: Correct.

sungam3r:
> You cannot use commands from each protocol within the same connection.

Of course.

src/Transports.AspNetCore/WebSockets/GraphQLWs/MessageType.cs
```diff
@@ -0,0 +1,155 @@
+using GraphQL.Transport;
+
+namespace GraphQL.Server.Transports.AspNetCore.WebSockets.GraphQLWs;
+
+/// <summary>
+/// Protocol message types defined in
+/// <see href="https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md"/>
+/// and
```
sungam3r, line 8: Why did you mention https://github.com/enisdenjo/graphql-ws/blob/master/src/common.ts ? It's implementation details for JS platform, not protocol bits.

src/Transports.AspNetCore/WebSockets/GraphQLWs/MessageType.cs
sungam3r: ok

src/Transports.AspNetCore/WebSockets/GraphQLWs/MessageType.cs
Shane32: I think it had more error code IDs than were listed in the protocol.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,488 @@
...
+    /// </summary>
+    protected virtual TimeSpan DefaultKeepAliveTimeout { get; } = Timeout.InfiniteTimeSpan;
+
+    /// <summary>
+    /// Returns the default connection timeout.
+    /// </summary>
+    protected virtual TimeSpan DefaultConnectionTimeout { get; } = TimeSpan.FromSeconds(10);
+
+    /// <summary>
+    /// Initailizes a new instance with the specified parameters.
```
sungam3r, line 48:
```suggestion
    /// Initializes a new instance with the specified parameters.
```

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,488 @@
...
+    /// </summary>
+    /// <param name="sendStream">The WebSockets stream used to send data packets or close the connection.</param>
+    /// <param name="options">Configuration options for this instance.</param>
+    public BaseSubscriptionServer(
```
sungam3r, line 52:
```suggestion
    protected BaseSubscriptionServer(
```

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,488 @@
...
+        IWebSocketConnection sendStream,
```
sungam3r, line 53:
```suggestion
        IWebSocketConnection connection,
```

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,488 @@
...
+        GraphQLHttpMiddlewareOptions options)
```
sungam3r, line 54: `GraphQLHttpMiddlewareOptions` seems excessive here except its usage in `AuthorizeAsync` (3 AuthXXX properties).  In ideal subscription server should accept `GraphQLWebSocketOptions`.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,488 @@
...
+    /// returns an unsuccessful <see cref="AuthorizationResult"/> for the specified policy.
+    /// <br/><br/>
+    /// Default implementation closes the WebSocket connection with a 4401 'Access denied'
+    /// error message via <see cref="ErrorAccessDeniedAsync"/>.
+    /// </summary>
+    protected virtual Task OnNotAuthorizedPolicyAsync(OperationMessage message, AuthorizationResult result)
+        => ErrorAccessDeniedAsync();
+
+    /// <summary>
+    /// Executes when the client is attempting to initalize the connection.
```
sungam3r, line 251:
```suggestion
    /// Executes when the client is attempting to initialize the connection.
```

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,488 @@
...
+        {
+            cts.Cancel();
+            cts.Dispose();
+            Subscriptions.Dispose();
+        }
+        GC.SuppressFinalize(this);
+    }
+
+    /// <summary>
+    /// Indicates if the connection has been initialized yet.
```
sungam3r, line 117:
```suggestion
    /// Indicates if the connection has been already initialized.
```

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,488 @@
...
+    /// <summary>
+    /// Executes when a message has been received from the client.
+    /// </summary>
+    /// <exception cref="OperationCanceledException"/>
+    public abstract Task OnMessageReceivedAsync(OperationMessage message);
+
+    /// <summary>
+    /// Executes upon a request to close the connection from the client.
+    /// </summary>
+    protected virtual Task OnCloseConnectionAsync()
```
sungam3r, line 138: I see only old protocol uses it reacting on `GQL_CONNECTION_TERMINATE`. What is alternative to terminate connection for the new protocol?

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,488 @@
...
+
+    /// <summary>
+    /// Executes when the client is attempting to initalize the connection.
+    /// <br/><br/>
+    /// By default, this first checks <see cref="AuthorizeAsync(OperationMessage)"/> to validate that the
+    /// request has passed authentication.  If validation fails, the connection is closed with an Access
+    /// Denied message.
+    /// <br/><br/>
+    /// Otherwise, the connection is acknowledged via <see cref="OnConnectionAcknowledgeAsync(OperationMessage)"/>,
+    /// <see cref="TryInitialize"/> is called to indicate that this WebSocket connection is ready to accept requests,
```
sungam3r, line 258:
```suggestion
    /// <see cref="TryInitialize"/> is called to indicate that this WebSocket connection is ready to accept requests,
```

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,488 @@
...
+        {
+            if (smartKeepAlive)
+                _ = StartSmartKeepAliveLoopAsync();
+            else
+                _ = StartKeepAliveLoopAsync();
+        }
+
+        async Task StartKeepAliveLoopAsync()
+        {
+            while (true)
```
sungam3r, line 283:
```suggestion
            while (!CancellationToken.IsCancellationRequested)
```

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,488 @@
...
+                await Task.Delay(keepAliveTimeout, CancellationToken);
+                await OnSendKeepAliveAsync();
+            }
+        }
+
+        /*
+         * This 'smart' keep-alive logic doesn't work with subscription-transport-ws because the client will
+         * terminate a connection 20 seconds after the last keep-alive was received (or it will never
+         * terminate if no keep-alive was ever received).
+         *
```
sungam3r, line 294:
```suggestion
         *
```

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,488 @@
...
+         * This 'smart' keep-alive logic doesn't work with subscription-transport-ws because the client will
+         * terminate a connection 20 seconds after the last keep-alive was received (or it will never
+         * terminate if no keep-alive was ever received).
+         * 
+         * See: https://github.com/graphql/graphql-playground/issues/1247
+         */
+        async Task StartSmartKeepAliveLoopAsync()
+        {
+            var lastKeepAliveSent = Connection.LastMessageSentAt;
+            while (true)
```
sungam3r, line 300:
```suggestion
            while (!CancellationToken.IsCancellationRequested)
```

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,488 @@
...
+            {
+                var lastSent = Connection.LastMessageSentAt;
+                var lastComm = lastKeepAliveSent > lastSent ? lastKeepAliveSent : lastSent;
```
sungam3r, line 303: What is Comm ?

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,488 @@
...
+    /// </summary>
+    protected abstract Task OnSendKeepAliveAsync();
+
+    /// <summary>
+    /// Executes when a connection request needs to be acknowledged.
+    /// </summary>
+    protected abstract Task OnConnectionAcknowledgeAsync(OperationMessage message);
+
+    /// <summary>
+    /// Executes when a new subscription request has occurred.
```
sungam3r, line 333:
```suggestion
    /// Executes when a new GraphQL request (typically subscription) has occurred.
```

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,488 @@
...
+                {
+                    await ErrorIdAlreadyExistsAsync(message);
+                    return;
+                }
+            }
+
+            var result = await ExecuteRequestAsync(message);
+            CancellationToken.ThrowIfCancellationRequested();
+            if (!Subscriptions.Contains(messageId, dummyDisposer))
+                return;
```
sungam3r, line 365: concurrency conditions?

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,488 @@
...
+            if (result.Streams?.Count == 1)
```
sungam3r, line 366: There may be more than 1 stream?

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,488 @@
...
+                await SendSingleResultAsync(message, result);
+            }
+            else
+            {
+                await SendErrorResultAsync(message, result);
+            }
+        }
+        catch (OperationCanceledException) when (CancellationToken.IsCancellationRequested)
+        {
+            throw;
```
sungam3r, line 395: In `WebSocketConnection` exception is swallowed:
```cs
 catch (WebSocketException) when (RequestAborted.IsCancellationRequested)
        {
        }
```

Why do you throw here?

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
sungam3r: And what does this condition mean? I see 3 places of use.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,488 @@
...
+    }
+
+    /// <summary>
+    /// Creates an <see cref="ExecutionError"/> for an unknown <see cref="Exception"/>.
+    /// </summary>
+    protected virtual Task<ExecutionError> HandleErrorDuringSubscribeAsync(OperationMessage message, Exception ex)
+        => Task.FromResult<ExecutionError>(new UnhandledError("Unable to set up subscription for the requested field.", ex));
+
+    /// <summary>
+    /// Sends a single result to the client for a subscription or request, along with a notice
```
sungam3r, line 419: > for a subscription **or** request

Or **only** for requests (query/mutation) ?

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,488 @@
...
+    /// that it was the last result in the event stream.
+    /// </summary>
+    protected virtual async Task SendSingleResultAsync(OperationMessage message, ExecutionResult result)
+    {
+        await SendDataAsync(message.Id!, result);
+        await SendCompletedAsync(message.Id!);
+    }
+
+    /// <summary>
+    /// Sends an execution error to the client during set-up of a subscription.
```
sungam3r, line 429:
```suggestion
    /// Sends an execution error to the client during set-up of a GraphQL request (typically subscription).
```

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,488 @@
...
+        => SendErrorResultAsync(message.Id!, executionError);
+
+    /// <summary>
+    /// Sends an execution error to the client during set-up of a subscription.
+    /// </summary>
+    protected virtual Task SendErrorResultAsync(string id, ExecutionError executionError)
+        => SendErrorResultAsync(id, new ExecutionResult { Errors = new ExecutionErrors { executionError } });
+
+    /// <summary>
+    /// Sends an error result to the client during set-up of a subscription.
```
sungam3r, line 441:
```suggestion
    /// Sends an error result to the client during set-up of a GraphQL request (typically subscription).
```

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,488 @@
...
+    /// </summary>
+    protected virtual Task SendErrorResultAsync(OperationMessage message, ExecutionResult result)
+        => SendErrorResultAsync(message.Id!, result);
+
+    /// <summary>
+    /// Sends an error result to the client during set-up of a subscription.
+    /// </summary>
+    protected abstract Task SendErrorResultAsync(string id, ExecutionResult result);
```
sungam3r, line 449: 4 overloads for `SendErrorResultAsync` look excessive.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,488 @@
...
+
+    /// <summary>
+    /// Sends a data packet to the client for a subscription event.
```
sungam3r, line 452:
```suggestion
    /// Sends a data packet to the client for a GraphQL request (typically subscription event).
```

I would check all xml comments to focus on the fact that any requests can be performed and not just subscriptions.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,488 @@
...
+    /// Sends a notice that a subscription has completed and no more data packets will be sent.
+    /// </summary>
+    protected abstract Task SendCompletedAsync(string id);
+
+    /// <summary>
+    /// Executes a GraphQL request. The request is inside <see cref="OperationMessage.Payload"/>
+    /// and will need to be deserialized by <see cref="IGraphQLSerializer.ReadNode{T}(object?)"/>
+    /// into a <see cref="GraphQLRequest"/> instance.
+    /// </summary>
+    protected abstract Task<ExecutionResult> ExecuteRequestAsync(OperationMessage message);
```
sungam3r, line 466: Both implementations are identical and code copy-pasted 😕 . Do you intend to keep it as is to keep implementations simple and independent?

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.Observer.cs
```diff
@@ -0,0 +1,88 @@
+namespace GraphQL.Server.Transports.AspNetCore.WebSockets;
+
+public abstract partial class BaseSubscriptionServer
+{
+    /// <summary>
+    /// Handles messages from the event source.
+    /// </summary>
+    private class Observer : IObserver<ExecutionResult>
+    {
+        private readonly BaseSubscriptionServer _handler;
```
sungam3r, line 10:
```suggestion
        private readonly BaseSubscriptionServer _server;
```

We already have processor interface. Let's no introduce new terms.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.Observer.cs
```diff
@@ -0,0 +1,88 @@
...
+            }
+            catch { }
+        }
+
+        public async void OnNext(ExecutionResult value)
+        {
+            if (Interlocked.CompareExchange(ref _done, 0, 0) == 1)
+                return;
+            if (value == null)
+                return;
```
sungam3r, line 76:
```suggestion
            if (value == null || Interlocked.CompareExchange(ref _done, 0, 0) == 1)
                return;
```

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.Observer.cs
```diff
@@ -0,0 +1,88 @@
...
+        public async void OnError(Exception error)
+        {
+            if (_closeAfterOnError)
+            {
+                if (Interlocked.Exchange(ref _done, 1) == 1)
+                    return;
+            }
+            else
+            {
+                if (Interlocked.CompareExchange(ref _done, 0, 0) == 1)
```
sungam3r, line 44:
```suggestion
                if (Thread.VolatileRead(ref _done) == 1)
```

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.Observer.cs
```diff
@@ -0,0 +1,88 @@
...
+                    return;
+            }
```
sungam3r, line 46:
```suggestion
            if (Thread.VolatileRead(ref _done) == 1)
                return;

            if (_closeAfterOnError && Interlocked.Exchange(ref _done, 1) == 1)
                return;
```

Looks more simple.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
Shane32: communication

Shane32: renamed

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
Shane32: Changed to "for a query or mutation request"

src/Transports.AspNetCore/WebSockets/SubscriptionsTransportWs/SubscriptionServer.cs
```diff
@@ -0,0 +1,243 @@
...
+    /// <inheritdoc/>
+    protected override async Task SendErrorResultAsync(string id, ExecutionResult result)
+    {
+        if (Subscriptions.TryRemove(id))
+        {
+            await Connection.SendMessageAsync(new OperationMessage
+            {
+                Id = id,
+                Type = MessageType.GQL_ERROR,
+                Payload = result,
```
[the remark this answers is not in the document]
Shane32: The two protocols differ in this respect.  It is necessary to review the reference implementation to determine this fact for the old protocol, if I remember right.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
Shane32: I fail to understand; OCE should always throw, and does so in WebSocketConnection as well.

Are you suggestion I should add this?

```cs
catch (WebSocketException)
{
    throw; // handled in WebSocketConnection
}
```

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
Shane32: There is no support within the GraphQL spec for multiple streams to be returned.  GraphQL.NET has validation rules to prevent this, but I think the execution strategy still allows for it (if the validation rule were not used).  Regardless, the protocol does not support such a scenario as there would need to be separate IDs for each stream.  So, if there were to be more than one stream returned, execution would fall through to SendErrorResultAsync -- but so long as the validation rules are in place, that should not be possible.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
Shane32: The new protocol simply uses the websocket close message to gracefully terminate the connection.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
Shane32: `return` would be hit if the client had both requested cancellation of the subscription, and then started a new one with the same id.  According to spec, once the use requests cancellation of the subscription, no more responses must be given -- including single responses or error responses.  So because I have to await the DocumentExecuter to set up the IObservable to begin with, I register a new instance of a temporary object (dummyDisposer) temporarily, and use these special checks to validate that it is still the active subscription for that id.

Shane32: (Sorry it seems so complicated, but I want to implement it correctly.)

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
Shane32: Yeah, although they are just overloads and it shouldn't hurt.  There's a reason also: one pair has `OperationMessage` representing the original message that triggered the error response, whereas the other pair just has the id.  In this way you can add logging or whatever from the original message.  And for the second parameter, there's a single error vs a whole ExecutionResult.  Seems to make sense, even if there are a few extra overloads here.

FYI my code uses 3 of the 4 overloads, and the 4th is called by one of the other 3 overloads.  I wouldn't change it.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
sungam3r: > Are you suggesting I should add this?

No. My question was about throw or not here when cathing OCE .

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
sungam3r, line 366:
```suggestion
            // There is no support within the GraphQL spec for multiple streams to be returned.
            // GraphQL.NET has validation rules to prevent this, but technically the execution
            // strategy still allows for it (if the validation rule were not used). Regardless, the
            // protocol does not support such a scenario as there would need to be separate
            // IDs for each stream. So, if there were to be more than one stream returned,
            // execution would fall through to SendErrorResultAsync -- but so long as the
            // validation rules are in place, that should not be possible.
            if (result.Streams?.Count == 1)
```

src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
```diff
@@ -0,0 +1,254 @@
...
+
+/// <inheritdoc/>
+public class SubscriptionServer : BaseSubscriptionServer
+{
+    private readonly IWebSocketAuthenticationService? _authenticationService;
+
+    /// <summary>
+    /// The WebSocket sub-protocol used for this protocol.
+    /// </summary>
+    public const string SubProtocol = "graphql-transport-ws";
```
sungam3r, line 16: It looks like you flipped protocols.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
sungam3r: 🤔 Let's consider such option - imagine separate auth settings for "usual" http middleware and for WebSocket transports. So copy 3 settings into `GraphQLWebSocketOptions`. Then add interface like `IAuthorizationOptions` with 3 settings and implement it by both `GraphQLHttpMiddlewareOptions` and `GraphQLWebSocketOptions`. Then change ` GraphQLHttpMiddlewareOptions middlewareOptions` to ` IAuthorizationOptions authorizationOptions` into `AuthorizationParameters`. Viola! Auth works for both transports independently. Even further - remove `WebSockets` property from `GraphQLHttpMiddlewareOptions` since now options are independent. `WebSocketConnection` can work with `GraphQLWebSocketOptions` directly. The same for `BaseSubscriptionServer`.

sungam3r: My intent to not mix options for middleware and for websockets.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
sungam3r: This definitely needs to be noted in the comments.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
Shane32: I don’t want another copy of the auth settings as visible to the user. They are not logically separate settings. We can make a third options class and copy the applicable settings into it from GraphQLHttpMiddlewareOptions and GraphQLWebSocketOptions if you like.

Shane32: WebSocket is just another form of transport, just as POST is different than GET. Clients can request query or mutations via GET, POST or WebSocket. Within POST they can submit via multipart content type or json content type. None of that affects the authorization requirements.

Shane32: Another option is to have the WebSocket constructor accept both GraphQLWebSocketOptions and IAuthorizationOptions.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
sungam3r: > In this way you can add logging or whatever from the original message.

Yes, but one can override both overloads adding logging and then call `base.SendErrorResultAsync` so logging will be doubled.

sungam3r: I mean caller does not know the implementation details. It sees 4 overload for SendErrorResultAsync. What should he do to add logging? I would override all that overloads.

src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
Shane32: It's stupid, but correct.  The repo at https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md (with npm package named `graphql-ws`) uses the sub-protocol `graphql-transport-ws`.  The repo at https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md (with npm package named `subscriptions-transport-ws`) uses the sub-protocol `graphql-ws`.  So here I've named the folders and namespaces according to the npm package name, which seems to be the most common naming (as they are both just called "GraphQL over WebSockets") -- and then have a const for the correct sub-protocol to use.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
sungam3r: > I don’t want another copy of the auth settings as visible to the user. They are not logically separate settings.

Why not? One can enable auth on one transport and disable on other. Or use policy for HTTP auth and roles for WebSocket.

изображение
It's not true. I do not like that this options object is passed to those places where its properties are completely not needed and not applicable. `AuthorizationParameters` is only place that holds reference to `GraphQLHttpMiddlewareOptions` and forces `BaseSubscriptionServer` to use them.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
Shane32: https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md

> The client closes the socket and the connection by dispatching a 1000: Normal Closure close event to the server indicating a normal closure.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
Shane32: Ok, I'll work on improving comments within this code block.

src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
sungam3r: WOW! We definitely should leave a comment about this so that the reader does not have an impression of an error (like mine).

Shane32: Agreed.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
Shane32: Well, actually you would need to override multiple of them anyway, assuming you want to log information that is not present in the abstract method (meaning information from OperationMessage).  That's part of the benefit of having the overloads in the first place.  Let's add comments here, if you like, and move on.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
sungam3r: > I don’t want another copy of the auth settings as visible to the user. They are not logically separate settings. We can make a **third options class and copy the applicable settings into it** from GraphQLHttpMiddlewareOptions and GraphQLWebSocketOptions if you like.

I will not figure out how it looks.

src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
Shane32:
```suggestion
    /// <remarks>
    /// Please note that the correct sub-protocol for the
    /// <see href="https://github.com/enisdenjo/graphql-ws">graphql-ws</see>
    /// protocol is <c>graphql-transport-ws</c> as is defined here.
    /// </remarks>
    public const string SubProtocol = "graphql-transport-ws";
```

src/Transports.AspNetCore/WebSockets/SubscriptionsTransportWs/SubscriptionServer.cs
```diff
@@ -0,0 +1,243 @@
...
+
+/// <inheritdoc/>
+public class SubscriptionServer : BaseSubscriptionServer
+{
+    private readonly IWebSocketAuthenticationService? _authenticationService;
+
+    /// <summary>
+    /// The WebSocket sub-protocol used for this protocol.
+    /// </summary>
+    public const string SubProtocol = "graphql-ws";
```
Shane32, line 16:
```suggestion
    /// <remarks>
    /// Please note that the correct sub-protocol for the
    /// <see href="https://github.com/apollographql/subscriptions-transport-ws">subscriptions-transport-ws</see>
    /// protocol is <c>graphql-ws</c> as is defined here.
    /// </remarks>
    public const string SubProtocol = "graphql-ws";
```

src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
sungam3r: 👍

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
sungam3r: You reason from the position that the client knows the implementation. I explain that these additional overloads are confused and I would not say that they have more benefits than harm in this regard. Yes, at least add a note that user should override only last overload with id and result.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
Shane32: Sending query/mutation GraphQL requests over a WebSocket channel is not unlike a [HTTP persistent connection.](https://en.wikipedia.org/wiki/HTTP_persistent_connection)

Shane32: I think you should understand that one of my goals here is to make supporting WebSocket connections easy to use.  Currently you need to all of this:
1. Add another nuget package, which bring additional dependencies
2. Add two additional services in the graphql builder
3. Configure ASP.NET Core to enable websocket support
4. Add another line in the app builder to configure the websocket endpoint

With the new code:
- No additional nuget package is needed (and there are no additional dependencies)
- No additional configuration necessary in GraphQL builder
- Still need to enable ASP.NET Core websocket support
- No additional code in app builder

Basically, assuming you've enabled WebSockets through ASP.NET Core, it just works.  And if the authorization options are separate, I see a great danger that someone sets authorization options for the main middleware and forgets to set them in the websocket options as well.  Not only is it typically redundant, but this is a huge security concern.

There actually already exists an easy way to configure different security options for websocket transport.  Just add another line in the app builder.

```cs
app.UseGraphQL("/graphql", opts => {
    opts.HandleWebSockets = false;
    // set auth settings for get/post here
});
app.UseGraphQL("/graphql", opts => {
    opts.HandleGet = false;
    opts.HandlePost = false;
    // set auth settings for websockets here
});
```

(I wouldn't even document this, I see it as such a rare use case.)

But that is much better than the alternative scenario, where someone accidentally only sets auth options only for get/post and forgets to set them on websocket connections.

sungam3r: > My intent is to not mix options for middleware and for websockets.

I'm open to resolve this in any way.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
Shane32: #818

Shane32:
At this point I am awaiting further comments from you @sungam3r

src/Transports.AspNetCore/AuthorizationHelper.cs
```diff
@@ -27,7 +27,7 @@ public static async ValueTask<bool> AuthorizeAsync<TState>(AuthorizationParamete
             }
         }
 
-        if (options.AuthorizedRoles?.Count > 0)
+        if (options.AuthorizedRoles?.Count() > 0)
```
sungam3r, line 30:
```suggestion
        if (options.AuthorizedRoles?.Any() == true)
```

src/Transports.AspNetCore/IAuthorizationOptions.cs
```diff
@@ -0,0 +1,37 @@
...
+/// Provides authorization options.
+/// </summary>
+public interface IAuthorizationOptions
+{
+    /// <summary>
+    /// If set, requires that <see cref="IIdentity.IsAuthenticated"/> return <see langword="true"/>
+    /// for the user within <see cref="HttpContext.User"/>
+    /// prior to executing the GraphQL request or accepting the WebSocket connection.
+    /// Technically this property should be named as AuthenticationRequired but for
+    /// ASP.NET Core / GraphQL.NET naming and design decisions it was called so.
```
sungam3r, line 18:
```suggestion
    /// ASP.NET Core / GraphQL.NET naming and design decisions it was called so.
```

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.Observer.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        public async void OnError(Exception error)
+        {
+            if (Thread.VolatileRead(ref _done) == 1)
+                return;
+            if (_closeAfterOnError && Interlocked.Exchange(ref _done, 1) == 1)
+                return;
+            try
+            {
+                if (error != null)
```
sungam3r, line 43: Why does this null-check here? And how does it correlate with `if (_closeAfterOnError && Interlocked.Exchange(ref _done, 1) == 1)` condition ?

Shane32: We do not have control of whether a event source calls `OnError` with `null` or an actual exception.  They certainly should not pass in `null`.  Nevertheless, I wrote code to prevent a NRE in that case.  There is no requirement that a message be sent in this case, so it is valid from a protocol point of view.

`_closeAfterOnError` behaves just as it should, closing the connection if it is `true`.  There is no requirement that an error packet/message be sent first.

sungam3r: Please add a comment about that to clarify intent.

sungam3r:
изображение
Almost done. I need to switch to daily work. I will continue tomorrow.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.Observer.cs
Shane32:
```suggestion
                // although error should never be null, if the event source does call OnError(null),
                // we will skip sending an error packet/message (allowed by spec)
                if (error != null)
```

sungam3r: 👍

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,502 @@
...
+            Subscriptions.Dispose();
+        }
+        GC.SuppressFinalize(this);
+    }
+
+    /// <summary>
+    /// Indicates if the connection has been already initialized.
+    /// </summary>
+    protected bool Initialized
+        => _initialized == 1;
```
sungam3r, line 124:
```suggestion
        => Thread.VolatileRead(ref _initialized) == 1;
```

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,502 @@
...
+using Microsoft.AspNetCore.Http;
+
+namespace GraphQL.Server.Transports.AspNetCore.WebSockets;
+
+/// <summary>
+/// Manages a WebSocket message stream.
+/// </summary>
+public abstract partial class BaseSubscriptionServer : IOperationMessageProcessor
+{
+    private volatile int _initialized;
```
sungam3r, line 15:
```suggestion
    private int _initialized;
```

volatile is not required for Interlocked API.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,502 @@
...
+    /// length of time configured in <see cref="GraphQLWebSocketOptions.KeepAliveTimeout"/>.
+    /// </summary>
+    protected virtual async Task OnConnectionInitAsync(OperationMessage message, bool smartKeepAlive)
+    {
+        if (!await AuthorizeAsync(message))
+        {
+            return;
+        }
+        await OnConnectionAcknowledgeAsync(message);
+        TryInitialize();
```
sungam3r, line 274:
```suggestion
        _ = TryInitialize();
```

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,502 @@
...
+        await SendDataAsync(message.Id!, result);
+        await SendCompletedAsync(message.Id!);
+    }
+
+    /// <summary>
+    /// Sends an execution error to the client during set-up of a GraphQL request (typically subscription).
+    /// The default implementation calls <see cref="SendErrorResultAsync(string, ExecutionError)"/>.
+    /// </summary>
+    protected virtual Task SendErrorResultAsync(OperationMessage message, ExecutionError executionError)
+        => SendErrorResultAsync(message.Id!, executionError);
```
sungam3r, line 444:
```suggestion
    /// <summary>
    /// Sends an execution error to the client during set-up of a GraphQL request (typically subscription).
    /// The default implementation calls <see cref="SendErrorResultAsync(string, ExecutionResult)"/>.
    /// </summary>
    protected virtual Task SendErrorResultAsync(OperationMessage message, ExecutionError executionError)
        => SendErrorResultAsync(message.Id!, new ExecutionResult { Errors = new ExecutionErrors { executionError } });
```

To simplify things.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,502 @@
...
+    /// </summary>
+    protected abstract Task<ExecutionResult> ExecuteRequestAsync(OperationMessage message);
+
+    /// <summary>
+    /// Unsubscribes from a subscription event stream.
+    /// </summary>
+    protected virtual Task UnsubscribeAsync(string? id)
+    {
+        if (id != null)
+            Subscriptions.TryRemove(id);
```
sungam3r, line 488:
```suggestion
            _ = Subscriptions.TryRemove(id);
```

src/Transports.AspNetCore/WebSockets/SubscriptionsTransportWs/SubscriptionServer.cs
```diff
@@ -0,0 +1,250 @@
...
+    /// </summary>
+    protected IUserContextBuilder UserContextBuilder { get; }
+
+    /// <summary>
+    /// Returns the <see cref="IGraphQLSerializer"/> used to deserialize <see cref="OperationMessage"/> payloads.
+    /// </summary>
+    protected IGraphQLSerializer Serializer { get; }
+
+    /// <summary>
+    /// Initailizes a new instance with the specified parameters.
```
sungam3r, line 49:
```suggestion
    /// Initializes a new instance with the specified parameters.
```

src/Transports.AspNetCore/WebSockets/SubscriptionsTransportWs/SubscriptionServer.cs
```diff
@@ -0,0 +1,250 @@
...
+    /// <param name="sendStream">The WebSockets stream used to send data packets or close the connection.</param>
+    /// <param name="options">Configuration options for this instance.</param>
+    /// <param name="authorizationOptions">Authorization options for this instance.</param>
+    /// <param name="executer">The <see cref="IDocumentExecuter"/> to use to execute GraphQL requests.</param>
+    /// <param name="serializer">The <see cref="IGraphQLSerializer"/> to use to deserialize payloads stored within <see cref="OperationMessage.Payload"/>.</param>
+    /// <param name="serviceScopeFactory">A <see cref="IServiceScopeFactory"/> to create service scopes for execution of GraphQL requests.</param>
+    /// <param name="userContextBuilder">The user context builder used during connection initialization.</param>
+    /// <param name="authenticationService">An optional service to authenticate connections.</param>
+    public SubscriptionServer(
+        IWebSocketConnection sendStream,
```
sungam3r, line 60:
```suggestion
        IWebSocketConnection connection,
```

src/Transports.AspNetCore/WebSockets/SubscriptionsTransportWs/SubscriptionServer.cs
```diff
@@ -0,0 +1,250 @@
...
+        GraphQLWebSocketOptions options,
+        IAuthorizationOptions authorizationOptions,
+        IDocumentExecuter executer,
+        IGraphQLSerializer serializer,
+        IServiceScopeFactory serviceScopeFactory,
+        IUserContextBuilder userContextBuilder,
+        IWebSocketAuthenticationService? authenticationService = null)
+        : base(sendStream, options, authorizationOptions)
```
sungam3r, line 68:
```suggestion
        : base(connection, options, authorizationOptions)
```

src/Transports.AspNetCore/WebSockets/SubscriptionsTransportWs/SubscriptionServer.cs
```diff
@@ -0,0 +1,250 @@
...
+
+    /// <summary>
+    /// Returns the <see cref="IGraphQLSerializer"/> used to deserialize <see cref="OperationMessage"/> payloads.
+    /// </summary>
+    protected IGraphQLSerializer Serializer { get; }
+
+    /// <summary>
+    /// Initailizes a new instance with the specified parameters.
+    /// </summary>
+    /// <param name="sendStream">The WebSockets stream used to send data packets or close the connection.</param>
```
sungam3r, line 51:
```suggestion
    /// <param name="connection">The WebSockets stream used to send data packets or close the connection.</param>
```

src/Transports.AspNetCore/WebSockets/SubscriptionsTransportWs/SubscriptionServer.cs
```diff
@@ -0,0 +1,250 @@
...
+        }
+        else if (message.Type == MessageType.GQL_CONNECTION_INIT)
+        {
+            if (Initialized)
+            {
+                await ErrorTooManyInitializationRequestsAsync(message);
+            }
+            else
+            {
+                await OnConnectionInitAsync(message, false);
```
sungam3r, line 93: 🤔 What if 2 `init` requests come concurrently? At least I suggest to test `TryInitialize` result in `OnConnectionInitAsync` to not start timers twice.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -0,0 +1,502 @@
...
+
+    /// <summary>
+    /// Sets the initialized flag if it has not already been set.
+    /// Returns <see langword="false"/> if it was already set.
+    /// </summary>
+    protected bool TryInitialize()
+        => Interlocked.Exchange(ref _initialized, 1) == 0;
+
+    /// <summary>
+    /// Executes when a message has been received from the client.
```
sungam3r, line 134:
```suggestion
    /// Executes when a message has been received from the client.
    /// This method is essentially implements a state machine of used WebSockets sub-protocol.
```

I may be not precise with wordings.

src/Transports.AspNetCore/WebSockets/SubscriptionsTransportWs/MessageType.cs
```diff
@@ -0,0 +1,92 @@
...
+    /// </summary>
+    public const string GQL_CONNECTION_ERROR = "connection_error"; // Server -> Client
+
+    /// <summary>
+    ///     Server message that should be sent right after each GQL_CONNECTION_ACK processed and then periodically to keep the
+    ///     client connection alive.
+    ///     The client starts to consider the keep alive message only upon the first received keep alive message from the
+    ///     server.
+    ///     <remarks>
+    ///         NOTE: This one here don't follow the standard due to connection optimization
```
sungam3r, line 40: Is this still actual?

src/Transports.AspNetCore/WebSockets/SubscriptionsTransportWs/SubscriptionServer.cs
```diff
@@ -0,0 +1,250 @@
...
+                Id = id,
+                Type = MessageType.GQL_COMPLETE,
+            });
+        }
+    }
+
+    /// <inheritdoc/>
+    protected override async Task<ExecutionResult> ExecuteRequestAsync(OperationMessage message)
+    {
+        var request = Serializer.ReadNode<GraphQLRequest>(message.Payload)!;
```
sungam3r, line 182: Payload may be `null` here, better to check.

src/Transports.AspNetCore/WebSockets/SubscriptionsTransportWs/SubscriptionServer.cs
```diff
@@ -0,0 +1,250 @@
...
+    /// <br/><br/>
+    /// This method will return <see langword="true"/> if authorization is successful, or
+    /// return <see langword="false"/> if not.
+    /// </summary>
+    protected override async ValueTask<bool> AuthorizeAsync(OperationMessage message)
+    {
+        if (_authenticationService != null)
+            await _authenticationService.AuthenticateAsync(Connection, SubProtocol, message);
+
+        var success = await base.AuthorizeAsync(message);
```
sungam3r, line 241:
```suggestion
        bool success = await base.AuthorizeAsync(message);
```

src/Transports.AspNetCore/WebSockets/SubscriptionsTransportWs/SubscriptionServer.cs
```diff
@@ -0,0 +1,250 @@
...
+            Payload = "Access denied",
+        });
+        await base.ErrorAccessDeniedAsync();
+    }
+
+    /// <summary>
+    /// Authorizes an incoming GraphQL over WebSockets request with the connection initialization message and initializes the <see cref="UserContext"/>.
+    /// <br/><br/>
+    /// The default implementation calls the <see cref="IWebSocketAuthenticationService.AuthenticateAsync(IWebSocketConnection, string, OperationMessage)"/>
+    /// method to authenticate the request, checks the authorization rules set in <see cref="GraphQLHttpMiddlewareOptions"/>,
```
sungam3r, line 223:
```suggestion
    /// method to authenticate the request (if <see cref="IWebSocketAuthenticationService"/> was specified), checks the authorization rules set in <see cref="GraphQLHttpMiddlewareOptions"/>,
```

src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
```diff
@@ -0,0 +1,261 @@
...
+
+    /// <summary>
+    /// The WebSocket sub-protocol used for this protocol.
+    /// </summary>
+    /// <remarks>
+    /// Please note that the correct sub-protocol for the
+    /// <see href="https://github.com/enisdenjo/graphql-ws">graphql-ws</see>
+    /// protocol is <c>graphql-transport-ws</c> as is defined here.
+    /// </remarks>
+    public const string SubProtocol = "graphql-transport-ws";
```
sungam3r, line 21: Constants use `SUB_PROTOCOL` naming conventions.

src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
```diff
@@ -0,0 +1,261 @@
...
+    /// </summary>
+    protected IUserContextBuilder UserContextBuilder { get; }
+
+    /// <summary>
+    /// Returns the <see cref="IGraphQLSerializer"/> used to deserialize <see cref="OperationMessage"/> payloads.
+    /// </summary>
+    protected IGraphQLSerializer Serializer { get; }
+
+    /// <summary>
+    /// Initailizes a new instance with the specified parameters.
```
sungam3r, line 49:
```suggestion
    /// Initializes a new instance with the specified parameters.
```

src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
```diff
@@ -0,0 +1,261 @@
...
+    /// </summary>
+    /// <param name="sendStream">The WebSockets stream used to send data packets or close the connection.</param>
```
sungam3r, line 51:
```suggestion
    /// <param name="connection">The WebSockets stream used to send data packets or close the connection.</param>
```

src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
```diff
@@ -0,0 +1,261 @@
...
+    /// <param name="options">Configuration options for this instance.</param>
+    /// <param name="authorizationOptions">Authorization options for this instance.</param>
+    /// <param name="executer">The <see cref="IDocumentExecuter"/> to use to execute GraphQL requests.</param>
+    /// <param name="serializer">The <see cref="IGraphQLSerializer"/> to use to deserialize payloads stored within <see cref="OperationMessage.Payload"/>.</param>
+    /// <param name="serviceScopeFactory">A <see cref="IServiceScopeFactory"/> to create service scopes for execution of GraphQL requests.</param>
+    /// <param name="userContextBuilder">The user context builder used during connection initialization.</param>
+    /// <param name="authenticationService">An optional service to authenticate connections.</param>
+    public SubscriptionServer(
+        IWebSocketConnection sendStream,
```
sungam3r, line 60:
```suggestion
        IWebSocketConnection connection,
```

src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
```diff
@@ -0,0 +1,261 @@
...
+        GraphQLWebSocketOptions options,
+        IAuthorizationOptions authorizationOptions,
+        IDocumentExecuter executer,
+        IGraphQLSerializer serializer,
+        IServiceScopeFactory serviceScopeFactory,
+        IUserContextBuilder userContextBuilder,
+        IWebSocketAuthenticationService? authenticationService = null)
+        : base(sendStream, options, authorizationOptions)
```
sungam3r, line 68:
```suggestion
        : base(connection, options, authorizationOptions)
```

src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
```diff
@@ -0,0 +1,261 @@
...
+                break;
+            default:
+                await ErrorUnrecognizedMessageAsync(message);
+                break;
+        }
+    }
+
+    /// <summary>
+    /// GQL_PONG is a requrired response to a ping, and also a unidirectional keep-alive packet,
+    /// whereas GQL_PING is a bidirectional keep-alive packet.
```
sungam3r, line 123:
```suggestion
    /// Pong is a required response to a ping, and also a unidirectional keep-alive packet,
    /// whereas ping is a bidirectional keep-alive packet.
```

Only SubscriptionsTransportWs namespace has GQL prefixes.

src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
```diff
@@ -0,0 +1,261 @@
...
+            else
+                scope.Dispose();
+        }
+    }
+
+    /// <summary>
+    /// Authorizes an incoming GraphQL over WebSockets request with the connection initialization message and initializes the <see cref="UserContext"/>.
+    /// <br/><br/>
+    /// The default implementation calls the <see cref="IWebSocketAuthenticationService.AuthenticateAsync(IWebSocketConnection, string, OperationMessage)"/>
+    /// method to authenticate the request, checks the authorization rules set in <see cref="GraphQLHttpMiddlewareOptions"/>,
```
sungam3r, line 234:
```suggestion
    /// method to authenticate the request (if <see cref="IWebSocketAuthenticationService"/> was specified), checks the authorization rules set in <see cref="GraphQLHttpMiddlewareOptions"/>,
```

src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
```diff
@@ -0,0 +1,261 @@
...
+    /// <br/><br/>
+    /// This method will return <see langword="true"/> if authorization is successful, or
+    /// return <see langword="false"/> if not.
+    /// </summary>
+    protected override async ValueTask<bool> AuthorizeAsync(OperationMessage message)
+    {
+        if (_authenticationService != null)
+            await _authenticationService.AuthenticateAsync(Connection, SubProtocol, message);
+
+        var success = await base.AuthorizeAsync(message);
```
sungam3r, line 252:
```suggestion
        bool success = await base.AuthorizeAsync(message);
```

sungam3r:
изображение

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
Shane32: > volatile is not required for Interlocked API.

It also makes calls to `Thread.VolatileRead` unnecessary.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
Shane32: I disagree; it seems simpler to me the way it is now.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
Shane32: Added text:
> This method should essentially implement a state machine for the implemented protocol.

src/Transports.AspNetCore/WebSockets/SubscriptionsTransportWs/MessageType.cs
Shane32: I don't even understand what it says.  In any case, I'll just remove it.

src/Transports.AspNetCore/WebSockets/SubscriptionsTransportWs/SubscriptionServer.cs
Shane32: Added check which throws `ArgumentNullException` here, which is caught by `SubscribeAsync`, encapsulated into an `ExecutionError` by `HandleErrorDuringSubscribeAsync`, and then sent as a connection-attempt-failure and removed from the subscription list by `SendErrorResultAsync`.  This is similar to what would happen if a deserialization error were to occur.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
sungam3r: I suggested that all overloads call last abstract method without intermediate hops - plain delegating instead of nested.

src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
sungam3r: If so then this section in editorconfig should be changed/removed
```
########## name all constant fields using UPPER_CASE ##########
# https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-naming-conventions?view=vs-2019
# dotnet_naming_rule.<namingRuleTitle>.symbols = <symbolTitle>
dotnet_naming_rule.constant_fields_should_be_upper_case.symbols = constant_fields

# dotnet_naming_symbols.<symbolTitle>.<property> = <value>
dotnet_naming_symbols.constant_fields.applicable_kinds = field
dotnet_naming_symbols.constant_fields.applicable_accessibilities = *
dotnet_naming_symbols.constant_fields.required_modifiers = const

# dotnet_naming_rule.<namingRuleTitle>.style = <styleTitle>
dotnet_naming_rule.constant_fields_should_be_upper_case.style = upper_case_style

# dotnet_naming_style.<styleTitle>.<property> = <value>
dotnet_naming_style.upper_case_style.capitalization = all_upper
dotnet_naming_style.upper_case_style.word_separator = _

# dotnet_naming_rule.<namingRuleTitle>.severity = <value>
dotnet_naming_rule.constant_fields_should_be_upper_case.severity = warning
```

We used this convention for a long time.

Shane32: Yeah, I updated the editorconfig for my project to match the code, and have many other changes in there.

In any case, what do you think -- should we match .NET here or not?  Excluding the MessageType constants, the two `SubProtocol` constants are the only public constants.

src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
Shane32: Ok

src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
sungam3r: I'm used to see CONSTANT_CASE for constants. Such a notation allows you to distinguish constants from properties. In addition, when I added rule for such a notation, I was guided not by my preferences, but by some recommendations, which, however, I no longer remember.

sungam3r: So I would change all constants to CONST_CASE.

Shane32: What if we just make it a static property?

Shane32: (I dislike constant case.)

sungam3r: static property vs const is a eternal battle. Not a big deal to use static (+readonly) property/field instead of const.

Shane32: I'll do that then, since I don't want to use constant case.

sungam3r: const value is embedded within caller side, property - not

sungam3r (approved):
I think it's time to merge. It was a long way.

Shane32:
#812 is ready

sungam3r:
tomorrow morning

PR #811, opened 2022-05-27
Document that AddServer is legacy.

breyed:


Shane32:
I don't think `AddServer()` is valid at all anymore, is it?  And this library doesn't support prior versions of .NET so we should probably just remove that line in the readme.

breyed:
Makes sense to me. It may make onboarding easiest for most people if the readme used the new minimal hosting model in ASP.NET Core 6.0 as a baseline. Instructions for older frameworks could be included as annotations (perhaps in another section to keep the main onboarding lean).

sungam3r:
`AddServer` method was removed in v6 so I would remove line from readme.

sungam3r:
Fixed in #831

PR #812, opened 2022-05-28
Add WebSocketHandler implementation

Shane32:
This finishes all changes required for new WebSocket implementations.

Depends on:
- [x] #809 
- [x] #810 

Does not yet add tests, documentation, remove old implementation, etc.

### DIFF
--- a/src/Transports.AspNetCore/WebSockets/WebSocketHandler.cs
+++ b/src/Transports.AspNetCore/WebSockets/WebSocketHandler.cs
@@ -106,6 +106,10 @@ public class WebSocketHandler : IWebSocketHandler
         catch (OperationCanceledException) when (appStoppingToken.IsCancellationRequested)
         {
             // terminate all pending WebSockets connections when the application is in the process of stopping
+
+            // note: we are consuming OCE in this case because ASP.NET Core does not consider the task as canceled when
+            // an OCE occurs that is not due to httpContext.RequestAborted; so to prevent ASP.NET Core from considering
+            // this a "regular" exception, we consume it here
         }
     }
 
--- a/src/Transports.AspNetCore/WebSockets/WebSocketHandler.cs
+++ b/src/Transports.AspNetCore/WebSockets/WebSocketHandler.cs
@@ -99,7 +99,7 @@ public class WebSocketHandler : IWebSocketHandler
             return;
         try
         {
-            var webSocketConnection = CreateWebSocketConnection(httpContext, webSocket, cts.Token);
+            using var webSocketConnection = CreateWebSocketConnection(httpContext, webSocket, cts.Token);
             using var operationMessageReceiveStream = CreateReceiveStream(webSocketConnection, subProtocol, userContextBuilder);
             await webSocketConnection.ExecuteAsync(operationMessageReceiveStream);
         }
--- a/src/Transports.AspNetCore/WebSockets/WebSocketHandler.cs
+++ b/src/Transports.AspNetCore/WebSockets/WebSocketHandler.cs
@@ -1,0 +1,153 @@
+using System.Net.WebSockets;
+using GraphQL.Transport;
+using GraphQL.Types;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace GraphQL.Server.Transports.AspNetCore.WebSockets;
+
+/// <inheritdoc cref="WebSocketHandler"/>
+public class WebSocketHandler<TSchema> : WebSocketHandler, IWebSocketHandler<TSchema>
+    where TSchema : ISchema
+{
+    /// <inheritdoc cref="WebSocketHandler(IGraphQLSerializer, IDocumentExecuter, IServiceScopeFactory, GraphQLHttpMiddlewareOptions, IHostApplicationLifetime, IWebSocketAuthenticationService)"/>
+    public WebSocketHandler(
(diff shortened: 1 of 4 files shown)


sungam3r:
Will review some time later today or tomorrow, I reviewed all other your PRs here and in main repo.

sungam3r:
изображение
I accidentally saw this option. It was not there before. Maybe we will enable it on our repositories?

src/Transports.AspNetCore/WebSockets/WebSocketHandler.cs
```diff
@@ -0,0 +1,165 @@
...
+        IServiceScopeFactory serviceScopeFactory,
+        GraphQLWebSocketOptions options,
+        IAuthorizationOptions authorizationOptions,
+        IHostApplicationLifetime hostApplicationLifetime,
+        IWebSocketAuthenticationService? authorizationService)
+        : base(serializer, executer, serviceScopeFactory, options, authorizationOptions, hostApplicationLifetime, authorizationService)
+    {
+    }
+
+    /// <inheritdoc cref="WebSocketHandler(IGraphQLSerializer, IDocumentExecuter, IServiceScopeFactory, GraphQLWebSocketOptions, IAuthorizationOptions, IHostApplicationLifetime, IWebSocketAuthenticationService)"/>
```
sungam3r, line 27: no IWebSocketAuthenticationService here

src/Transports.AspNetCore/WebSockets/WebSocketHandler.cs
```diff
@@ -0,0 +1,165 @@
...
+            // an OCE occurs that is not due to httpContext.RequestAborted; so to prevent ASP.NET Core from considering
+            // this a "regular" exception, we consume it here
+        }
+    }
+
+    /// <summary>
+    /// Creates an <see cref="IWebSocketConnection"/>, a WebSocket message pump.
+    /// </summary>
+    protected virtual IWebSocketConnection CreateWebSocketConnection(HttpContext httpContext, WebSocket webSocket, CancellationToken cancellationToken)
+        => new WebSocketConnection(httpContext, webSocket, _serializer, Options, cancellationToken);
```
sungam3r, line 130: `_serializer` is not available for overrides without resaving it into field. Let's add `protected` property.

sungam3r: The same concern about overrides for `CreateReceiveStream` but of course one may just resave all needed arguments into fields.

src/Transports.AspNetCore/WebSockets/WebSocketHandler.cs
```diff
@@ -0,0 +1,165 @@
...
+
+    /// <summary>
+    /// Builds an <see cref="IOperationMessageProcessor"/> for the specified sub-protocol.
+    /// </summary>
+    protected virtual IOperationMessageProcessor CreateReceiveStream(IWebSocketConnection webSocketConnection, string subProtocol, IUserContextBuilder userContextBuilder)
```
sungam3r, line 135:
```suggestion
    protected virtual IOperationMessageProcessor CreateMessageProcessor(IWebSocketConnection webSocketConnection, string subProtocol, IUserContextBuilder userContextBuilder)
```

src/Transports.AspNetCore/WebSockets/WebSocketHandler.cs
```diff
@@ -0,0 +1,165 @@
...
+    {
+        if (subProtocol == GraphQLWs.SubscriptionServer.SubProtocol)
```
sungam3r, line 137: Why did you decide to make one WebSocketHandler and not two?

sungam3r: Honestly, it seems to me that it was possible to write all code without `WebSocketHandler` component at all. I think you want to provide more virtual methods here.

Shane32: Each handler is prioritized in the order of registration. For its priority, however, it may choose which of the available sub protocols to use. This is important and is why it needs to be one not two. 

The WebSocket handler is essentially a ASP.NET middleware and could be designed so. I wouldn’t do this, for a couple reasons, one of which is passing along the proper options class.  

We could roll the WebSocketHandler code directly into the main middleware and could add a few protected methods there that match the ones that exist here. 

As for adding more protected methods here, this class hardly contains any code at all as it is. What did you think I should be adding?

src/Transports.AspNetCore/WebSockets/WebSocketHandler.cs
Shane32: Ok, I can make protected properties for all fields... (??)

src/Transports.AspNetCore/WebSockets/WebSocketHandler.cs
Shane32: That is correct.  All intellisense comments are accurate, however.

src/Transports.AspNetCore/WebSockets/WebSocketHandler.cs
Shane32: Done

src/Transports.AspNetCore/WebSockets/WebSocketHandler.cs
Shane32: See #822

sungam3r:
#822 looks better

PR #814, opened 2022-06-05
Don't process incoming messages after requesting close

Shane32:
Mirrors https://github.com/Shane32/GraphQL.AspNetCore3/pull/37 which includes tests


sungam3r (approved):
Interesting change. I suggest to move ` _closeRequested = true;` before `Post` since `Post` _may_ be executed syncronously so it _may_ require some time to complete. During this time new messages can be received and dispatched in `while` loop.

Shane32:
> `Post` _may_ be executed syncronously

True; good catch.

FYI in my implementation, `CloseAsync` is only called from the the message handler, which is called from the message loop, and so multithreading as it pertains to this check isn't applicable.  But from a design perspective, `CloseAsync` _can_ be called at any time, so it is a good change.

PR #815, opened 2022-06-05
Split POST parsing into separate method

Shane32:
Allows for overriding `ReadPostContentAsync` to support additional form content types.  One specific example is to support attached files for form content types.  See sample at https://github.com/Shane32/GraphQL.AspNetCore3/blob/3caec91cf9fdce9d6a141423ace5144194c2a225/src/Tests/Middleware/FileUploadTests.cs#L59

Mirrors https://github.com/Shane32/GraphQL.AspNetCore3/pull/40

### DIFF
--- a/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
+++ b/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
@@ -304,7 +304,7 @@ public abstract class GraphQLHttpMiddleware
     /// <summary>
     /// Parses a request into a single or batch <see cref="GraphQLRequest"/> instance.
     /// <br/><br/>
-    /// In case of an error, this method will handle the request (e.g. by callling
+    /// In case of an error, this method will handle the request (e.g. by calling
     /// <see cref="WriteErrorResponseAsync(HttpContext, HttpStatusCode, ExecutionError)">WriteErrorResponseAsync</see>)
     /// and return <see langword="null"/>.
     /// </summary>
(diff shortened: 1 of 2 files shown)


src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -681,7 +698,6 @@ private static async Task<GraphQLRequest> DeserializeFromGraphBodyAsync(Stream b
         return new GraphQLRequest { Query = query };
     }
 
-#if NET5_0_OR_GREATER
```
sungam3r, line 684: ?

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -353,6 +301,75 @@ public virtual async Task InvokeAsync(HttpContext context)
         }
     }
 
+    /// <summary>
+    /// Parses a request into a single or batch <see cref="GraphQLRequest"/> instance.
+    /// <br/><br/>
+    /// In case of an error, this method will handle the request (e.g. by callling
```
sungam3r, line 307:
```suggestion
    /// In case of an error, this method will handle the request (e.g. by calling
```

sungam3r (approved):

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
Shane32: Charset is now decoded for any TFM, and used when reading media type application/graphql for any TFM. Since no transcoding stream available, it does not yet affect json reading on older TFMs

sungam3r: ok

PR #822, opened 2022-06-13
Add WebSocketHandler code in main middleware

Shane32:
Alternative design to #812 

Todo: create PR in GraphQL.AspNetCore3 to match with updated tests to ensure everything works correctly


src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -157,18 +109,50 @@ protected override async Task<ExecutionResult> ExecuteRequestAsync(HttpContext c
...
 
 /// <summary>
 /// ASP.NET Core middleware for processing GraphQL requests. Handles both single and batch requests,
 /// and dispatches WebSocket requests to the registered <see cref="IWebSocketHandler"/>.
 /// </summary>
-public abstract class GraphQLHttpMiddleware
+public abstract class GraphQLHttpMiddleware : IUserContextBuilder
 {
     private readonly IGraphQLTextSerializer _serializer;
-    private readonly IEnumerable<IWebSocketHandler>? _webSocketHandlers;
```
sungam3r, line 169: I see warnings about `IWebSocketHandler` in xml comments.

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -63,24 +43,8 @@ public class GraphQLHttpMiddleware<TSchema> : GraphQLHttpMiddleware
...
-    /// </summary>
-    protected GraphQLHttpMiddleware(
-        RequestDelegate next,
-        IGraphQLTextSerializer serializer,
-        IDocumentExecuter<TSchema> documentExecuter,
-        IServiceScopeFactory serviceScopeFactory,
-        GraphQLHttpMiddlewareOptions options,
-        IEnumerable<IWebSocketHandler<TSchema>>? webSocketHandlers = null)
-        : base(next, serializer, options, webSocketHandlers)
+        : base(next, serializer, options, hostApplicationLifetime)
```
sungam3r, line 47: Technically `hostApplicationLifetime` _may_ be optional. At least this is nothing more than an additional source of information about cancellation.

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
Shane32: Fixed

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
sungam3r: Although I forget (again) that the project is called Transports.**AspNetCore** so there is always should be `hostApplicationLifetime`.

PR #823, opened 2022-06-14
Simplify middleware

Shane32:
@sungam3r What do you think about moving the methods implemented within `GraphQLHttpMiddleware<TSchema>` into `GraphQLHttpMiddleware` ?


sungam3r:
I wanted to offer this yesterday, but I forgot. I would go further and remove typed middleware and `IDocumentExecuter<TSchema>` interface or a least think about it since in most cases there is single schema in app. `IDocumentExecuter<TSchema>` is just a way to set appropriate schema in execution options. It can be done with `ConfigureExecutionOptions` that you added later.

sungam3r:
Looking in more detail at the use of typed document executer, I came to the conclusion that it is easier to keep things as is.

sungam3r (approved):

PR #824, opened 2022-06-14
Remove old subscription code

Shane32:
Still need to add all the new tests, btw.  Just want to demonstrate that the samples server and whatever old tests that are still applicable work properly.


samples/Samples.Schemas.Chat/ChatSubscriptions.cs
```diff
@@ -35,8 +34,7 @@ public ChatSubscriptions(IChat chat)
 
     private IObservable<Message> SubscribeById(IResolveFieldContext context)
     {
-        var messageContext = (MessageHandlingContext)context.UserContext;
-        var user = messageContext.Get<ClaimsPrincipal>("user");
+        var user = (ClaimsPrincipal)context.UserContext["user"];
```
Shane32, line 37: Previously the subscription code would create its own UserContext.  Now the user context is whatever is defined by the user.

samples/Samples.Server/Startup.cs
```diff
@@ -49,8 +48,8 @@ public void ConfigureServices(IServiceCollection services)
             })
             .AddSystemTextJson()
             .AddErrorInfoProvider<CustomErrorInfoProvider>()
-            .AddWebSockets()
             .AddDataLoader()
+            .AddUserContextBuilder(context => new Dictionary<string, object> { { "user", context.User.Identity.IsAuthenticated ? context.User : null } })
```
Shane32, line 52: So for this sample we actually need a user context builder where none was defined previously.

samples/Samples.Server/Startup.cs
```diff
@@ -62,7 +61,6 @@ public void Configure(IApplicationBuilder app)
 
         app.UseWebSockets();
 
-        app.UseGraphQLWebSockets<ChatSchema>();
         app.UseGraphQL<GraphQLHttpMiddlewareWithLogs<ChatSchema>>("/graphql", new GraphQLHttpMiddlewareOptions());
```
Shane32, line 64: Note that it would be much easier to use ConfigureExecution to add logging rather than to derive from GraphQLHttpMiddleware.

tests/Transports.Subscriptions.WebSockets.Tests/WebSocketsConnectionFacts.cs
```diff
@@ -46,14 +46,8 @@ public async Task Should_accept_websocket_connection()
...
-        /* Given */
-        /* When */
-        var socket = await ConnectAsync("do-not-accept");
-        var segment = new ArraySegment<byte>(new byte[1024]);
-        var received = await socket.ReceiveAsync(segment, CancellationToken.None);
-
-        /* Then */
-        received.CloseStatus.ShouldBe(WebSocketCloseStatus.ProtocolError);
+        (await Should.ThrowAsync<InvalidOperationException>(async () => await ConnectAsync("do-not-accept")))
+            .Message.ShouldBe("Incomplete handshake, status code: 400");
```
Shane32, line 50: Previously the code would accept a websocket connection for an invalid protocol, and immediately close it with an error message.  Now it is refused at the HTTP level, returning a 400 bad result status code (with a proper message, which unfortunately is not visible through the C# WebSocket client class).

Shane32:
I did not see any end-to-end websocket tests (e.g. a client opening a websocket connection through a default server config, requesting a subscription, posting some test messages, and closing the connection).  Once posted in a PR, my tests will include end-to-end tests for subscriptions -- for both protocols.

samples/Samples.Schemas.Chat/ChatSubscriptions.cs
Shane32: Note that previously the ClaimsPrincipal would likely always be null from a browser, because it is impossible for a browser to send the HTTP Authorize header.  I'm not even sure if it can send cookies for cookie authentication.

sungam3r:
I was waiting for this PR to take a last look at the code that I never bothering to study 😄 .

samples/Samples.Server/Startup.cs
sungam3r: I agree if you mean general logging about request execution but middleware provides many virtual methods to override. I propose to override more of them and add logging there to demonstrate the main idea.

samples/Samples.Schemas.Chat/Samples.Schemas.Chat.csproj
```diff
@@ -9,7 +9,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\..\src\Transports.Subscriptions.Abstractions\Transports.Subscriptions.Abstractions.csproj" />
+    <PackageReference Include="GraphQL" Version="$(GraphQLVersion)" />
+    <PackageReference Include="System.Reactive" Version="5.0.0" />
   </ItemGroup>
```
sungam3r, line 14:
```suggestion
  </ItemGroup>

```

sungam3r (approved):
Great heartfelt thanks for the work done. I myself did not have enough strength, time, knowledge and desire to go through it.

Shane32:
> Great heartfelt thanks for the work done. I myself did not have enough strength, time, knowledge and desire to go through it.

No problem!  We are not done yet though. Need to get the authorization rule merged, then tests, samples, readme / documentation, and migration document.

PR #829, opened 2022-06-17
Update LICENSE.md

pekkah:
As I haven't done any active work on the codebase for ages I thought I mark the end there.


Shane32 (approved):
@sungam3r to approve also

sungam3r (approved):
A little sad when I hear something like this, although nothing bad happens.

PR #836, opened 2022-06-22
Restore old auth package with compatibility code

Shane32:
With this change, there would be zero breaking changes to the public API for the authorization package.  However, all new features such as `@skip` and `@include` support and `AllowAnonymous` support would function correctly.

Below is the only changes to the API compared to v6:

image

I would consider releasing this package along with the rest of v7, but immediately marking it as deprecated via nuget and do not release any past 7.0.0 -- and/or, adding the `[Obsolete]` tag to each class.

I think we should consider it because of the significantly different architecture of v7 authentication vs v6 authentication.  While v7 authentication should be a seamless transition for most people, anyone who has customized v6 authentication will need to rewrite their logic to work with v7.

Another reason to consider this is because the v6 authentication package will be incompatible with GraphQL.NET v7, due to `INodeVisitor` being having async methods in v7.


Shane32:
With this change, there would be zero breaking changes to the public API for the authorization package.  However, all new features such as skip and include support would function correctly.

image

Shane32:
If we decide not to merge this PR, the code may be useful if a user needs it.  I don't need it, but there are 400k downloads of the old nuget package, some users of which may appreciate the additional time to transition.

sungam3r:
I'm 50/50 here. For myself, I already feel the transition to a new authorization code as a completed stage. I reason so that it is better to take a step forward here and forget about the old code than to try to support it further and melt the strength on it. Merge/publish/deprecate/whatever you see fit.

Shane32:
I added `[Obsolete]` tags and added the old tests.  The only test that doesn't pass is the check for a policy set on an input type.  This is a known limitation of the new codebase, and we need to be sure it is prominently displayed in the migration notes.

Shane32:
I feel strongly that we should merge this PR.  This will help make the transition easier for users, and alert us if someone needs the error message builder or claims accessor without immediately alienating those users.  All of the "added" code is in a separate nuget package that is not otherwise necessary.

sungam3r (approved):

PR #838, opened 2022-06-23
Add tests for Transports.AspNetCore, increasing coverage to 99%

Shane32:
## Cleanup build props
- Apply NRT to entire solution
- Remove unneeded NoWarn entries
- Set all warnings as errors
- Internals visible to test project

## Tests
- Remove some old test projects
- Replace chat schema (since my tests heavily depend on it)
- Add new test project

## Coverage

| Module                               | Line   | Branch | Method |
|--------------------------------------|--------|--------|--------|
| GraphQL.Server.Samples.Schemas.Chat  | 91.25% | 90%    | 96.42% |
| GraphQL.Server.Transports.AspNetCore | 97.81% | 92.94% | 99.64% |
| GraphQL.Server.Ui.Altair             | 23.43% | 16.66% | 50%    |
| GraphQL.Server.Ui.GraphiQL           | 25%    | 14.28% | 55.55% |
| GraphQL.Server.Ui.Playground         | 28.57% | 14.7%  | 50%    |
| GraphQL.Server.Ui.Voyager            | 22.58% | 16.66% | 46.66% |


Shane32:
@sungam3r As promised, here are the tests to accompany the code that I've submitted over the last month or two, with coverage around 98%.  It's a lot of code; over 7,500 lines.  Likely as not you will want to review every file.  If not, I would be happy to direct your attention to the files that I think you would be most interested in reviewing:

- All csproj files, props and targets files
- GitHub workflow yml file
- `Samples.Schema.Chat` files (you'd probably want to compare the new sample schema to the old sample schema)
- New `ExecutionResultActionResult.cs` file
- Api approvals
- `AuthorizationTests.cs` (the one in the Middleware folder)
- `AuthorizationTests.cs` (the other one, for the validation rule)

I have made heavy use of mocks to test the websocket implementation code.  Yet, there is full end-to-end tests in the following places, which I suggest you review:
- `GetTests.cs`
- `PostTests.cs`
- `BatchTests.cs`
- `ChatTests.cs` (which covers WebSocket implementations)

In addition to the mundane tests to cover the extension methods, websocket implementation details, and so forth, which mostly rely on mocking, I've also included CORS tests, to be sure the ASP.Net Core CORS implementation plays nicely with this library, and a FileUpload test, simulating how one can derive from `GraphQLHttpMiddleware` to allow a multipart form upload submission to include files that are injected as variables to the GraphQL engine (versus a base64 encoding or similar).

I did not write code to test the UI packages, but I left the old Samples.Server.Tests code project, which is exercising some of the UI packages' code.

tests/Transports.AspNetCore.Tests/ChatTests.cs
```diff
@@ -0,0 +1,445 @@
...
+        {
+            Type = subProtocol == "graphql-ws" ? "start" : "subscribe",
+            Id = "123",
+            Payload = new GraphQLRequest
+            {
+                Query = "subscription { events { type message { id message from } } }",
+            },
+        });
+
+        await Task.Delay(1000);
```
Shane32, line 170: I would like to remove this delay, but the test then typically fails.  I believe this is because the prior call to `SendMessageAsync` gets processed more slowly than the following message via `AddMessageInternal`.  The latter call runs on a separate thread over a separate HttpClient, so there is no synchronization that ensures that the prior subscription request has completed before posting the new message.  Since the subscription server does not send any response, I cannot simply wait for a response here.  My only thought (which I have not tried) is that perhaps if some subsequent message that requires a response was sent via the WebSocket connection (e.g. CountMessages), then we could await that response before sending the add-message via a separate HttpClient.  Since there is no asynchronous code present in the chat service, this should eliminate the need for Task.Delay.  Still this is not a guaranteed fix because the BaseSubscriptionHandler is designed to process requests over the same WebSocket connection asynchronously. (See #821 )

Shane32: This same issue (necessary call to Task.Delay) is present elsewhere in this file and also in `TestServerExtensions.cs`

samples/Samples.Schemas.Chat/Schema/ChatSchema.cs
```diff
@@ -0,0 +1,11 @@
+namespace GraphQL.Samples.Schemas.Chat;
+
+public class ChatSchema : Schema
```
Shane32, line 3: My tests do not use this file, but rather uses `AddAutoSchema` to configure the schema; but the old `Samples.Server.Tests` project does.  I can probably eliminate this file and update the `Samples.Server.Tests` project if you like.  This is the only "code-first" file here; everything else is "type-first", making the sample very easy to read.

Shane32:
@sungam3r I have removed the added feature and moved it to a separate PR.  There is only one line of executable code added - a simple null check.  As this PR simply comprises of tests, I will merge this PR if you do not have time to review it.  Please let me know if you plan to review it soon.

sungam3r:
I'll review all pending PRs from you in coming weekend.

samples/Samples.Schemas.Chat/Schema/ChatSchema.cs
sungam3r: > I can probably eliminate this file and update the Samples.Server.Tests project if you like.  ... making the sample very easy to read

50/50 here,

Directory.Build.targets
```diff
@@ -4,7 +4,15 @@
     <CoverletOutputFormat>opencover</CoverletOutputFormat>
     <CoverletOutput>$(MSBuildThisFileDirectory).coverage/$(AssemblyName)/</CoverletOutput>
     <Exclude>[GraphQL.Samples.Schemas.Chat]*,[GraphQL.Samples.Server]*</Exclude>
-    <NoWarn>$(NoWarn);1591</NoWarn>
```
sungam3r, line 7: 👍

Directory.Build.targets
```diff
@@ -4,7 +4,15 @@
...
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(IsPackable)' != 'true'">
+    <NoWarn>$(NoWarn);CS1591;IDE1006;IDE0053;CS1998</NoWarn>
```
sungam3r, line 10: CS1998 ?

Directory.Build.props
```diff
@@ -19,8 +19,10 @@
     <DebugType>embedded</DebugType>
     <ImplicitUsings>enable</ImplicitUsings>
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <EnableNETAnalyzers>true</EnableNETAnalyzers>
-    <PackageReadmeFile Condition="'$(IsPackable)' == 'true'">README.md</PackageReadmeFile>
+    <Nullable>enable</Nullable>
+    <NoWarn>$(NoWarn);IDE0056;IDE0057</NoWarn>
```
sungam3r, line 25:
```suggestion
    <NoWarn>$(NoWarn);IDE0056;IDE0057</NoWarn> <!-- Index/Range operators -->
```

samples/Samples.Schemas.Chat/Samples.Schemas.Chat.csproj
```diff
@@ -1,16 +1,17 @@
...
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <WarningsNotAsErrors>CS8618</WarningsNotAsErrors>
     <IsPackable>false</IsPackable>
-    <NoWarn>$(NoWarn);1591</NoWarn>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="GraphQL" Version="$(GraphQLVersion)" />
     <PackageReference Include="System.Reactive" Version="5.0.0" />
+    <Using Include="GraphQL"/>
```
sungam3r, line 12: I would move implicit usings into separate group.

samples/Samples.Schemas.Chat/Samples.Schemas.Chat.csproj
```diff
@@ -1,16 +1,17 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
-    <AssemblyName>GraphQL.Samples.Schemas.Chat</AssemblyName>
-    <RootNamespace>GraphQL.Samples.Schemas.Chat</RootNamespace>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
```
sungam3r, line 4: TargetFramework**S** ?

.github/workflows/test.yml
```diff
@@ -63,7 +63,7 @@ jobs:
         if: ${{ startsWith(matrix.os, 'ubuntu') }}
         uses: codecov/codecov-action@v3
         with:
-          files: .coverage/GraphQL.Server.Authorization.AspNetCore.Tests/coverage.net5.opencover.xml,.coverage/GraphQL.Server.Samples.Server.Tests/coverage.net5.opencover.xml,.coverage/GraphQL.Server.Transports.Subscriptions.Abstractions.Tests/coverage.net5.opencover.xml,.coverage/GraphQL.Server.Transports.Subscriptions.WebSockets.Tests/coverage.net5.opencover.xml,.coverage/GraphQL.Server.Transports.AspNetCore.Tests/coverage.net5.opencover.xml
+          files: .coverage/GraphQL.Server.Transports.AspNetCore.Tests/coverage.net6.opencover.xml,.coverage/GraphQL.Server.Transports.AspNetCore.Tests/coverage.netcoreapp2.1.opencover.xml,.coverage/GraphQL.Server.Samples.Server.Tests/coverage.net6.opencover.xml
```
sungam3r, line 66: изображение
Looks odd.

samples/Samples.Server/CustomErrorInfoProvider.cs
```diff
@@ -26,7 +26,7 @@ private string GetAuthorizationErrorMessage(AccessDeniedError error)
         var errorMessage = new StringBuilder();
         errorMessage.Append(error.Message);
 
-        foreach (var failedRequirement in error.PolicyAuthorizationResult.Failure.FailedRequirements)
+        foreach (var failedRequirement in error.PolicyAuthorizationResult!.Failure!.FailedRequirements)
```
sungam3r, line 29: Why `!` operator?

samples/Samples.Server/StartupWithRouting.cs
```diff
@@ -47,7 +48,7 @@ public void ConfigureServices(IServiceCollection services)
             .AddSystemTextJson()
             .AddErrorInfoProvider<CustomErrorInfoProvider>()
             .AddDataLoader()
-            .AddUserContextBuilder(context => new Dictionary<string, object> { { "user", context.User.Identity.IsAuthenticated ? context.User : null } })
+            .AddUserContextBuilder(context => new Dictionary<string, object?> { { "user", context.User.Identity!.IsAuthenticated ? context.User : null } })
```
sungam3r, line 51: Also check README.md for API changes and other stuff.

src/Transports.AspNetCore/Assembly.cs
```diff
@@ -0,0 +1,3 @@
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("GraphQL.Server.Transports.AspNetCore.Tests")]
```
sungam3r, line 3: @Shane32 What I have learnt not so much time ago - https://www.meziantou.net/declaring-internalsvisibleto-in-the-csproj.htm  `InternalsVisibleTo` attribute became "native" since NET5.

tests/Samples.Server.Tests/Serializer.cs
```diff
@@ -51,7 +51,7 @@ internal static FormUrlEncodedContent ToFormUrlEncodedContent(GraphQLRequest req
             dictionary["extensions"] = ToJson(request.Extensions);
         }
 
-        return new FormUrlEncodedContent(dictionary);
+        return new FormUrlEncodedContent(dictionary.Cast<KeyValuePair<string?, string?>>());
```
sungam3r, line 54: ?

tests/Transports.AspNetCore.Tests/AuthorizationTests.cs
```diff
@@ -0,0 +1,728 @@
...
+        if (authenticated)
+            SetAuthorized();
+
+        _schema.RegisterType(interfaceGraphType);
+
+        var ret = Validate(testInterface ? "{ ... on TestInterface { test } }" : "{ test }");
+        ret.IsValid.ShouldBe(expectedIsValid);
+    }
+
+    // these tests will fail core rules validation and so it does not really matter
```
sungam3r, line 635:
```suggestion
    // these tests will fail core validation rules and so it does not really matter
```

samples/Samples.Server/CustomErrorInfoProvider.cs
Shane32: `error.PolicyAuthorizationResult` may be null if a role check failed rather than a policy check.  `Failure` is defined as nullable by .NET because `AuthorizationResult` may be a successful result -- but it will never be so here.

This code is old unchanged code.  I'll wrap the `switch` in an `if` to make a better sample.  `Failure!` is still proper.

tests/Transports.AspNetCore.Tests/ChatTests.cs
sungam3r: Anyway it  would be better to place a comment

samples/Samples.Server/CustomErrorInfoProvider.cs
sungam3r: ok

tests/Samples.Server.Tests/Serializer.cs
Shane32: It's perhaps a little stupid, but due to nullability differences, it won't compile as it was before.  Dictionaries can't have nullable keys, so I can't redefine the key of the dictionary to `string?`.

.github/workflows/test.yml
Shane32: Yeah i guess.... I didn't change that.

src/Transports.AspNetCore/Assembly.cs
Shane32: Are you suggesting we convert to csproj prop?

samples/Samples.Server/StartupWithRouting.cs
Shane32: I'm planning to rewrite the readme, subject to your review.

tests/Samples.Server.Tests/Serializer.cs
sungam3r: изображение
Seems an issue for old tfms.

src/Transports.AspNetCore/Assembly.cs
sungam3r: yes

Directory.Build.targets
Shane32: CS1998 won't let you write async methods if they don't contain await.  Which although makes sense for the main project, often causes a decrease in readability of tests. 

For example, when a delegate is `Func<ContextClass, Task<object>>` you can't write `async context => context.Widget`, but rather are required to write `context => Task.FromResult<object>(context.Widget)`.

So for tests CS1998 is disabled.

tests/Transports.AspNetCore.Tests/ChatTests.cs
Shane32: ok i'll add a comment along the lines of what's written above

tests/Transports.AspNetCore.Tests/Middleware/BatchTests.cs
```diff
@@ -0,0 +1,248 @@
...
+        _options.ValidationErrorsReturnBadRequest = badRequest;
+        using var response = await PostJsonAsync("[null,null]");
+        // validation errors do not return 400 within a batch request
+        await response.ShouldBeAsync(false, @"[{""errors"":[{""message"":""GraphQL query is missing."",""extensions"":{""code"":""QUERY_MISSING"",""codes"":[""QUERY_MISSING""]}}]},{""errors"":[{""message"":""GraphQL query is missing."",""extensions"":{""code"":""QUERY_MISSING"",""codes"":[""QUERY_MISSING""]}}]}]");
+    }
+
+    [Fact]
+    public async Task Mutation()
+    {
+        using var response = await PostBatchRequestAsync(new GraphQLRequest() { Query = "mutation{clearMessages}" });
```
sungam3r, line 173:
```suggestion
        using var response = await PostBatchRequestAsync(new GraphQLRequest { Query = "mutation{clearMessages}" });
```

and for other usages

tests/Samples.Server.Tests/Serializer.cs
Shane32: Probably.

tests/Transports.AspNetCore.Tests/Middleware/BatchTests.cs
Shane32: I agree and will accept suggestions where you make them along these lines.  I'm not searching the tests to change them.

tests/Transports.AspNetCore.Tests/ChatTests.cs
Shane32:
```suggestion
        // Delay is necessary to allow time for the asynchronous websocket handler code
        // to execute prior to the independent call to AddMessageInternal below.
        // Since the websocket call does not return a response when the subscription
        // has completed being set up, there is no response we can await to determine
        // when to call AddMessageInternal.
        await Task.Delay(1000);
```

Shane32: is this sufficient?

sungam3r: yes

src/Transports.AspNetCore/Assembly.cs
Shane32: Changed.

samples/Samples.Schemas.Chat/Schema/ChatSchema.cs
Shane32: I'll leave it as-is.

tests/Transports.AspNetCore.Tests/Middleware/MiscTests.cs
```diff
@@ -0,0 +1,104 @@
...
+        _ = new GraphQLHttpMiddleware<ISchema>(next, serializer, executer, scopeFactory, options, appLifetime);
+
+        Should.Throw<ArgumentNullException>(() => new GraphQLHttpMiddleware(null!, serializer, executer, scopeFactory, options, appLifetime));
+        Should.Throw<ArgumentNullException>(() => new GraphQLHttpMiddleware(next, null!, executer, scopeFactory, options, appLifetime));
+        Should.Throw<ArgumentNullException>(() => new GraphQLHttpMiddleware(next, serializer, null!, scopeFactory, options, appLifetime));
+        Should.Throw<ArgumentNullException>(() => new GraphQLHttpMiddleware(next, serializer, executer, null!, options, appLifetime));
+        Should.Throw<ArgumentNullException>(() => new GraphQLHttpMiddleware(next, serializer, executer, scopeFactory, null!, appLifetime));
+        Should.Throw<ArgumentNullException>(() => new GraphQLHttpMiddleware(next, serializer, executer, scopeFactory, options, null!));
+        _ = new GraphQLHttpMiddleware(next, serializer, executer, scopeFactory, options, appLifetime);
+
```
sungam3r, line 33:
```suggestion
```

tests/Transports.AspNetCore.Tests/TestServerExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+        response.EnsureSuccessStatusCode();
+        var str = await response.Content.ReadAsStringAsync();
+        return str;
+    }
+
+    public static async Task<string> ExecutePost(this TestServer server, string url, string query, object? variables = null)
+    {
+        var client = server.CreateClient();
+        var data = System.Text.Json.JsonSerializer.Serialize(new { query = query, variables = variables });
+        var content = new StringContent(data, Encoding.UTF8, "application/json");
```
sungam3r, line 18:
```suggestion
        var content = new StringContent(data, Encoding.UTF8, "application/graphql+json");
```
?

tests/Transports.AspNetCore.Tests/Transports.AspNetCore.Tests.csproj
```diff
@@ -1,19 +1,49 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <Import Project="../../Tests.props" />
 
-  <PropertyGroup>
-    <TargetFrameworks>net6;net5;netcoreapp3.1</TargetFrameworks>
+  <PropertyGroup Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Windows)))' != 'true'">
```
sungam3r, line 4: Let's solve this task by [Microsoft.NETFramework.ReferenceAssemblies](https://www.nuget.org/packages/Microsoft.NETFramework.ReferenceAssemblies.net461/) 

https://andrewlock.net/using-reference-assemblies-to-build-net-framework-libararies-on-linux-without-mono/

sungam3r (approved):
Tooooooo much for me for single read. At first I tried to check everything line by line, but very quickly I realized that it would take up a lot of time. I believe that you did everything to write good tests. 👍

tests/Transports.AspNetCore.Tests/Transports.AspNetCore.Tests.csproj
Shane32: Sounds good; I'll add an issue so I can investigate and test

Shane32: #845

PR #841, opened 2022-06-26
Include MicrosoftDI and SystemTextJson packages

Shane32:
These changes do not bring any dependencies that are not already required by ASP.NET Core:

### 1. ASP.NET Core already depends on Microsoft.Extensions.Options 2.0.0+:

For ASP.NET Core 3.1+ see:
- https://github.com/dotnet/aspnetcore/blob/main/eng/SharedFramework.External.props

For ASP.NET Core 2.1 see:
- https://www.nuget.org/packages/Microsoft.AspNetCore/2.1.7#dependencies-body-tab
- https://www.nuget.org/packages/Microsoft.AspNetCore.Hosting/2.1.1#dependencies-body-tab

### 2. .NET Core 3.1 already includes System.Text.Json

The reference to the `System.Text.Json` package for .NET Core 3.1 was removed in https://github.com/graphql-dotnet/graphql-dotnet/pull/3199

### Summary

Along with changing the extension methods to be within the GraphQL namespace, these changes simplify the required packages to use GraphQL.Server without creating any 3rd party dependencies.  Users may still utilize GraphQL.NewtonsoftJson if desired, and as far as I know, .NET Core will not load an assembly at runtime until referenced.  I cannot think of a good reason not to include these packages.

### Notes

For ASP.NET Core 2.1, users will still need to reference either the GraphQL.NewtonsoftJson or GraphQL.SystemTextJson package.  The Microsoft.AspNetCore.App package has an indirect dependency on Newtonsoft.Json 11.0.2+, as at the time it was the default serializer, but the GraphQL.NewtonsoftJson brings a newer version of Newtonsoft.Json so it is not included.

We may still include other GraphQL packages within the GraphQL.Server.All meta package as desired, such as GraphQL.DataLoader or GraphQL.MemoryCache.  The changes included here are typically required to use GraphQL.Server, and don't require external dependencies.

### DIFF
--- a/tests/Transports.AspNetCore.Tests/Transports.AspNetCore.Tests.csproj
+++ b/tests/Transports.AspNetCore.Tests/Transports.AspNetCore.Tests.csproj
@@ -8,8 +8,7 @@
   <ItemGroup>
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
     <PackageReference Include="Castle.Core" Version="5.0.0" />
-    <PackageReference Include="GraphQL.SystemTextJson" Version="5.1.1" />
-    <PackageReference Include="GraphQL.NewtonsoftJson" Version="5.1.1" />
+    <PackageReference Include="GraphQL.NewtonsoftJson" Version="$(GraphQLVersion)" />
   </ItemGroup>
 
   <ItemGroup>
--- a/tests/Transports.AspNetCore.Tests/Transports.AspNetCore.Tests.csproj
+++ b/tests/Transports.AspNetCore.Tests/Transports.AspNetCore.Tests.csproj
@@ -8,6 +8,7 @@
   <ItemGroup>
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
     <PackageReference Include="Castle.Core" Version="5.0.0" />
+    <PackageReference Include="GraphQL.SystemTextJson" Version="$(GraphQLVersion)" />
     <PackageReference Include="GraphQL.NewtonsoftJson" Version="$(GraphQLVersion)" />
   </ItemGroup>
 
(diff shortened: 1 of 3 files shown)


tests/Transports.AspNetCore.Tests/Transports.AspNetCore.Tests.csproj
```diff
@@ -8,8 +8,8 @@
   <ItemGroup>
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
     <PackageReference Include="Castle.Core" Version="5.0.0" />
-    <PackageReference Include="GraphQL.SystemTextJson" Version="5.1.1" />
-    <PackageReference Include="GraphQL.NewtonsoftJson" Version="5.1.1" />
+    <PackageReference Include="GraphQL.SystemTextJson" Version="$(GraphQLVersion)" />
```
sungam3r, line 11: I do not see any usages of GraphQLVersion in test.yml

sungam3r: Like in auth/relay project I mean.

Shane32: No, this is just the version we are referencing project-wide, and is called `GraphQLVersion`, unlike `GraphQLTestVersion` that is used by auth/relay to provide testing for multiple versions against a the compiled package.

We could certainly update the CI scripts to provide additional testing in a similar manner as the auth/relay packages do.

Shane32: I can update the CI scripts in a subsequent PR.

sungam3r (approved):

PR #843, opened 2022-06-29
Add ExecutionResultActionResult

Shane32:
As there are many users who still use or prefer to use MVC controller actions to execute a GraphQL request, this `IActionResult` implementation allows for proper serializing of the `ExecutionResult`.  As this project caters to ASP.Net Core, it seems appropriate to add it here.  It is very small with only 9 executable lines of code; the whole file is just 35 lines.

Tests will be added after #838 is merged

### DIFF
--- a/src/Transports.AspNetCore/ExecutionResultActionResult.cs
+++ b/src/Transports.AspNetCore/ExecutionResultActionResult.cs
@@ -1,0 +1,34 @@
+namespace GraphQL.Server.Transports.AspNetCore;
+
+/// <summary>
+/// An action result that formats the <see cref="ExecutionResult"/> as JSON.
+/// </summary>
+public sealed class ExecutionResultActionResult : IActionResult
+{
+    private readonly ExecutionResult _executionResult;
+    private readonly HttpStatusCode _statusCode;
+
+    /// <inheritdoc cref="ExecutionResultActionResult"/>
+    public ExecutionResultActionResult(ExecutionResult executionResult)
+    {
+        _executionResult = executionResult;
+        _statusCode = executionResult.Executed ? HttpStatusCode.OK : HttpStatusCode.BadRequest;
+    }
+
+    /// <inheritdoc cref="ExecutionResultActionResult"/>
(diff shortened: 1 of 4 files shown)


src/Transports.AspNetCore/ExecutionResultActionResult.cs
```diff
@@ -0,0 +1,34 @@
...
+        _executionResult = executionResult;
+        _statusCode = statusCode;
+    }
+
+    /// <inheritdoc/>
+    public async Task ExecuteResultAsync(ActionContext context)
+    {
+        var writer = context.HttpContext.RequestServices.GetRequiredService<IGraphQLSerializer>();
+        var response = context.HttpContext.Response;
+        response.ContentType = GraphQLHttpMiddleware.CONTENTTYPE_GRAPHQLJSON;
```
sungam3r, line 30: Since `SelectResponseContentType` is virtual I suggest to add get/set public property here with default value of `GraphQLHttpMiddleware.CONTENTTYPE_GRAPHQLJSON`.

src/Transports.AspNetCore/ExecutionResultActionResult.cs
```diff
@@ -0,0 +1,34 @@
...
+    public ExecutionResultActionResult(ExecutionResult executionResult, HttpStatusCode statusCode)
+    {
+        _executionResult = executionResult;
+        _statusCode = statusCode;
+    }
+
+    /// <inheritdoc/>
+    public async Task ExecuteResultAsync(ActionContext context)
+    {
+        var writer = context.HttpContext.RequestServices.GetRequiredService<IGraphQLSerializer>();
```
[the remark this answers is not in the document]
Shane32: Done

src/Transports.AspNetCore/ExecutionResultActionResult.cs
sungam3r: At least it is consistent with middleware code.

sungam3r (approved):

PR #851, opened 2022-07-06
Support .NET 4.8 for UI packages

Shane32:
Notes:
- ASP.NET 2.1 contains a reference to Newtonsoft.Json, while .NET Core 3.1 includes System.Text.Json.  So either way there is a serializer available without an additional reference.
- There are no ifdefs for the net5 and net6 TFMs, so they have been removed.


src/Ui.Altair/Internal/AltairPageModel.cs
```diff
@@ -37,11 +36,20 @@ public string Render()
...
 
             _altairCSHtml = _options.PostConfigure(_options, builder.ToString());
         }
 
         return _altairCSHtml;
     }
+
+    private static string JsonSerialize(object value)
+    {
+#if !NETSTANDARD2_0
```
sungam3r, line 49: I would invert condition.

sungam3r:
> There are no ifdefs for the net5 and net6 TFMs, so they have been removed.

I would prefer to have explicit tfms even without ifdefs.

sungam3r:
Nice and simple change.

src/Ui.Altair/Internal/AltairPageModel.cs
Shane32: Ok

Shane32:
> I'm a bit concerned about restoring 3.1 tfm from 5.0/6.0 apps.

Please define your concern and I will address it.  There is no difference here between targeting .NET Standard 2.0 or 2.1, except we have defined a minimum baseline of .NET Core 3.1.  Going forward there is no .NET Standard, which is there are ifdefs like NETCOREAPP3_1_OR_GREATER.  There is no need to target every newly released .NET Core version, and we do not do so in any of our other published packages.

src/Ui.Altair/Internal/AltairPageModel.cs
Shane32: Btw, you have requested this change many times, and I do not agree with the general principle that conditions should be inverted if they have a `!`.  In this instance the ifdef would have matched the other ifdef in the project, and would place the typical/default/primary code first.  I think you should reconsider your viewpoint.  Regardless, I changed it as you requested.

sungam3r: I reason from the simplicity of reading code when you do not mentally invert condition even if typical/default/primary code implies `!`.

sungam3r:
Something like phantom fears.

sungam3r (approved):

PR #855, opened 2022-07-10
Add underlying support for AWS AppSync subscription variant

Shane32:
For AWS, subscription operations differ from the `subscriptions-transport-ws` protocol in three ways:

1. Keep-alive messages are sent every minute by the server
2. The `connection_ack` message includes a keep-alive timeout period which the client should honor and terminate the connection if no keep-alive messages are received within this period
3. During a subscription request, a `start_ack` message is sent from the server to the client once the subscription has been set up without an error

The first two points above can be implemented easily in a derived class.  This PR adds a protected method for the third bullet point.  No functionality has changed, so the default behavior will meet the spec.  Adding this method could also assist with logging/debugging.

Note that as AWS AppSync uses the same WebSocket sub-protocol of `graphql-ws`, we cannot simply add a third protocol.  We could add an option to `GraphQLWebSocketOptions` for AWS AppSync compatibility, of course.

See https://docs.aws.amazon.com/appsync/latest/devguide/real-time-websocket-client.html

### DIFF
--- a/src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
+++ b/src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
@@ -471,6 +471,10 @@ public abstract partial class BaseSubscriptionServer : IOperationMessageProcesso
 
     /// <summary>
     /// Sends a notification that the specified subscription has been set up successfully.
+    /// Note that both graphql-ws and subscriptions-transport-ws protocols do not need
+    /// such notification but some custom protocol may require it; for example, see start_ack message
+    /// from https://docs.aws.amazon.com/appsync/latest/devguide/real-time-websocket-client.html
+    /// protocol for AWS AppSync. 
     /// </summary>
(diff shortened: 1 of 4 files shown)


src/Transports.AspNetCore/WebSockets/BaseSubscriptionServer.cs
```diff
@@ -468,6 +469,11 @@ protected virtual Task SendErrorResultAsync(OperationMessage message, ExecutionR
     /// </summary>
     protected abstract Task SendErrorResultAsync(string id, ExecutionResult result);
 
+    /// <summary>
+    /// Sends a notification that the specified subscription has been set up successfully.
```
sungam3r, line 473:
```suggestion
    /// Sends a notification that the specified subscription has been set up successfully.
    /// Note that both graphql-ws and subscriptions-transport-ws protocols do not need
    /// such notification but some custom protocol may require it, for example, see start_ack message
    /// from https://docs.aws.amazon.com/appsync/latest/devguide/real-time-websocket-client.html
    /// protocol for AWS AppSync. 
```

sungam3r (approved):

PR #857, opened 2022-07-10
Update target framework for All metapackage

Shane32:
- Allows All metapackage to be used for ASP.NET Core 2.1
- Serializer and MicrosoftDI is already included within Transports.AspNetCore

### DIFF
--- a/src/All/All.csproj
+++ b/src/All/All.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net6;net5;netcoreapp3.1</TargetFrameworks>
+    <TargetFramework>netstandard2.0</TargetFramework>
     <Description>GraphQL Server meta package with all the packages you need to get started</Description>
     <PackageTags>GraphQL;transport;server;WebSockets;subscriptions;middleware;authentication;authorization</PackageTags>
     <IncludeBuildOutput>false</IncludeBuildOutput>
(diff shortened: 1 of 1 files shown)


src/All/All.csproj
```diff
@@ -17,8 +17,6 @@
 
     <PackageReference Include="GraphQL.MemoryCache" Version="$(GraphQLVersion)" />
     <PackageReference Include="GraphQL.DataLoader" Version="$(GraphQLVersion)" />
-    <PackageReference Include="GraphQL.MicrosoftDI" Version="$(GraphQLVersion)" />
```
sungam3r, line 20: I am curious here. Create simple console net5/net6 app, then reference this metapackage. What version (I mean what TFM) of GraphQL.MemoryCache/GraphQL dlls will you see in output folder? netstandard2.0 of net5/net6 ?

Shane32: I created a console app only referencing the All meta-package.  I then verified that the DLLs are the appropriate version with a binary file comparison.  Specifically, GraphQL.SystemTextJson (pulled from nuget) is the net6 version (of version 6.0.0-preview-602) and Transports.AspNetCore (pulled from the local build) is the net5 version.  I did not check other files.

Shane32:
I did notice that Transports.AspNetCore does not include GraphQL.NewtonsoftJson because that would bump the Newtonsoft.Json version unnecessarily.  So I included it here for .NET Standard 2.0 projects (which includes .NET Core 2.1).

src/All/All.csproj
sungam3r: OK. So TFM resolution works "globally" and is not constrained by consuming package.

sungam3r (approved):

PR #854, opened 2022-07-09
Add v7 migration documentation

Shane32:
This document will be referenced by the README.md file in a subsequent PR.

PR #862, opened 2022-07-12
Change order of UI extension method arguments

Shane32:
I think the UI middleware extension methods should require the path first and then the options, rather than the other way around.  I believe this just makes logical sense that you would define the endpoint address and then configure it.  Plus it matches the argument ordering of the `UseGraphQL` method (both now and previously).

Also I generally find that it works best when optional method arguments are additive to existing/required arguments, and this change fixes that issue as well.

This will require a note in the migration doc ( #854 ) if it is merged.

Some other options:
- Leave the old methods, deprecated
- Deprecate the path-only syntax leaving only the (options, path) syntax
- Change the options argument to a builder delegate (if we do this then I would deprecate the old ones rather than removing them)

Since we have a large number of changes going into v7, perhaps now is a good time to make this change.

### DIFF
--- a/samples/Samples.Authorization/Program.cs
+++ b/samples/Samples.Authorization/Program.cs
@@ -62,13 +62,13 @@ app.UseWebSockets();
 app.UseGraphQL("/graphql");
 // configure Playground at "/ui/graphql"
 app.UseGraphQLPlayground(
+    "/ui/graphql",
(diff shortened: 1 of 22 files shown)


sungam3r (approved):
Yep, I was thinking about it.

PR #866, opened 2022-07-13
Pull ClaimsPrincipal from ValidationContext

Shane32:


src/Transports.AspNetCore/Extensions/GraphQLBuilderExtensions.cs
```diff
@@ -137,7 +137,6 @@ public static IGraphQLBuilder AddWebSocketAuthentication(this IGraphQLBuilder bu
     public static IGraphQLBuilder AddAuthorizationRule(this IGraphQLBuilder builder)
     {
         builder.AddValidationRule<AuthorizationValidationRule>(true);
-        builder.Services.TryRegister<IHttpContextAccessor, HttpContextAccessor>(DI.ServiceLifetime.Singleton);
```
Shane32, line 140: No more need to use `IHttpContextAccessor`

sungam3r: 👍

PR #870, opened 2022-07-14
Additional incoming message optimization

### LINKED ISSUE #813: Investigate a better way of using buffers to handle WebSocket messages
Do you have any questions here?  Yes it is complicated to read.  Yes it would certainly be more optimized with the MS package you previously referenced.  But everything should work.

Note that I do not know how likely it is for a message to come through as a single packet (and so retain in the 16kb buffer).  It is possible that messages often come through as multiple packets, especially if TCP breaks apart packets over 1500 bytes.  This could be further optimized by reusing the 16kb buffer until it was full before switching to the stream.  

It's also possible that the message comes through as a packet and then the end-of-message packet comes through separately (just as we send messages).  This would currently hit the stream instead of staying within the 16kb buffer (without further optimization).

_Originally posted by @Shane32 in https://github.com/graphql-dotnet/server/pull/809#discussion_r885764367_

Shane32:
Closes #813

### DIFF
--- a/src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
+++ b/src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
@@ -142,7 +142,7 @@ public class WebSocketConnection : IWebSocketConnection
                     }
                     else
                     {
-                        // if there is any data in this block, add it to the buffer
+                        // if there is any data in byte[] buffer, flush it into stream buffer
                         if (bufferOffset > 0)
                             receiveStream.Write(buffer, 0, bufferOffset);
                         bufferOffset = 0;
--- a/src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
+++ b/src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
@@ -81,17 +81,17 @@ public class WebSocketConnection : IWebSocketConnection
         try
         {
             await operationMessageProcessor.InitializeConnectionAsync();
-            // set up a buffer in case a message is longer than one block
+            // set up a buffer in case a message is longer than 16KB
             var receiveStream = new MemoryStream();
-            // set up a 16KB data block
+            // set up a 16KB block to accumulate data from WebSocket
             byte[] buffer = new byte[BUFFER_SIZE];
             int bufferOffset = 0;
-            // prep a reader stream
+            // prep a reader stream - just a view of the 16KB buffer for the deserializer so the data does not need to be copied to the memory stream
             var bufferStream = new ReusableMemoryReaderStream(buffer);
             // read messages until an exception occurs, the cancellation token is signaled, or a 'close' message is received
             while (true)
             {
-                // prep a Memory instance pointing to the block
+                // prep a Memory instance pointing to the free part of block
 #if NETSTANDARD2_0
                 var bufferMemory = new ArraySegment<byte>(buffer, bufferOffset, BUFFER_SIZE - bufferOffset);
 #else
@@ -161,7 +161,7 @@ public class WebSocketConnection : IWebSocketConnection
                     // if there is room in the buffer for more information, continue to fill the buffer
                     if (bufferOffset < BUFFER_SIZE)
                         continue;
-                    // if there is any data in this block, add it to the buffer
+                    // if there is any data in byte[] buffer, flush it into stream buffer
                     if (bufferOffset > 0)
                         receiveStream.Write(buffer, 0, bufferOffset);
                     bufferOffset = 0;
--- a/src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
+++ b/src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
@@ -26,6 +26,8 @@ public class WebSocketConnection : IWebSocketConnection
     private readonly TimeSpan _closeTimeout;
     private int _executed;
 
+    private const int BUFFER_SIZE = 16384;
+
     /// <inheritdoc/>
     public CancellationToken RequestAborted { get; }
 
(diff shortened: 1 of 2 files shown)


sungam3r:
I can't figure out where is optimization.

Shane32:
Since TCP packets are usually under 1500 bytes, it is likely that receiving data from a WebSocket will be under 1500 bytes. If the WebSocket packet was not marked as the end of the message, the old code would immediately start copying the data to a memory stream. 

The new code will continue to use the 16kb buffer until it is full before copying to a memory stream. This provides a very slight enhancement for GraphQL requests which are larger than 1500 bytes but smaller than 16kb.

The output buffer is already optimal, which is what is more important anyway.

Shane32:
The memory stream isn't reset per loop, so realistically this optimization is quite minimal.  Another option would be simply to reduce my buffer size, since anything larger than the MTU of the network is probably useless.  Note that TcpClient.ReceiveBufferSize returns a value of 8192.

https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.tcpclient.receivebuffersize?view=net-6.0

Shane32:
Note that this PR does not need to be merged before v7 is released.

sungam3r:
> If the WebSocket packet was not marked as the end of the message, the old code would immediately start copying the data to a memory stream.

So? Is it something bad? This stream is not newing on each loop iteration. Eventually message is deserialized either from `ReusableMemoryReaderStream` or `MemoryStream`. After looking at the code again, I don’t even catch the point of having two different streams. Or... I see _write_ operations only for MemoryStream. Code does not call writes for `ReusableMemoryReaderStream` at all.

Anyway I have a feeling that we solve the classical problem of streaming deserialization. Something that System.IO.Pipelines deals with.

Shane32:
> Is it something bad?

No the old code worked perfectly. I’m perfectly content to simply close this PR. 

> Code does not call writes for ReusableMemoryReaderStream at all

Correct. ReusableMemoryReaderStream is just a view of the 16kb buffer for the deserializer so the data does not need to be copied to the memory stream.

Shane32:
> System.IO.Pipelines 

I don’t want to introduce another dependency and do not see the need.

sungam3r:
I don't want either. So these changes save copy operations to MemoryStream, right? And the code in a larger number of cases gets into `ReadAsync` call on `ReusableMemoryReaderStream` instead of `MemoryStream` since `receiveStream.Length == 0` is true.

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
```diff
@@ -82,19 +84,21 @@ public virtual async Task ExecuteAsync(IOperationMessageProcessor operationMessa
...
-            byte[] buffer = new byte[16384];
-            // prep a Memory instance pointing to the block
-#if NETSTANDARD2_0
-            var bufferMemory = new ArraySegment<byte>(buffer);
-#else
-            var bufferMemory = new Memory<byte>(buffer);
-#endif
+            byte[] buffer = new byte[BUFFER_SIZE];
+            int bufferOffset = 0;
             // prep a reader stream
```
sungam3r, line 89:
```suggestion
            // prep a reader stream - just a view of the 16KB buffer for the deserializer so the data does not need to be copied to the memory stream
```

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
```diff
@@ -149,9 +158,13 @@ public virtual async Task ExecuteAsync(IOperationMessageProcessor operationMessa
                 }
                 else
                 {
+                    // if there is room in the buffer for more information, continue to fill the buffer
+                    if (bufferOffset < BUFFER_SIZE)
+                        continue;
                     // if there is any data in this block, add it to the buffer
```
sungam3r, line 164:
```suggestion
                    // if there is any data in buffer, flush it into stream
```

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
```diff
@@ -82,19 +84,21 @@ public virtual async Task ExecuteAsync(IOperationMessageProcessor operationMessa
             // set up a buffer in case a message is longer than one block
             var receiveStream = new MemoryStream();
             // set up a 16KB data block
```
sungam3r, line 86:
```suggestion
            // set up a 16KB block to accumulate data from WebSocket
```

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
```diff
@@ -82,19 +84,21 @@ public virtual async Task ExecuteAsync(IOperationMessageProcessor operationMessa
             // set up a buffer in case a message is longer than one block
```
sungam3r, line 84:
```suggestion
            // set up a buffer in case a message is longer than 16KB
```

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
```diff
@@ -82,19 +84,21 @@ public virtual async Task ExecuteAsync(IOperationMessageProcessor operationMessa
...
-            var bufferMemory = new Memory<byte>(buffer);
-#endif
+            byte[] buffer = new byte[BUFFER_SIZE];
+            int bufferOffset = 0;
             // prep a reader stream
             var bufferStream = new ReusableMemoryReaderStream(buffer);
             // read messages until an exception occurs, the cancellation token is signaled, or a 'close' message is received
             while (true)
             {
+                // prep a Memory instance pointing to the block
```
sungam3r, line 94:
```suggestion
                // prep a Memory instance pointing to the free part of block
```

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
```diff
@@ -115,28 +119,33 @@ public virtual async Task ExecuteAsync(IOperationMessageProcessor operationMessa
...
                 // if close has been requested, ignore incoming messages
                 if (_closeRequested)
+                {
+                    bufferOffset = 0;
                     continue;
+                }
                 // if this is the last block terminating a message
                 if (result.EndOfMessage)
                 {
                     // if only one block of data was sent for this message
```
sungam3r, line 129:
```suggestion
                    // if message was less then 16 KB
```

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
```diff
@@ -115,28 +119,33 @@ public virtual async Task ExecuteAsync(IOperationMessageProcessor operationMessa
...
+                        bufferStream.ResetLength(bufferOffset);
                         var message = await _serializer.ReadAsync<OperationMessage>(bufferStream, RequestAborted);
+                        bufferOffset = 0;
                         // dispatch the message
                         if (message != null)
                             await OnDispatchMessageAsync(operationMessageProcessor, message);
                     }
                     else
                     {
                         // if there is any data in this block, add it to the buffer
```
sungam3r, line 145:
```suggestion
                        // if there is any data in byte[] buffer, flush it into stream buffer
```

sungam3r (approved):

PR #871, opened 2022-07-15
Fix intermittent issue with tests

Shane32:
Having a required `Delay(...)` in a test never works in the long run.  I knew this could be an issue.  There shouldn't be any other tests that required a delay in order to pass.

On the bright side, tests will run faster with this patch.


Shane32:
Note that coverlet is still giving us problems:
- https://github.com/graphql-dotnet/server/runs/7359158293?check_suite_focus=true

Shane32:
Note that this PR does not need to be merged before v7 is released.

sungam3r (approved):

PR #873, opened 2022-07-15
Update NuGet package versioning

Shane32:


Shane32:
Since the UI packages don't depend on GraphQL.NET or GraphQL.NET Server, I'll remove the versioning changes from those from the All package

Shane32:
When merging to develop, I will update GraphQLVersion and NextVersion appropriately

sungam3r:
How does it help with https://github.com/graphql-dotnet/graphql-dotnet/issues/3229 exactly?

Shane32:
> How does it help with [graphql-dotnet/graphql-dotnet#3229](https://github.com/graphql-dotnet/graphql-dotnet/issues/3229) exactly?

Sorry, that's how https://github.com/graphql-dotnet/graphql-dotnet/pull/3231 would have prevented the issue.  This PR would prevent the issue in the future because the packages prevent dependencies on future (incompatible) versions of GraphQL.NET.

I believe only the authorization package is actually incompatible with GraphQL.NET 7.  We could only apply the changes to the Authorization and All nuget packages.  The UI packages are unaffected since they have no dependencies on GraphQL.NET or the rest of the Server packages.

Shane32:
@sungam3r Please review.  I would like to release 6.1.1 ahead of 7

Directory.Build.targets
```diff
@@ -6,4 +6,19 @@
     <Exclude>[GraphQL.Samples.Schemas.Chat]*,[GraphQL.Samples.Server]*</Exclude>
   </PropertyGroup>
 
+  <Target Name="UseExplicitPackageVersions" BeforeTargets="GenerateNuspec">
```
sungam3r, line 9:
```suggestion
  <!--Workaround for https://github.com/NuGet/Home/issues/5556 and https://github.com/NuGet/Home/issues/5525-->
  <Target Name="UseExplicitPackageVersions" BeforeTargets="GenerateNuspec">
```

sungam3r (approved):

PR #876, opened 2022-07-18
Change access denied status code to 403 forbidden

Shane32:
- 401 Unauthorized should be used when a request cannot be authenticated
- 403 Forbidden should be used when a request does not meet the policy requirements

See: https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

This change makes role/policy requirements return 403 when the policy or role list does not meet the requirements.  However, this also means for a 401 to be properly returned in such cases, `AuthorizationRequired` must also be set.

Also makes it so that setting a required policy or set of roles implies also that authentication is required.  Checking that the user is authenticated now occurs first, so that the response code for HTTP requests can be set appropriately to 401 if the user is not authenticated in the event that just a policy requirement was set, as this would be a more appropriate status code, rather than 403.

### DIFF
--- a/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
+++ b/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
@@ -292,9 +292,14 @@ public class GraphQLHttpMiddleware : IUserContextBuilder
         // Normal execution with single graphql request
         var userContext = await BuildUserContextAsync(context, null);
         var result = await ExecuteRequestAsync(context, gqlRequest, context.RequestServices, userContext);
-        var statusCode = _options.ValidationErrorsReturnBadRequest && !result.Executed
-            ? HttpStatusCode.BadRequest
-            : HttpStatusCode.OK;
+        HttpStatusCode statusCode = HttpStatusCode.OK;
+        if (!result.Executed)
+        {
+            if (result.Errors?.Any(e => e is HttpMethodValidationError) == true)
+                statusCode = HttpStatusCode.MethodNotAllowed;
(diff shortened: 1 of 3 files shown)


src/Transports.AspNetCore/IAuthorizationOptions.cs
```diff
@@ -19,6 +19,7 @@ public interface IAuthorizationOptions
     /// for the user within <see cref="HttpContext.User"/>
     /// for at least one role in the list prior to executing the GraphQL request or accepting
     /// the WebSocket connection.  If no roles are specified, authorization is not checked.
+    /// Also implies <see cref="AuthorizationRequired"/> is set to <see langword="true"/>.
```
sungam3r, line 22:
```suggestion
    /// Also implies <see cref="AuthorizationRequired"/> is enabled even if it is set to <see langword="false"/>.
```

It seems to me that such a wording is more understandable.

src/Transports.AspNetCore/IAuthorizationOptions.cs
```diff
@@ -27,6 +28,7 @@ public interface IAuthorizationOptions
     /// return a successful result for the user within <see cref="HttpContext.User"/>
     /// for the specified policy before executing the GraphQL
     /// request or accepting the WebSocket connection.
+    /// Also implies <see cref="AuthorizationRequired"/> is set to <see langword="true"/>.
```
sungam3r, line 31:
```suggestion
    /// Also implies <see cref="AuthorizationRequired"/> is enabled even if it is set to <see langword="false"/>.
```

sungam3r (approved):

PR #877, opened 2022-07-18
Return 405 Method Not Allowed for invalid request types

### LINKED ISSUE #794: Immediately respond with 405 instead of 400
I agree but the parsing has not occurred prior. You are welcome to create a PR that works better. For now this is an improvement over nothing, which is what happens now.

_Originally posted by @Shane32 in https://github.com/graphql-dotnet/server/pull/774#discussion_r877585267_

Shane32:
Fixes #794


sungam3r (approved):
Perfectly

PR #879, opened 2022-07-22
Add GraphQLHttpMiddlewareOptions.ValidationErrorsReturnHeader

sungam3r:
It helps when migrating to v7 to distinguish 400 from graphql validation layer and 400 from something else. In the first case caller **can** try to deserialize response body into "normal" graphql response.

### DIFF
--- a/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
+++ b/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
@@ -301,7 +301,7 @@ public class GraphQLHttpMiddleware : IUserContextBuilder
             {
                 statusCode = HttpStatusCode.BadRequest;
                 if (!string.IsNullOrEmpty(_options.ValidationErrorsReturnHeader))
-                    context.Response.Headers.Add(_options.ValidationErrorsReturnHeader, Microsoft.Extensions.Primitives.StringValues.Empty);
+                    context.Response.Headers.Add(_options.ValidationErrorsReturnHeader, "true");
             }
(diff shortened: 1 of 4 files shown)


src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -298,7 +298,11 @@ protected virtual ValueTask<bool> HandleAuthorizeWebSocketConnectionAsync(HttpCo
             if (result.Errors?.Any(e => e is HttpMethodValidationError) == true)
                 statusCode = HttpStatusCode.MethodNotAllowed;
             else if (_options.ValidationErrorsReturnBadRequest)
+            {
                 statusCode = HttpStatusCode.BadRequest;
+                if (!string.IsNullOrEmpty(_options.ValidationErrorsReturnHeader))
+                    context.Response.Headers.Add(_options.ValidationErrorsReturnHeader, "true");
```
sungam3r, line 304: I doubt about `true` value but without any value I don't see added header into response.

sungam3r: Another option - add delegate to options class instead of string but it looks complicated.

Shane32:
Why isn’t checking content type enough? Do you have some type of situation in mind?

Shane32:
Does any other GraphQL library have a similar return header? Listed in draft spec or anything?

sungam3r:
Content-Type... 🤦

PR #885, opened 2022-08-01
Add JWT bearer authentication sample

Shane32:
Due to the number of questions we answer regarding JWT bearer authentication, I thought it best to write a proper sample.

The sample confines token generation and validation to a `JwtHelper` class.  An `OAuthController` class simulates an extremely basic OAuth2-compliant endpoint.  The sample also demonstrates authenticated subscription support from a browser, which is a real pain to get working properly without a working sample.

The sample uses a symmetric security key, but also includes some code demonstrating how to use asymmetric key pairs.

### DIFF
--- a/samples/Samples.Jwt/Controllers/OAuthController.cs
+++ b/samples/Samples.Jwt/Controllers/OAuthController.cs
@@ -15,7 +15,7 @@ public class OAuthController : Controller
         if (grant_type == "client_credentials" && client_id == "sampleClientId" && client_secret == "sampleSecret")
         {
             // provide a signed JWT token with an 'Administrator' role claim
-            var token = JwtHelper.CreateSignedToken(new Claim("role", "Administrator"));
+            var token = JwtHelper.Instance.CreateSignedToken(new Claim("role", "Administrator"));
             return Json(new
             {
                 access_token = token.Token,
--- a/samples/Samples.Jwt/Controllers/OAuthController.cs
+++ b/samples/Samples.Jwt/Controllers/OAuthController.cs
@@ -3,7 +3,7 @@ using Microsoft.AspNetCore.Mvc;
 
 namespace GraphQL.Server.Samples.Jwt.Controllers;
 
-public class AuthController : Controller
+public class OAuthController : Controller
 {
     // sample OAuth2-compatible authorization endpoint supporting the 'client credentials' flow
     [HttpGet]
--- a/samples/Samples.Jwt/JwtWebSocketAuthenticationService.cs
+++ b/samples/Samples.Jwt/JwtWebSocketAuthenticationService.cs
@@ -37,7 +37,7 @@ public class JwtWebSocketAuthenticationService : IWebSocketAuthenticationService
                 var token = authPayload.Authorization.Substring(7);
                 // parse the token in the same manner that the .NET AddJwtBearer() method does
                 var handler = new Microsoft.IdentityModel.JsonWebTokens.JsonWebTokenHandler();
-                var result = handler.ValidateToken(token, JwtHelper.TokenValidationParameters);
+                var result = handler.ValidateToken(token, JwtHelper.Instance.TokenValidationParameters);
                 if (result.IsValid)
                 {
                     // convert JWT tokens with "role" as the claim type (a JWT defined claim type) to the proper .NET claim type constant
--- a/samples/Samples.Jwt/JwtWebSocketAuthenticationService.cs
+++ b/samples/Samples.Jwt/JwtWebSocketAuthenticationService.cs
@@ -1,0 +1,66 @@
+using System.Security.Claims;
+using GraphQL.Server.Transports.AspNetCore.WebSockets;
+using GraphQL.Transport;
+using Microsoft.AspNetCore.Authentication.JwtBearer;
+
+namespace GraphQL.Server.Samples.Jwt;
+
+/// <summary>
+/// Authenticates WebSocket connections via the 'payload' of the initialization packet.
+/// This is necessary because WebSocket connections initiated from the browser cannot
+/// authenticate via HTTP headers.
+/// <br/><br/>
+/// This class is not used when authenticating over GET/POST.
+/// </summary>
+public class JwtWebSocketAuthenticationService : IWebSocketAuthenticationService
+{
+    private readonly IGraphQLSerializer _graphQLSerializer;
+
+    public JwtWebSocketAuthenticationService(IGraphQLSerializer graphQLSerializer)
+    {
+        _graphQLSerializer = graphQLSerializer;
+    }
+
+    public Task AuthenticateAsync(IWebSocketConnection connection, string subProtocol, OperationMessage operationMessage)
+    {
+        try
+        {
+            // for connections authenticated via HTTP headers, no need to reauthenticate
+            if (connection.HttpContext.User.Identity?.IsAuthenticated ?? false)
+                return Task.CompletedTask;
+
+            // attempt to read the 'Authorization' key from the payload object and verify it contains "Bearer: XXXXXXXX"
+            var authPayload = _graphQLSerializer.ReadNode<AuthPayload>(operationMessage.Payload);
+            if (authPayload != null && authPayload.Authorization != null && authPayload.Authorization.StartsWith("Bearer ", StringComparison.Ordinal))
+            {
+                // pull the token from the value
+                var token = authPayload.Authorization.Substring(7);
+                // parse the token in the same manner that the .NET AddJwtBearer() method does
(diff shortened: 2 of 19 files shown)


Shane32:
I added more code and comments in the hope of making the source code more useful to developers.  I know how much time I spent trying to figure out asymmetric key pairs and stuff, and so having a working demo here might save other developers hours or perhaps days of research.

tests/Samples.Tests/WebSocketExtensions.cs
```diff
@@ -45,4 +45,19 @@ public static async Task<OperationMessage> ReceiveMessageAsync(this WebSocket so
...
+        var mem = new MemoryStream();
+        ValueWebSocketReceiveResult response;
+        do
+        {
+            var buffer = new byte[1024];
+            response = await socket.ReceiveAsync(new MemoryBytes(buffer), cts.Token);
+            mem.Write(buffer, 0, response.Count);
+        } while (!response.EndOfMessage);
+        response.MessageType.ShouldBe(WebSocketMessageType.Close);
+    }
```
sungam3r, line 62:
```suggestion
    public static async Task ReceiveCloseMessageAsync(this WebSocket socket)
    {
        using var cts = new CancellationTokenSource(5000);
        var drainBuffer = new byte[1024];
        ValueWebSocketReceiveResult response;
        do
        {
            response = await socket.ReceiveAsync(new MemoryBytes(drainBuffer), cts.Token);
        } while (!response.EndOfMessage);
        response.MessageType.ShouldBe(WebSocketMessageType.Close);
    }
```

samples/Samples.Jwt/JwtWebSocketAuthenticationService.cs
```diff
@@ -0,0 +1,66 @@
...
+            if (authPayload != null && authPayload.Authorization != null && authPayload.Authorization.StartsWith("Bearer ", StringComparison.Ordinal))
+            {
+                // pull the token from the value
+                var token = authPayload.Authorization.Substring(7);
+                // parse the token in the same manner that the .NET AddJwtBearer() method does
+                var handler = new Microsoft.IdentityModel.JsonWebTokens.JsonWebTokenHandler();
+                var result = handler.ValidateToken(token, JwtHelper.Instance.TokenValidationParameters);
+                if (result.IsValid)
+                {
+                    // convert JWT tokens with "role" as the claim type (a JWT defined claim type) to the proper .NET claim type constant
```
sungam3r, line 43: What is _proper .NET claim type constant_ here?

samples/Samples.Jwt/JwtWebSocketAuthenticationService.cs
```diff
@@ -0,0 +1,66 @@
...
+
+                    // set the ClaimsPrincipal for the HttpContext; authentication will take place against this object
+                    connection.HttpContext.User = principal;
+                }
+            }
+        }
+        catch
+        {
+            // no errors during authentication should throw an exception
+            // specifically, attempting to validate an invalid JWT token will result in an exception, which should be ignored
```
sungam3r, line 56: I think this is a good case to use logging in example - inject MS logger and use it here to track invalid JWT tokens.

samples/Samples.Jwt/JwtWebSocketAuthenticationService.cs
Shane32: See: https://docs.microsoft.com/en-us/dotnet/api/system.security.claims.claimsidentity.roleclaimtype?view=net-6.0

The constant is: `http://schemas.microsoft.com/ws/2008/06/identity/claims/role`.

See: https://docs.microsoft.com/en-us/dotnet/api/system.security.claims.claimtypes.role?view=net-6.0

The ASP.NET JWT code does this conversion, I believe, but since we must recreate similar code for use with WebSockets, I've done so where described by the comments there.  If this conversion does not take place, .NET will not recognize the JWT roles as roles when evaluating via `IsInRole`.

Please keep in mind this is a _sample_ implementation.

samples/Samples.Jwt/JwtWebSocketAuthenticationService.cs
Shane32: Well, I wouldn't for production use any more than I'd use `ILogger` to track every invalid username / password attempted on the site.

Again, this is just sample code.  If you think adding an logger to it would benefit the sample, we could do so.

Shane32: I did try to make the sample somewhat useful for production applications, as it seems that many 'sample implementations' get used in production, as can be seen by the number of times we are asked questions regarding the GraphQL.NET Harness sample's middleware for ASP.NET Core.

samples/Samples.Jwt/Controllers/OAuthController.cs
```diff
@@ -0,0 +1,29 @@
...
+    [HttpGet]
+    [HttpPost]
+    [Route("token")]
+    public IActionResult Token(string grant_type, string client_id, string client_secret)
+    {
+        // validate the provided client id and client secret
+        if (grant_type == "client_credentials" && client_id == "sampleClientId" && client_secret == "sampleSecret")
+        {
+            // provide a signed JWT token with an 'Administrator' role claim
+            var token = JwtHelper.Instance.CreateSignedToken(new Claim("role", "Administrator"));
```
Shane32, line 18: Notice here the JWT claim type of "role".  Finding documentation describing that "role" is the proper claim type is not easy, as many lists of JWT claim types omit this claim type.  Some list it as "roles", and perhaps that is more correct.  Nevertheless, it is apparent that the official MS JWT implementation honors the claim type of "role" and translates it for use within .NET accordingly.  So that's what I have been using.

samples/Samples.Jwt/Program.cs
```diff
@@ -0,0 +1,67 @@
...
+        opts.DefaultScheme = JwtBearerDefaults.AuthenticationScheme;
+        // configure custom authorization policies here, if any
+    })
+    .AddJwtBearer(opts => opts.TokenValidationParameters = JwtHelper.Instance.TokenValidationParameters);
+
+var app = builder.Build();
+app.UseDeveloperExceptionPage();
+// enable WebSocket support within the ASP.NET Core framework
+app.UseWebSockets();
+// use ASP.Net Core authentication (again, for GET/POST requests mainly)
```
sungam3r, line 43:
```suggestion
// use ASP.NET Core authentication (again, for GET/POST requests mainly)
```

samples/Samples.Jwt/Controllers/OAuthController.cs
Shane32: See: https://www.iana.org/assignments/jwt/jwt.xhtml#claims

We could test "roles" here to see if it works also...

samples/Samples.Jwt/Program.cs
```diff
@@ -0,0 +1,67 @@
...
+    // that carry the "Administrator" role to connect; connections are refused if this policy is not met
+
+    // because of this rule, anonymous access to perform introspection queries is not available
+
+    // remove this authorization policy if you wish to allow anonymous access to your schema and set
+    // authorization policies on individual graphs and fields
+
+    opts.AuthorizedRoles.Add("Administrator");
+});
+// enable ASP.Net Core routing for razor pages and MVC controllers
```
sungam3r, line 58:
```suggestion
// enable ASP.NET Core routing for razor pages and MVC controllers
```

samples/Samples.Jwt/JwtWebSocketAuthenticationService.cs
sungam3r, line 43:
```suggestion
                    // convert JWT tokens with "role" as the claim type (a JWT defined claim type) to the proper .NET claim type constant http://schemas.microsoft.com/ws/2008/06/identity/claims/role
```

sungam3r: Understood, thanks.

samples/Samples.Jwt/JwtWebSocketAuthenticationService.cs
sungam3r, line 56:
```suggestion
            // specifically, attempting to validate an invalid JWT token will result in an exception, which may be logged or simply ignored to not generate an inordinate amount of logs without purpose
```

samples/Samples.Jwt/Controllers/OAuthController.cs
sungam3r: As you wish.

sungam3r (approved):

samples/Samples.Jwt/JwtWebSocketAuthenticationService.cs
sungam3r: I posted a suggestion regarding comment.

PR #892, opened 2022-08-13
Feature/assembly-strongname

Shane32:
See:
- https://github.com/graphql-dotnet/graphql-dotnet/pull/3265

Thanks to @anderslaub for the effort towards this PR.


Directory.Build.targets
```diff
@@ -15,4 +15,13 @@
     <AnalysisMode>Recommended</AnalysisMode>
     <PackageReadmeFile>README.md</PackageReadmeFile>
   </PropertyGroup>
+
+  <PropertyGroup Condition="'$(SignAssembly)' == 'true'">
+    <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)graphql.snk</AssemblyOriginatorKeyFile>
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(IsTestProject)' != 'true'">
```
sungam3r, line 23: I would move this ItemGroup to another one to keep them together.

Directory.Build.props
```diff
@@ -28,7 +28,11 @@
     <RootNamespace>GraphQL.Server.$(MSBuildProjectName)</RootNamespace>
     <PackageId>GraphQL.Server.$(MSBuildProjectName)</PackageId>
 
-    <GraphQLVersion>6.0.0-preview-612</GraphQLVersion>
+    <GraphQLVersion>7.0.0-preview-641</GraphQLVersion>
+
+    <SignAssembly>true</SignAssembly>
+    <GenerateAssemblyInfo>true</GenerateAssemblyInfo>
```
sungam3r, line 34:
```suggestion
```

https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#generateassemblyinfo

I would remove this property from other projects too.

.github/workflows/test.yml
```diff
@@ -57,8 +57,12 @@ jobs:
         run: dotnet build --no-restore -c Release
       - name: Build solution [Debug]
         run: dotnet build --no-restore -c Debug
-      - name: Test solution [Debug]
+      - name: Test solution [Debug] with code coverage
+        if: ${{ startsWith(matrix.os, 'ubuntu') }}
         run: dotnet test --no-restore --no-build -p:CollectCoverage=true
+      - name: Test solution [Debug] without code coverage
```
sungam3r, line 63: Interesting. Some problems or just enchancement?

sungam3r: https://github.com/graphql-dotnet/parser/pull/266

sungam3r (approved):

.github/workflows/test.yml
Shane32: No idea why but won’t pass signature checks with code coverage enabled. Since signatures are only verified on .NET, and as .NET only runs on Windows, and since we don’t upload coverage reports on Windows anyway, I just disabled code coverage calculations on Windows. 

I suspect that .NET or something must inject code into the loaded dlls during code coverage reports, which breaks the signature, but I am entirely just guessing. I can reproduce the issue locally and can verify that the dlls have valid signatures — in fact the exact same dlls that pass without code coverage enabled fail when code coverage is enabled.

Shane32: It does show that signatures are working properly however!

Directory.Build.targets
Shane32: The rest are PropertyGroup and so cannot be merged.

.github/workflows/test.yml
Shane32: To reproduce,
1. `dotnet build`
2. `dotnet test --no-restore --no-build` which passes
3. `dotnet test --no-restore --no-build -p:CollectCoverage=true` which fails due to signature failure

sungam3r: It is Coverlet instrumentation.

PR #893, opened 2022-08-16
Bump to GraphQL.NET 7.0.0

Shane32:


tests/Samples.Complex.Tests/BaseTest.cs
```diff
@@ -15,6 +16,14 @@ protected BaseTest()
     {
         Host = Program.CreateHostBuilder(Array.Empty<string>())
              .ConfigureWebHost(webBuilder => webBuilder.UseTestServer())
+             .ConfigureServices(services =>
+             {
+                 services.AddGraphQL(b => b.ConfigureExecution((options, next) =>
```
sungam3r, line 21: How did it work before without `AddGraphQL`?

Shane32: It runs `AddGraphQL` within the `Program.cs` file.  For some reason the change to `UseApolloTracing` now started returning tracing results in the tests (which it should, I believe).  So this call to `AddGraphQL` serves only to disable metrics for the purposes of the tests, leaving the sample project in tact.

Shane32: Oh, the reason is obvious -- the call to ConfigureExecutionOptions sets EnableMetrics, which now occurs prior to `UseApolloTracing`.  I should have changed `UseApolloTracing()` to `UseApolloTracing(false)`, as the sample is now incorrect.

Shane32: #903

PR #515, opened 2021-03-09
Federated tracing support

nur858:
added GraphQLHttpRequestExtensions  to check if federated tracing support is enabled through http header

PR #894, opened 2022-08-16
Add federation tracing extension method

Shane32:
Replaces:
- #515


Shane32:
Merging as this was approved 8 months ago with no objections.

PR #897, opened 2022-08-16
Fix packing readme and analysis recommendations

Shane32:
On NuGet.org, the readme file is not appearing.  This is because the PackageReadmeFile did not get set while building.  This is because the condition `IsPackable == true` does not work if `IsPackable` is not explicitly set, even though when it is not set, it is implied to be true for the purposes of packing and within ItemGroups.  This also caused the analysis to be disabled.

The analysis mode revealed that some xml comments on some old code was missing.

I ran the pack locally and examined the output to be sure it built correctly this time.

It is not critical to release 7.0.1, as there are no breaking changes and the 7.0.0 code should work just fine.  Nevertheless, we can release 7.0.1 if you think we should @sungam3r , before everyone downloads it.  Or, we could wait a week or two to see if there are any other issues that need to be addressed.

### DIFF
--- a/src/Authorization.AspNetCore/DefaultAuthorizationErrorMessageBuilder.cs
+++ b/src/Authorization.AspNetCore/DefaultAuthorizationErrorMessageBuilder.cs
@@ -5,6 +5,9 @@ using Microsoft.AspNetCore.Authorization.Infrastructure;
 
 namespace GraphQL.Server.Authorization.AspNetCore;
 
+/// <summary>
+/// Default error message builder implementation
+/// </summary>
 [Obsolete("This class will be removed in v8 as revealing authorization requirements may be a security risk; please use ErrorInfoProvider if you require detailed access-denied error messages.")]
 public class DefaultAuthorizationErrorMessageBuilder : IAuthorizationErrorMessageBuilder
 {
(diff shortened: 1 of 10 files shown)


src/Authorization.AspNetCore/IAuthorizationErrorMessageBuilder.cs
```diff
@@ -4,6 +4,9 @@
 
 namespace GraphQL.Server.Authorization.AspNetCore;
 
+/// <summary>
+/// Error message builder for authorization errors
```
sungam3r, line 8:
```suggestion
/// Error message builder for authorization errors.
```

Shane32:
> Does this mean that other projects also have this problem? 

Some of my other projects certainly did.

> The same for AnalysisMode.

I really don't like the analyzer running for tests.  It's very annoying, when it nit-picks at stuff that really doesn't make a difference, often causing test code that is inappropriate or harder to read than before.  Here's some examples that annoy me for tests:

1. `async _ => true` vs `Task.FromResult(true)`
2. Methods containing one line of executable line of code vs `=>` syntax.
3. Misc naming conventions, which are often not appropriate for tests
4. Unreferenced methods warnings (necessary for tests)
5. Warnings that methods may be marked as static (necessary for tests)
6. Warnings about unnecessary `using` statements (which are still a problem even without the analyzer)
7. Warnings about one-line methods

I just looked at the 60 other warnings besides CA1707.  There's only one warning that makes any sense for tests -- the `int.Parse` culture warnings.  The rest do not apply.  I definitely don't want the analyzer to run for tests.  Sample code, maybe, but not tests.

Note that item 2 above is a particular problem for sample code; example below:

```cs
// -- this will not compile --
services.AddGraphQL(b => b
    .AddUserContextBuilder(httpContext =>
    {
        // you may add any data from the user context, like the value
        // of httpContext.User or Environment.IsDevelopment() or whatever you like
        return new MyUserContext(httpContext.User, Environment.IsDevelopment()));
    })
);

// -- .NET complains and wants this instead, which can be harder to read and does not leave room for comments --
services.AddGraphQL(b => b
    .AddUserContextBuilder(httpContext => new MyUserContext(httpContext.User, Environment.IsDevelopment()))
);
```

I suggest we merge this to fix the analyzer and readme as it was intended to be.  If you want to create a subsequent PR with additional changes to the build process, I'll review it at that time.

src/Authorization.AspNetCore/DefaultAuthorizationErrorMessageBuilder.cs
```diff
@@ -5,6 +5,9 @@
 
 namespace GraphQL.Server.Authorization.AspNetCore;
 
+/// <summary>
+/// Default error message builder implementation
```
Shane32, line 9:
```suggestion
/// Default error message builder implementation.
```

sungam3r:
OK. Then I propose to disable all the rules that are not suitable for tests while turning the analysis on.

sungam3r (approved):

PR #903, opened 2022-08-17
Fix call to UseApolloTracing in Complex sample

Shane32:


samples/Samples.Complex/Startup.cs
```diff
@@ -45,7 +43,8 @@ public void ConfigureServices(IServiceCollection services)
             .AddErrorInfoProvider<CustomErrorInfoProvider>()
             .AddDataLoader()
             .AddUserContextBuilder(context => new Dictionary<string, object?> { { "user", context.User.Identity!.IsAuthenticated ? context.User : null } })
-            .AddGraphTypes(typeof(ChatSchema).Assembly));
+            .AddGraphTypes(typeof(ChatSchema).Assembly)
+            .UseApolloTracing(_ => Environment.IsDevelopment()));
```
Shane32, line 47: The call to UseApolloTracing was moved to the bottom of the list, not because it executes differently, but because it is more similar to how ASP.NET Core is ordered -- and this is the order it will be executed in.

sungam3r (approved):

PR #910, opened 2022-08-20
Enable analyzers and consolidate project settings for samples and tests

sungam3r:
Rel #897


Shane32:
I believe I made my feelings on this matter exceeding clear. The analyzer should not be enabled for tests, and almost all of the changes here I disagree with.

samples/Samples.Jwt/Controllers/OAuthController.cs
```diff
@@ -9,10 +9,10 @@ public class OAuthController : Controller
     [HttpGet]
     [HttpPost]
     [Route("token")]
-    public IActionResult Token(string grant_type, string client_id, string client_secret)
+    public IActionResult Token([Bind(Prefix = "grant_type")] string grantType, [Bind(Prefix = "client_id")] string clientId, [Bind(Prefix = "client_secret")] string clientSecret)
```
Shane32, line 12: This should be reverted. It just makes the code harder to read.

samples/Samples.Schemas.Chat/Services/Chat.cs
```diff
@@ -20,7 +20,7 @@ public IEnumerable<Message> GetAllMessages()
     public IEnumerable<Message> GetMessageFromUser(string from)
     {
         lock (_messages)
-            return _messages.Where(x => string.Equals(x.From, from, StringComparison.InvariantCultureIgnoreCase)).ToList();
+            return _messages.Where(x => string.Equals(x.From, from, StringComparison.Ordinal)).ToList();
```
Shane32, line 23: This change should be reverted. It was explicitly set to a case insensitive search, as would be appropriate in this example.

samples/Samples.Schemas.Chat/Services/Chat.cs
```diff
@@ -61,7 +61,7 @@ public Message PostMessage(MessageInput message)
     public IObservable<Message> SubscribeAll() => _broadcaster.Where(x => x.Type == EventType.NewMessage).Select(x => x.Message!);
 
     public IObservable<Message> SubscribeFromUser(string from)
-        => SubscribeAll().Where(x => string.Equals(x.From, from, StringComparison.InvariantCultureIgnoreCase));
+        => SubscribeAll().Where(x => string.Equals(x.From, from, StringComparison.Ordinal));
```
Shane32, line 64: Ditto

tests/Authorization.AspNetCore.Tests/AuthorizationValidationRuleTests.cs
```diff
@@ -311,6 +311,7 @@ private static ISchema BasicSchema<T>()
     public class BasicQueryWithAttributesAndClassPolicy
     {
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Style", "IDE0060:Remove unused parameter", Justification = "for tests")]
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Performance", "CA1822:Mark members as static", Justification = "for tests")]
```
Shane32, line 314: The mark as static warning should be globally ignored.

tests/Samples.Jwt.Tests/EndToEndTests.cs
```diff
@@ -133,7 +133,7 @@ private async Task<string> GetJwtToken()
         return await ProcessOAuthResponseAsync(response);
     }
 
-    private async Task<string> ProcessOAuthResponseAsync(HttpResponseMessage response)
+    private static async Task<string> ProcessOAuthResponseAsync(HttpResponseMessage response)
```
Shane32, line 136: There is no reason we should be prompted to mark this as static. This is a test. Performance does not matter and it is highly questionable if performance would be any different. Marking it as static does not help readability. Ditto for many of the rest of the changes

tests/Transports.AspNetCore.Tests/WebSockets/AsyncMessagePumpTests.cs
```diff
@@ -14,7 +16,7 @@ public void CallsSynchronously()
         var sb = new StringBuilder();
         var pump = new AsyncMessagePump<string>(t =>
         {
-            Thread.Sleep(int.Parse(t));
+            Thread.Sleep(int.Parse(t, CultureInfo.InvariantCulture));
```
Shane32, line 19: Even these culture additions have extremely limited value. They will not affect production code. They will not affect the tests in the cloud. They can only possibly affect a developer that runs the tests locally with an incompatible culture config. (And presumably you have not had an issue running the tests.)

tests/Transports.AspNetCore.Tests/WebSockets/ReusableMemoryReaderStreamTests.cs
```diff
@@ -10,6 +10,8 @@ public ReusableMemoryReaderStreamTests()
         _stream = new ReusableMemoryReaderStream(_buffer);
     }
 
+    public void Dispose() => _stream.Dispose();
```
Shane32, line 13: This is correct … but does it matter? The GC will catch it. These are just tests. We are not creating 1000 of these instances.

samples/Samples.Complex/CustomErrorInfoProvider.cs
```diff
@@ -21,7 +21,7 @@ public override ErrorInfo GetInfo(ExecutionError executionError)
         return info;
     }
 
-    private string GetAuthorizationErrorMessage(AccessDeniedError error)
+    private static string GetAuthorizationErrorMessage(AccessDeniedError error)
```
Shane32, line 24: This use of “static” makes sense. Probably the only one I agree with.

Shane32:
I have completed a full review of this PR.

samples/Samples.Jwt/Controllers/OAuthController.cs
sungam3r: Why did you use `snake_case` naming notation here?

tests/Authorization.AspNetCore.Tests/AuthorizationValidationRuleTests.cs
sungam3r: In general, this rule works with the exceptions of special cases.

tests/Samples.Jwt.Tests/EndToEndTests.cs
sungam3r: > There is no reason we should be prompted to mark this as static.

There is even if this is not so important.

> Marking it as static does not help readability.

This is just a static modifier, come on.

tests/Transports.AspNetCore.Tests/WebSockets/AsyncMessagePumpTests.cs
sungam3r: Little benefit is better than nothing. I remember the problems with the culture with the dash sign.

tests/Transports.AspNetCore.Tests/WebSockets/ReusableMemoryReaderStreamTests.cs
sungam3r: Whether this change has a value here/now or not, it is not important. The code should be correct, regardless of whether it is from test or not. The analyzers helped to find the lost call and it's good.

sungam3r:
I have the conviction that the test code should be analyzed, all the more so when the inclusion of analyzers allows to remove additional jumping around `IsPackable`.

samples/Samples.Jwt/Controllers/OAuthController.cs
Shane32: Part of OAuth spec

tests/Authorization.AspNetCore.Tests/AuthorizationValidationRuleTests.cs
Shane32: No. For production code it typically, but not always, makes sense. Usually it indicates where the user forgot to mark a method as virtual more than it marks where a method can be made static. This is a dumb analyzer that is not taking the code into context when considering the recommendation. Static methods should be helper functions and similar, not methods that provide class specific logic flow. (Perhaps that’s my opinion however.)

When it comes to tests there is often no reason to mark methods as static or virtual. Marking these methods either way is entirely non beneficial.

tests/Transports.AspNetCore.Tests/WebSockets/AsyncMessagePumpTests.cs
Shane32: I don’t have a problem with this change or this particular analyzer rule, even if it is pointless. I do have a problem with adding all the rest of the analyzer rules. We shouldn’t add the analyzer just to disable 99% of the rules. The analyzers are unnecessary and should not be added.

tests/Transports.AspNetCore.Tests/WebSockets/ReusableMemoryReaderStreamTests.cs
Shane32: No, I’m sorry but this is ridiculous and unnecessary. These are tests. You write them and move on. 

Did you even look at the code here? There is no implementation of the dispose method. This method does not do ANYTHING. Even if disposal mattered.

Shane32:
> I have the conviction that the test code should be analyzed, all the more so when the inclusion of analyzers allows to remove additional jumping around `IsPackable`.

I completely disagree.

Directory.Build.props
```diff
@@ -22,8 +23,8 @@
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <EnableNETAnalyzers>true</EnableNETAnalyzers>
+    <AnalysisMode>Recommended</AnalysisMode>
     <Nullable>enable</Nullable>
-    <IsPackable>true</IsPackable>
```
Shane32, line 26: This must be reverted.  Conditions that rely on `IsPackable` do not work properly without it.

Directory.Build.targets
```diff
@@ -6,20 +6,11 @@
     <Exclude>[GraphQL.Samples.Schemas.Chat]*,[GraphQL.Samples.Server]*</Exclude>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(OutputType)' == 'Exe'">
-    <IsPackable>false</IsPackable>
-  </PropertyGroup>
-
```
Shane32, line 12: This likely has to be reverted after `<IsPackable>true<IsPackable>` is reverted

Directory.Build.targets
```diff
@@ -6,20 +6,11 @@
...
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(OutputType)' == 'Exe'">
-    <IsPackable>false</IsPackable>
-  </PropertyGroup>
-  
   <PropertyGroup Condition="'$(IsPackable)' != 'true'">
     <NoWarn>$(NoWarn);CS1591;IDE1006;IDE0053;CS1998</NoWarn>
     <WarningsNotAsErrors>$(WarningsNotAsErrors);IDE0060</WarningsNotAsErrors>
   </PropertyGroup>
```
Shane32, line 12: This always evaluates to `true` and should not be so.  You need to add `<IsPackable>true</IsPackable>` in the `Directory.Build.props` file to fix this, along with the other changes to set `IsPackable` to `false` when appropriate.  It is too bad that msbuild does not do a better job with this, but we cannot change this.  If you find a better solution, let me know.

Tests.props
```diff
@@ -3,8 +3,7 @@
 
   <PropertyGroup>
     <SingleTestPlatform Condition="'$(SingleTestPlatform)' == ''">false</SingleTestPlatform>
-    <NoWarn>$(NoWarn);IDE0017</NoWarn>
-    <IsPackable>false</IsPackable>
```
Shane32, line 7: This `<IsPackable>false</IsPackable>` line must be reverted

Tests.props
```diff
@@ -3,8 +3,7 @@
...
+    <NoWarn>$(NoWarn);IDE0017;CA1707;CA1816;CA1825;CA2201</NoWarn>
```
Shane32, line 6:
```suggestion
    <NoWarn>$(NoWarn);IDE0017;CA1707;CA1816;CA1825;CA2201;CA1822</NoWarn>
```
If analyzers are important to you disable CA1822 and revert associated changes (`static` keywords) and I'll review again.

samples/Samples.Jwt/Controllers/OAuthController.cs
Shane32: This should be reverted.  Naming rules should be excluded in `Tests.props`.

samples/Samples.Schemas.Chat/Services/Chat.cs
```diff
@@ -20,7 +20,7 @@ public IEnumerable<Message> GetAllMessages()
     public IEnumerable<Message> GetMessageFromUser(string from)
     {
         lock (_messages)
-            return _messages.Where(x => string.Equals(x.From, from, StringComparison.InvariantCultureIgnoreCase)).ToList();
+            return _messages.Where(x => string.Equals(x.From, from, StringComparison.OrdinalIgnoreCase)).ToList();
```
Shane32, line 23: I don't understand the logic of the change here.  Did it violate a rule?  Sorta puzzled...

tests/Transports.AspNetCore.Tests/Transports.AspNetCore.Tests.csproj
```diff
@@ -11,6 +11,10 @@
     <TargetFrameworks>net6</TargetFrameworks>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <NoWarn>$(NoWarn);CA1835</NoWarn>
```
Shane32, line 15: CA1835 should be a global ignore within tests, as some frameworks tested do not support such a change.  Within the main project we use ifdef to use the ideal overload.

Shane32 (changes requested):
I have reviewed this PR again showing the changes required before I would consider merging.

samples/Samples.Jwt/Controllers/OAuthController.cs
sungam3r: Reverted. `Tests.props` file is not used for samples. I added `Samples.props`.

tests/Authorization.AspNetCore.Tests/AuthorizationValidationRuleTests.cs
sungam3r: ignored

tests/Transports.AspNetCore.Tests/WebSockets/AsyncMessagePumpTests.cs
sungam3r: > We shouldn’t add the analyzer just to disable 99% of the rules

6 rules are disabled for tests and 4 for samples. It is far from 99%.

tests/Transports.AspNetCore.Tests/WebSockets/ReusableMemoryReaderStreamTests.cs
sungam3r: The point is not what method does now or not, but to write the code so as to minimize possible problems, no matter how they arise.

> ridiculous and unnecessary

In the same way, we can answer the use of analyzers in the main code for some cases, but this is not a reason not to use them. I do not want to keep in my head (and I can’t) that something is not necessary in this place, but it in other does, and selectively apply the rules depending on what I now consider important and what is not.

This change does not harm the code, but brings its quality a bit closer to the main code.

Directory.Build.props
sungam3r: I thought about it. It looks strange, especially since before everything seemed to work properly with maybe the exception of Readme not showing on nuget.org, which we simply may not notice. The same doubts about all other projects. I changed some conditions around `IsPackable`. Only one condition left that should work for `ItemGroup`:
```
<ItemGroup Condition="'$(IsPackable)' == 'true'">
    <None Include="..\..\assets\logo.64x64.png" Pack="true" PackagePath="\" />
    <None Include="..\..\README.md" Pack="true" PackagePath="\" />
    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
  </ItemGroup>
```

Directory.Build.targets
sungam3r: https://github.com/graphql-dotnet/server/pull/910#discussion_r951075226

Directory.Build.targets
sungam3r: >  It is too bad that msbuild does not do a better job with this

100%

> If you find a better solution, let me know.

I removed this condition and added explicit nowarns into tests.props and samples.props.

Tests.props
sungam3r: `IsPackable` is false by default for tests as it was before.

Tests.props
sungam3r: reverted

samples/Samples.Schemas.Chat/Services/Chat.cs
sungam3r: Yes, https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1309

tests/Transports.AspNetCore.Tests/Transports.AspNetCore.Tests.csproj
sungam3r: ok

Directory.Build.props
Shane32: True, but this is confusing.  Nobody is going to understand in the future why `'$(IsPackable)' == 'true'` works in one spot and not another.  And if anyone does need a conditional on a `PropertyGroup` item, they will not be able to do and would need to revert these changes.

In addition, we have a consistent build process for the GraphQL.NET project, this project and the relay project.  It works well.  We should not make a change here that differs from the other projects.

Please revert this change.

Tests.props
Shane32: Only when `IsPackable` is not set in `Directory.Build.props`.

Directory.Build.targets
Shane32: This could be moved in to Samples.props now.

samples/Samples.Authorization/Samples.Authorization.csproj
```diff
@@ -1,9 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
+  <Import Project="../../Samples.props" />
```
Shane32, line 2: I designed the sample projects to have a csproj that is valid by itself, based on the .NET template, as part of the sample.  The only required change is the reference to the nuget package rather than the project.  For that reason, I do not want to change this file any more than absolutely necessary.

The projects are designed to be as close to user-code as possible; users should be able to open a new template, and copy the code directly in and/or see the difference between the sample and the template.  I would like to update these samples with each new version of Visual Studio so that the sample matches the current template, and simply contains the few changes necessary for GraphQL.NET Server to operate.  That is the goal of the samples, and I feel this is important and valuable.

This is why there are no global using statements present.  It was done this way on purpose, specifically so that users can copy files from the sample directly into their project with no changes.  (Assuming their project is based on the current .NET template, of course.)

If we require some additional settings for the sample projects to facilitate the tests, let's add another `Directory.Build.props` in the samples folder, so that no additional tags are necessary within this csproj.

samples/Samples.Authorization/Schema/Query.cs
```diff
@@ -1,5 +1,4 @@
 using System.ComponentModel;
-using GraphQL;
```
Shane32, line 2: Please see comment above and revert

Shane32 (changes requested):
This PR now contains too many out-of-scope changes (changes that do not involve enabling analyzers, the topic of this PR).  Please revert the other build process changes, and I'll review again.

Shane32:
I have very few comments remaining regarding the analyzers.  If we can revert the other build process changes, we can do a proper review of this PR, titled "Enable analyzers for all projects".

Directory.Build.targets
sungam3r: I added IsPackable=false into samples.props.

Directory.Build.props
sungam3r: I reverted it a bit but I can not agree 100% that I like current status quo.

samples/Samples.Authorization/Samples.Authorization.csproj
sungam3r: Understandable but there is always some 'context' information to configure csprojs - `Directory.Build.props` for example how you mentioned before. Project configuration should be convenient for us as well. So I renamed `Samples.props` into nested `Directory.Build.props`.

tests/Transports.AspNetCore.Tests/AuthorizationHelperTests.cs
```diff
@@ -68,7 +68,7 @@ public async Task Authorize(bool expected)
         ret.ShouldBe(expected);
 
         bool ranHandler = false;
-        ret = await AuthorizationHelper.AuthorizeAsync(new AuthorizationParameters<object?>(mockContext.Object, options, async _ => { ranHandler = true; }, null, null), null);
+        ret = await AuthorizationHelper.AuthorizeAsync(new AuthorizationParameters<object?>(mockContext.Object, options, _ => { ranHandler = true; return Task.CompletedTask; }, null, null), null);
```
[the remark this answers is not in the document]
sungam3r: I just fixed compiler warning. What is bad?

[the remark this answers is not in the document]
Shane32: it is harder to read.  i don't want all the tests to have `return Task.CompletedTask` in there.  we can do that in production code to optimize performance.  but it just makes the code harder to read unnecessarily.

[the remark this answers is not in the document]
sungam3r: >  i don't want all the tests to have return `Task.CompletedTask` in there

Not all. Only those that have `async` without `await`. Only 3 places for 3465 tests. Are you serious? Really?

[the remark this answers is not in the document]
Shane32: Yes really.  You're the one that is trying to add analyzers here, not me.  We do not need analyzers here.  The code is just fine as-is.  What we need is review on code-first federation support.  I've written a LOT of tests where this is a real pain, and every time I turn around I step on this stupid warning, where it might be 3 times PER TEST, not just 3 total.  Perhaps here there's only 3 places in this project.  Bad code design is bad code design no matter if it's 3 or 300.

sungam3r:
I applied most of your proposals.

Shane32:
> I applied most of your proposals.

Ok I will review again.

samples/Directory.Build.props
```diff
@@ -0,0 +1,16 @@
...
+    <NoWarn>$(NoWarn);IDE0053;1591;CA1707;CA1848</NoWarn>
+    <WarningsNotAsErrors>$(WarningsNotAsErrors);8618;CA1716</WarningsNotAsErrors>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Using Include="GraphQL" />
+    <Using Include="GraphQL.Types" />
+  </ItemGroup>
+
```
Shane32, line 15: Remove

samples/Samples.Authorization/Program.cs
```diff
@@ -1,6 +1,5 @@
 using AuthorizationSample.Data;
 using AuthorizationSample.Schema;
-using GraphQL;
```
Shane32, line 3: This is a constant pain point on our GitHub issues board -- people not understanding what `using` statements are required -- I feel like I've answered this question a half dozen times at least.  I do not think we should simplify the using statements for these statements for that reason.

Of course I would put the using statement into the csproj or Directory.Build.props.  For the main project here you'll see I added enough global usings that there are very few cases needed otherwise.  Similarly for tests there are many global usings.  However, these are samples intended for users to most easily be able to copy and paste and/or compare to their own code.  We should leave the explicit using statements.

samples/Samples.Authorization/Samples.Authorization.csproj
```diff
@@ -2,8 +2,6 @@
 
   <PropertyGroup>
     <TargetFramework>net6.0</TargetFramework>
-    <Nullable>enable</Nullable>
-    <ImplicitUsings>enable</ImplicitUsings>
```
Shane32, line 6: These lines are part of the template.  I would not remove them.  If someone copied the sample project, it would not work in their environment because the likely would not copy the Directory.Build.props file.

tests/Transports.AspNetCore.Tests/AuthorizationHelperTests.cs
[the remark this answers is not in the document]
sungam3r: >  We do not need analyzers here. The code is just fine as-is.

This is only a subjective judgment. It is not based on formal rules, but only on emotions and mood. If you are so important to go against the recommendations of the MS here... OK. Reverted.

Shane32 (commented):
I think this is my only comment remaining.

samples/Samples.Authorization/Program.cs
Shane32: (ditto elsewhere in the samples)

Shane32: Alternatively we can update the readme to suggest putting the using statement into the csproj.  I wouldn't do so probably, but I'm okay with that option if you want to.  Then we should move it back out of the samples props file into the csproj of each sample so the samples can run "standalone".

Shane32: Another thing we may want to consider is specifically having the samples props file NOT inherit from the main one, so that we can be sure there are no settings being inherited unintentionally, so that the sample projects are "vanilla".  Then we only add stuff to the samples props file that we specifically want for testing purposes, like InternalsVisibleTo or enabling the analyzers.  Just an idea.

samples/Samples.Authorization/Samples.Authorization.csproj
sungam3r: All csprojs in repo already implicitly use `Directory.Build.props`/`Directory.Build.targets` from root. I reverted only using statements.

Shane32: These lines are part of the template. I would not remove them. If someone copied the sample project, it would not work in their environment because they likely would not copy the Directory.Build.props file.

sungam3r: >  If someone copied the sample project

Sample project needs to be launched (preferably), not copied. I have never considered example projects as something that should be copied. Who knows what else can be forgotten when copying. There are a lot of options to miss something.

Shane32: Nevertheless we have countless issues where the sample code provided was copied directly from a sample, often erroneously.  These samples ARE designed to be the basis of a new project.  There is no reason why we should break this capability in this PR.

The more that users can copy and paste the samples into their own project, the better.  We constantly refer to project source/sample code within issues and instruct people to copy from there, as was done recently with the directives/validation sample.

Shane32: For example, within the JWT sample project, I included all the proper code for all this functionality:
- Use symmetric or asymmetric security keys
- Configure such keys from the app settings (which include app secrets or environment variables)
- Generate new symmetric or asymmetric security keys
- Configure audience and issuer constants
- Proper validation of JWT tokens over GET, POST and WebSocket connections
- Basic OAuth2 endpoint to authenticate against (meets spec)

This allows full use of the sample project as a _template_ for a new JWT-based GraphQL server, and includes all the necessary pieces except the mechanism to authenticate and/or store user credentials.  Compare this to other samples seen online, where the JWT sample code consists only of some extremely basic code that includes no signature validation at all, and the user is left to figure out all the complex details for themselves.

Providing fully-functioning _templates_ from which to provide a basis for a new application should be a valued part of this project as a whole.

samples/Samples.Authorization/Program.cs
sungam3r: I added usings back.

samples/Samples.Authorization/Samples.Authorization.csproj
sungam3r: I expanded Samples section in readme.

Shane32: I disagree with those changes.

sungam3r: So do you want me to revert all changes in every csproj file into samples folder, right?

Shane32: Yes absolutely. And the readme notes. If you want to discuss, do it in a separate PR.

Shane32: This PR was (originally) discussing adding analyzers, which I still disagree with. But if we disable all the pointless and unnecessary rules then we can merge that in - if you insist. I believe we have already discussed those changes. But changing the build props and csproj files is not within scope of this PR.

samples/Samples.Schemas.Chat/Samples.Schemas.Chat.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
-    <WarningsNotAsErrors>CS8618</WarningsNotAsErrors>
+    <WarningsNotAsErrors>CS8618;CA1716</WarningsNotAsErrors>
```
sungam3r, line 5: I had to add new ignore explicitly sine you do not want to be linked to root props files

samples/Samples.Authorization/Samples.Authorization.csproj
sungam3r: I reverted all csprojs except one additional ignore in [Samples.Schemas.Chat.csproj](https://github.com/graphql-dotnet/server/pull/910/files#diff-f2ec7d4adc21fd8dada75dcc370233ccd772cac10afcb6b035c4aff2d0ccd09e)

Shane32 (approved):

PR #913, opened 2022-08-26
Allow the configuration of the response content type via options

jeffw-wherethebitsroam:
This allows users to override the default response content type without
having to implement their own middleware.

### DIFF
--- a/src/Transports.AspNetCore/GraphQLHttpMiddlewareOptions.cs
+++ b/src/Transports.AspNetCore/GraphQLHttpMiddlewareOptions.cs
@@ -93,4 +93,9 @@ public class GraphQLHttpMiddlewareOptions : IAuthorizationOptions
     /// Returns an options class for WebSocket connections.
     /// </summary>
     public GraphQLWebSocketOptions WebSockets { get; set; } = new();
+
+    /// <summary>
+    /// The Content-Type to use for GraphQL responses
+    /// </summary>
+    public string ResponseContentType { get; set; } = GraphQLHttpMiddleware.CONTENTTYPE_GRAPHQLJSON;
 }
--- a/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
+++ b/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
@@ -454,7 +454,7 @@ public class GraphQLHttpMiddleware : IUserContextBuilder
     /// <see cref="WriteJsonResponseAsync{TResult}(HttpContext, HttpStatusCode, TResult)"/>.
     /// </summary>
     protected virtual string SelectResponseContentType(HttpContext context)
-        => CONTENTTYPE_GRAPHQLJSON;
+        => _options.ResponseContentType;
 
     /// <summary>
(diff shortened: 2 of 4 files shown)


Shane32:
@jeffw-wherethebitsroam If you run the tests locally, it will automatically update the apiapprovals file; commit the change and tests should pass again.

Shane32:
Out of curiosity, how has this change to the default response content type been affecting your application?

jeffw-wherethebitsroam:
@Shane32 we have a javascript app which uses `graphql-request` with code generated with `graphql-codegen`. Responses from v7 look identical except for the new content-type. For some reason the response was interpreted as an error with the 'application/graphql+json' content type. There could be a fix for the package, but I just want to get the update out without waiting for frontend changes.

Other than that, the upgrade was fairly smooth. It was good to have a migration doc. A lot of deprecation warnings for the corresponding GraphQL v7 upgrade.

Shane32:
Ref:
- https://github.com/prisma-labs/graphql-request/issues/373

Shane32 (approved):
@sungam3r Please review.  I'll update the readme in a separate PR.

Shane32:
@jeffw-wherethebitsroam @sungam3r Before merging this, I would like to consider modifying the middleware to inspect the `Accept` header and providing an appropriate response.  This should provide wider compatibility.  This option then could become a _default_ response content-type.

Thoughts?

Shane32:
@sungam3r It seems that there may be many applications which require `application/json` as the content-type rather than `application/graphql-response+json` due to proper support missing from client-side tooling.  As such, I'm considering this a critical enhancement and will be releasing 7.1 with this enhancement shortly if you have no comments.  See also #917 and #919

jeffw-wherethebitsroam:
@Shane32 I'm happy for you to add support for the `Accept` header and that seems like a more 'correct' way to handle it. But I would also like this PR merged first to give an easier way of overriding the default. Also, `graphql-request` seems to send `Accept: */*`, so we would still need to override the default.

jeffw-wherethebitsroam:
Also, please see my comment on the `Accept` PR, https://github.com/graphql-dotnet/server/pull/919/files#r956978257. Only supporting the `Accept` header with a hard coded default creates more work for handling misbehaving clients, since you would now need to subclass `GraphQLHttpMiddleware` and copy the `Accept` logic.

Shane32:
I agree we should still add the configuration option.

src/Transports.AspNetCore/GraphQLHttpMiddlewareOptions.cs
```diff
@@ -93,4 +93,9 @@ public class GraphQLHttpMiddlewareOptions : IAuthorizationOptions
     /// Returns an options class for WebSocket connections.
     /// </summary>
     public GraphQLWebSocketOptions WebSockets { get; set; } = new();
+
+    /// <summary>
+    /// The Content-Type to use for GraphQL responses
+    /// </summary>
+    public string ResponseContentType { get; set; } = GraphQLHttpMiddleware.CONTENTTYPE_GRAPHQLJSON;
```
sungam3r, line 100: Now it's better to place CONTENTTYPE_GRAPHQLJSON internal constant in options class, I think.

sungam3r:
reviewed

Issue #917, opened 2022-08-26
[Feature] Add proper support of Accept HTTP header


Shane32:
If the request includes a supported response content type within the Accept header, the response content type should match the content type requested by the client.

See also:
- #916 

Supported response content types should be:

- `application/json`
- `application/graphql-response+json` (current draft spec)
- `application/graphql+json` (previous draft spec and v7.0 behavior) -- probably remove in some future version

Shane32:
@sungam3r What do you think makes the most sense here:

## Option 1

Assuming the requested content-type is supported, return exactly the content-type that was passed into the Accept header.

For example, when requesting `APPLICATION/JSON`, return `APPLICATION/JSON` and not `application/json; charset=utf-8`.

The accept header may return a wildcard, so in those situations we would need to pick an applicable content type.

## Option 2

Try to find a best match, and return a _compatible_ content-type with the request.

For example, when requesting `APPLICATION/JSON`, return `application/json; charset=utf-8`.

The ASP.NET-provided method for determining compatibility has changed across different versions of ASP.NET.  In .NET 6, the same sample above would return `application/graphql-response+json; charset=utf-8` because that is considered compatible (assuming of course that we gave priority to that response).  We would need to decide to either (a) use the implementation provided by whatever version of .NET is in use, or (b) copy an implementation and use that.

Note that all versions from ASP.NET 2.1 through ASP.NET 6.0 do not support a quoted charset - e.g. `charset="UTF-8"`.

PR #919, opened 2022-08-26
Select return content type based on Accept header

Shane32:


Shane32:
@sungam3r It seems that there may be many applications which require `application/json` as the content-type rather than `application/graphql-response+json`.  As such, I'm considering this a critical bug and will be releasing 7.1 with this fix shortly if you have no comments.  I would really appreciate comments on the implementation - see issue #917 .

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -454,7 +468,23 @@ protected virtual async Task<ExecutionResult> ExecuteRequestAsync(HttpContext co
...
+                for (int j = 0; j < _validMediaTypes.Length; j++)
+                {
+                    var (mediaType, responseType) = _validMediaTypes[j];
+                    if (acceptHeaders[i].MediaType.Equals(mediaType, StringComparison.OrdinalIgnoreCase))
+                        return responseType;
+                }
+            }
+        }
+
+        return CONTENTTYPE_GRAPHQLRESPONSEJSON;
```
jeffw-wherethebitsroam, line 486: The problem here is that if you only have a single misbehaving client, then you would need to subclass the `GraphQLHttpMiddleware` and recreate the logic above with a different fall-back. It would still be really good to be able to _just_ override the default response content type.

Shane32: I agree.

Shane32:
@jeffw-wherethebitsroam Want to review these changes for me?  I think I have the logic figured out, but just wondering if perhaps the `MediaTypeHeaderValue` type would be better than `string` for the option, considering the comparison logic now included here.

@sungam3r I would appreciate your comments, but have not heard from you in the last week.  I will wait another 24 hours for a review before merging these changes and releasing 7.1.  Thanks!

General logic this PR now has:

1. Loops through each media type listed in the Accept header
2. Prioritizes so that specific media types are given priority, whereas wildcard media types are given the least priority
3. For each type listed:
  a. Checks if it matches the default content type.  Includes matching on charset.  So a request of `application/json` will match on a default content type of `application/json; charset=utf-8`.
  b. Checks if it matches `application/graphql+response+json; charset=utf-8`.
  c. Checks if it matches `application/json; charset=utf-8`.
  d. Checks if it matches the deprecated `application/graphql+json; charset=utf-8`.
4. The first match found is returned; if no matches are found, the default content type is returned.

This means that by default, these responses are given:

| Accept request header | Content-Type response header |
|-|-|
| (none) | `application/graphql-response+json; charset=utf-8` |
| `*/*` | `application/graphql-response+json; charset=utf-8` |
| `application/*` | `application/graphql-response+json; charset=utf-8` |
| `application/*+json` | `application/graphql-response+json; charset=utf-8` |
| `application/graphql-response+json` | `application/graphql-response+json; charset=utf-8` |
| `application/graphql+json` | `application/graphql+json; charset=utf-8` |
| `application/json` | `application/json; charset=utf-8` |
| `*/*, application/json` | `application/json; charset=utf-8` |

By changing the default to `application/json`, the prior default, then you get these results:

| Accept request header | Content-Type response header |
|-|-|
| (none) | `application/json` |
| `*/*` | `application/json` |
| `application/*` | `application/json` |
| `application/*+json` | `application/graphql-response+json; charset=utf-8` |
| `application/graphql-response+json` | `application/graphql-response+json; charset=utf-8` |
| `application/graphql+json` | `application/graphql+json; charset=utf-8` |
| `application/json` | `application/json` |
| `*/*, application/json` | `application/json` |

tests/ApiApprovalTests/net5+netcoreapp31/GraphQL.Server.Transports.AspNetCore.approved.txt
```diff
@@ -109,7 +110,6 @@ namespace GraphQL.Server.Transports.AspNetCore
         public bool ReadExtensionsFromQueryString { get; set; }
         public bool ReadQueryStringOnPost { get; set; }
         public bool ReadVariablesFromQueryString { get; set; }
-        public string ResponseContentType { get; set; }
```
Shane32, line 112: Not a breaking change; this property has not been released yet.

jeffw-wherethebitsroam:
Hi @Shane32. Busy weekend, so didn't get a chance to look before now, but it looks good to me.

bithavoc:
Thanks so much, this fixed an issue while trying to consume from graphql-request where it always expects `application/json`

tests/ApiApprovalTests/net5+netcoreapp31/GraphQL.Server.Transports.AspNetCore.approved.txt
sungam3r: ok

tests/Transports.AspNetCore.Tests/Middleware/GetTests.cs
```diff
@@ -65,6 +65,103 @@ public async Task BasicTest()
...
+    [InlineData("application/json", "application/*, application/graphql-response+json", "application/graphql-response+json; charset=utf-8")]
+    [InlineData("application/json", "application/*, application/graphql+json", "application/graphql+json; charset=utf-8")]
+    [InlineData("application/json", "application/*, application/json", "application/json")]
+    [InlineData("application/json", "application/*, application/pdf", "application/json")]
+    [InlineData("application/json", "application/*+json, application/graphql-response+json", "application/graphql-response+json; charset=utf-8")]
+    [InlineData("application/json", "application/*+json, application/graphql+json", "application/graphql+json; charset=utf-8")]
+    [InlineData("application/json", "application/*+json, application/json", "application/json")]
+    [InlineData("application/json", "application/*+json, application/pdf", "application/graphql-response+json; charset=utf-8")]
+    [InlineData("application/json", "application/graphql+json, application/json", "application/graphql+json; charset=utf-8")]
+    public async Task AcceptHeaderHonored(string? defaultMediaType, string mediaType, string expected)
```
sungam3r, line 150: wooooooow

Shane32: You know I like my `[Theory]`s...

tests/Transports.AspNetCore.Tests/Middleware/GetTests.cs
```diff
@@ -65,6 +65,103 @@ public async Task BasicTest()
...
+    [InlineData(null, "*/*, application/json", "application/json; charset=utf-8")]
+    [InlineData(null, "*/*, application/pdf", "application/graphql-response+json; charset=utf-8")]
+    [InlineData(null, "application/*, application/graphql-response+json", "application/graphql-response+json; charset=utf-8")]
+    [InlineData(null, "application/*, application/graphql+json", "application/graphql+json; charset=utf-8")]
+    [InlineData(null, "application/*, application/json", "application/json; charset=utf-8")]
+    [InlineData(null, "application/*, application/pdf", "application/graphql-response+json; charset=utf-8")]
+    [InlineData(null, "application/*+json, application/graphql-response+json", "application/graphql-response+json; charset=utf-8")]
+    [InlineData(null, "application/*+json, application/graphql+json", "application/graphql+json; charset=utf-8")]
+    [InlineData(null, "application/*+json, application/json", "application/json; charset=utf-8")]
+    [InlineData(null, "application/*+json, application/pdf", "application/graphql-response+json; charset=utf-8")]
```
sungam3r, line 135: I think thre should be test for "application/pdf" accept header when server returns error. From https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.2:
изображение

src/Transports.AspNetCore/GraphQLHttpMiddlewareOptions.cs
```diff
@@ -95,7 +97,8 @@ public class GraphQLHttpMiddlewareOptions : IAuthorizationOptions
     public GraphQLWebSocketOptions WebSockets { get; set; } = new();
 
     /// <summary>
-    /// The Content-Type to use for GraphQL responses
+    /// The Content-Type to use for GraphQL responses, if it matches the 'Accept'
+    /// HTTP request header. Defaults to "application/graphql-response+json; charset=utf-8".
     /// </summary>
-    public string ResponseContentType { get; set; } = GraphQLHttpMiddleware.CONTENTTYPE_GRAPHQLJSON;
+    public MediaTypeHeaderValueMs DefaultResponseContentType { get; set; } = MediaTypeHeaderValueMs.Parse(GraphQLHttpMiddleware.CONTENTTYPE_GRAPHQLRESPONSEJSON);
```
sungam3r, line 103: I suggest to move CONTENTTYPE_GRAPHQLRESPONSEJSON into options class.

src/Transports.AspNetCore/GraphQLHttpMiddlewareOptions.cs
```diff
@@ -95,7 +97,8 @@ public class GraphQLHttpMiddlewareOptions : IAuthorizationOptions
     public GraphQLWebSocketOptions WebSockets { get; set; } = new();
 
     /// <summary>
-    /// The Content-Type to use for GraphQL responses
+    /// The Content-Type to use for GraphQL responses, if it matches the 'Accept'
```
sungam3r, line 100: I read it several times and this comment still seems strange to me. I don't understand the meaning of _matches_.

tests/Transports.AspNetCore.Tests/Middleware/GetTests.cs
Shane32: According to the draft spec (as of today; it's a moving target!):

> In alignment with the [HTTP 1.1 Accept](https://tools.ietf.org/html/rfc7231#section-5.3.2) specification, when a client does not include at least one supported media type in the Accept HTTP header, the server MUST either:
> 
> - Disregard the Accept header and respond with the default media type of `application/json`, specifying this in the Content-Type header; OR
> - Respond with a 406 Not Acceptable status code and stop processing the request.

I elected the first option.  I would agree that the latter would be better, but I was thinking this change would come in 8.0, as it would involve processing the header earlier in the flow and so would change the design of the middleware.

We could also support additional character encodings; we do so for requests -- why not responses also?  I have no urgency to do so; you're the only one I know of that mentioned using charsets besides utf-8 -- just an idea.

src/Transports.AspNetCore/GraphQLHttpMiddlewareOptions.cs
Shane32: If you wish to make a PR in this regard, please make it separately from other of your comments here - thanks!  As of now, I'm not sure why this would make sense to do, given that this and a number of other constants are required in the middleware class.  We should keep all of the content type / media type constants together.

src/Transports.AspNetCore/GraphQLHttpMiddlewareOptions.cs
```diff
@@ -1,3 +1,5 @@
+using MediaTypeHeaderValueMs = Microsoft.Net.Http.Headers.MediaTypeHeaderValue;
```
sungam3r, line 1: Just to shorten line 103 ?

src/Transports.AspNetCore/GraphQLHttpMiddlewareOptions.cs
sungam3r: Nope. Now I see a bunch of other constants in GraphQLHttpMiddleware.

src/Transports.AspNetCore/GraphQLHttpMiddlewareOptions.cs
Shane32: There are two different `MediaTypeHeaderValue` classes both within scope.

src/Transports.AspNetCore/GraphQLHttpMiddlewareOptions.cs
sungam3r: I mean in generally I understand what you want to say but from the point of view of some "external" developer it may look strange.

docs/migration/migration7.md
```diff
@@ -188,27 +188,17 @@ app.UseGraphQL<MySchema>("/graphqlsubscription", o => {
...
-    protected override string SelectResponseContentType(HttpContext context)
-        => "application/json";
-}
-
-app.UseGraphQL<MyMiddleware<ISchema>>("/graphql", new GraphQLHttpMiddlewareOptions());
+// with utf-8 charset specified
+app.UseGraphQL("/graphql", o => o.DefaultResponseContentType = new("application/json") { Charset = "utf-8" });
 ```
 
+Note that if a request is received with a specific supported media type such as `application/graphql-response+json`,
```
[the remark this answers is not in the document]
Shane32: I'd add "the", and a comma:

in the `Accept` header, such as

src/Transports.AspNetCore/GraphQLHttpMiddlewareOptions.cs
Shane32: I do the best I can.  Sometimes I'm tongue tied.  Here it seemed to make sense to me, but I get it.  Feel free to suggest an improvement.

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -1,5 +1,8 @@
 #pragma warning disable CA1716 // Identifiers should not match keywords
 
+using Microsoft.Extensions.Primitives;
+using MediaTypeHeaderValueMs = Microsoft.Net.Http.Headers.MediaTypeHeaderValue;
```
sungam3r, line 4: Why alias?

Shane32: There are two different `MediaTypeHeaderValue` classes both within scope (if I recall correctly).

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -444,17 +449,192 @@ protected virtual async Task<ExecutionResult> ExecuteRequestAsync(HttpContext co
...
+            // give priority to specific types, then to types with wildcards
+            foreach (var acceptHeader in acceptHeaders.OrderBy(x => x.MatchesAllTypes ? 4 : x.MatchesAllSubTypes ? 3 : x.MatchesAllSubTypesWithoutSuffix ? 2 : 1))
+            {
+                var response = CheckForMatch(acceptHeader);
+                if (response != null)
+                    return response;
+            }
+        }
+
+        // return the default content type if no match is found, or if there is no 'Accept' header
```
sungam3r, line 490: Hmm, what if client specified "application/pdf" or "application/some_unknown_mime" in accept header?

From https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.2
> A request without any Accept header field implies that the user agent
   will accept any media type in response.  If the header field is
   present in a request and none of the available representations for
   the response have a media type that is listed as acceptable, the
   origin server can either honor the header field by sending a 406 (Not
   Acceptable) response **or disregard the header field by treating the
   response as if it is not subject to content negotiation.**


Do you want to go with second option?

Shane32: That’s the case now (merged in this PR). Please read my response to your comments on this earlier today. TLDR: maybe update in v8 to reject with 406

src/Transports.AspNetCore/GraphQLHttpMiddleware.cs
```diff
@@ -444,17 +449,192 @@ protected virtual async Task<ExecutionResult> ExecuteRequestAsync(HttpContext co
...
+        return _options.DefaultResponseContentType.ToString();
```
sungam3r, line 491:
```
public override string ToString()
        {
            StringBuilder stringBuilder = new StringBuilder();
            stringBuilder.Append(_mediaType);
            NameValueHeaderValue.ToString(_parameters, ';', leadingSeparator: true, stringBuilder);
            return stringBuilder.ToString();
        }
```
😕 called each time

sungam3r:
Reviewed. LGTM. I left some comments and posted #931

tests/Transports.AspNetCore.Tests/Middleware/GetTests.cs
sungam3r: agree

sungam3r: #932

sungam3r: Regarding encodings - spec become rather stric about encoding:
изображение

MUST is strong requirement )

docs/migration/migration7.md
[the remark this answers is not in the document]
sungam3r: #933

tests/Transports.AspNetCore.Tests/Middleware/GetTests.cs
Shane32: https://github.com/graphql/graphql-over-http/issues/229

PR #966, opened 2022-12-07
Add Azure Functions support

Shane32:
This PR adds support for Azure Functions.  To use:

1. Configure the Azure Function to use DI - see https://learn.microsoft.com/en-us/azure/azure-functions/functions-dotnet-dependency-injection
2. Configure GraphQL via `AddGraphQL` the same way as is done elsewhere
3. Add a call to `.AddAzureFunctionsMiddleware()` within the `AddGraphQL` call
4. Add a HTTP function that returns an appropriate `ActionResult`:

```cs
    [FunctionName("GraphQL")]
    public static IActionResult RunGraphQL(
        [HttpTrigger(AuthorizationLevel.Anonymous, "get", "post", Route = null)] HttpRequest req,
        ILogger log)
    {
        log.LogInformation("C# HTTP trigger function processed a request.");

        return new AzureGraphQLActionResult(req);
    }
```

UI packages add an `ActionResult` to easily use them with controller projects or Azure Functions.

```cs
    [FunctionName("Playground")]
    public static IActionResult RunPlayground(
        [HttpTrigger(AuthorizationLevel.Anonymous, "get")] HttpRequest req,
        ILogger log)
    {
        _ = req;
        log.LogInformation("Getting Playground UI.");

        return new PlaygroundActionResult(opts => opts.GraphQLEndPoint = "/api/graphql");
    }
```

Notes
- The sample is based on the .NET template, and I think I read it has to be 6.0 becuase that's a LTS version (??)
- I didn't actually try this in the cloud
- Multiple schemas are supported by using typed calls - `.AddAzureFunctionsMiddleware<T>()` and `return new AzureGraphQLActionResult<T>()`.
- The middleware can have options configured on it via `.AddAzureFunctionsMiddleware<T>(opts => { });`
- It is currently not possible with the new helper functions to have two different endpoints with different configurations for the same schema.
- No breaking changes
- UI packages add a dependency when using them with ASP.NET Core 2.x -- but no new dependency for ASP.NET Core 3.1+

Comments / questions:
- Technically the new `AzureGraphQLActionResult` is not specific to Azure Functions.  It _could_ be used for ASP.NET Core also, but it is harder to use and inferior to the middleware technique we already support.  In any case, we could drop the "Azure Functions" naming from the new classes/interfaces and call it something like `GraphQLExecutionActionResult`
- We could eliminate the need to register the middleware in DI, similar to how the UI action results work, but then there is a bit of startup overhead for each request.  Not sure if this occurs anyway for azure functions or not, or which would be a better approach.

Todo:
- [x] Peer review
- [x] Add similar ActionResult classes to the other UI projects
- [x] Add documentation in the readme for usage within Azure Functions
- [x] Add tests to ensure that the new sample application works correctly


src/Ui.Playground/PlaygroundActionResult.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// <summary>
+    /// Initializes the playground action result with the specified options.
+    /// </summary>
+    public PlaygroundActionResult(PlaygroundOptions options)
+    {
+        _middleware = new(_ => Task.CompletedTask, options);
+    }
+
+    /// <summary>
+    /// Initializes the playground action result with the specified option configuration delegate.
```
sungam3r, line 21:
```suggestion
    /// Initializes the playground action result with the specified optional configuration delegate.
```

src/Ui.GraphiQL/GraphiQLActionResult.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// <summary>
+    /// Initializes the playground action result with the specified options.
+    /// </summary>
+    public GraphiQLActionResult(GraphiQLOptions options)
+    {
+        _middleware = new(_ => Task.CompletedTask, options);
+    }
+
+    /// <summary>
+    /// Initializes the playground action result with the specified option configuration delegate.
```
sungam3r, line 21:
```suggestion
    /// Initializes the playground action result with the specified optional configuration delegate.
```

src/Ui.GraphiQL/GraphiQLActionResult.cs
```diff
@@ -0,0 +1,33 @@
+using Microsoft.AspNetCore.Mvc;
+
+namespace GraphQL.Server.Ui.GraphiQL;
+
+/// <summary>
+/// An action result that returns a GraphiQL user interface
```
sungam3r, line 6:
```suggestion
/// An action result that returns a GraphiQL user interface.
```

src/Transports.AspNetCore/AzureFunctions/AzureGraphQLActionResult.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// <param name="httpRequest"></param>
+    public AzureGraphQLActionResult(HttpRequest httpRequest)
+    {
+        _httpRequest = httpRequest;
+        _middleware = httpRequest.HttpContext.RequestServices.GetRequiredService<IAzureGraphQLMiddleware<TSchema>>();
+    }
+
+    /// <inheritdoc/>
+    public virtual Task ExecuteResultAsync(ActionContext context)
+        => _middleware.InvokeAsync(_httpRequest, static httpContext =>
```
sungam3r, line 24: Why do you need to pass `next` delagate at all?

src/Transports.AspNetCore/Extensions/GraphQLBuilderExtensions.cs
```diff
@@ -143,4 +145,56 @@ public static IGraphQLBuilder AddAuthorizationRule(this IGraphQLBuilder builder)
...
+            DI.ServiceLifetime.Singleton);
+
+        builder.Services.TryRegister<IAzureGraphQLMiddleware>(
+            provider => provider.GetRequiredService<IAzureGraphQLMiddleware<TSchema>>(),
+            DI.ServiceLifetime.Singleton);
+
+        return builder;
+    }
+
+    private class AzureGraphQLMiddlewareMapper<TSchema> : IAzureGraphQLMiddleware
```
sungam3r, line 188: What is it?

samples/Samples.AzureFunctions/GraphQL.cs
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    [FunctionName("Playground")]
+    public static IActionResult RunPlayground(
+        [HttpTrigger(AuthorizationLevel.Anonymous, "get")] HttpRequest request,
+        ILogger log)
+    {
+        log.LogInformation("C# HTTP trigger function processed a request for the GraphQL Playground UI.");
+
+        return new PlaygroundActionResult(opts => opts.GraphQLEndPoint = "/api/graphql");
```
sungam3r, line 32: > /api/graphql

Where is it configured?

Shane32: I don't know much about Azure Functions.  I created the sample and renamed the `[FunctionName]` and that's the url it was at.  It appears in the console app when you start debugging.

src/Transports.AspNetCore/AzureFunctions/AzureGraphQLActionResult.cs
Shane32: The `GraphQLHttpMiddleware` is designed around having a `next` delegate available if, for example, GET requests are disabled.  In this case the ActionResult returns a NotFound message, similar to the default behavior of ASP.NET Core.

samples/Samples.AzureFunctions/GraphQL.cs
sungam3r: I mean I did not find the code that starts GraphQL endpoint at `/api/graphql`.

Shane32: Please see line 53 and consult with Azure Functions documentation here:

https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-http-webhook-trigger?tabs=in-process%2Cfunctionsv2&pivots=programming-language-csharp#customize-the-http-endpoint

src/Transports.AspNetCore/AzureFunctions/AzureGraphQLActionResult.cs
sungam3r: https://github.com/dotnet/aspnetcore/blob/1ef289a7589b3b095d0015b55b47aebd8e9667fd/src/Http/Http/src/Builder/ApplicationBuilder.cs#L135

Shane32: These are Azure Functions, and does not use the ASP.NET Core pipeline.  There is no IApplicationBuilder.

Shane32: Even within ASP.NET Core, the code you pointed to only occurs if the end of the pipeline is reached.  An `ActionResult` is an end of a pipeline, and ASP.NET Core considers the request to be handled at that point.  It would not reach the code you pointed to.

Shane32: Notice the signature: `Task ExecuteResultAsync(ActionContext context)`.  There is no `next` delegate.  If there were, this would contain the code you linked to.  But there isn't.  The code runs here, and ASP.NET Core or Azure Functions considers the request handled.  That is why while mapping onto our middleware, I provide the middleware with the code to set a 404 response -- in case for some reason the middleware doesn't handle the request, it will respond with 404.

Shane32: > These are Azure Functions, and does not use the ASP.NET Core pipeline. There is no IApplicationBuilder.

Well, it does have a number of dependencies on ASP.NET Core 2.x libraries.  I honestly do not know how much or how little it uses ASP.NET Core functionality.

src/Transports.AspNetCore/Extensions/GraphQLBuilderExtensions.cs
sungam3r: I do not see usages of private class.

samples/Samples.AzureFunctions/GraphQL.cs
sungam3r: Ah! )

sungam3r, line 32:
```suggestion
        return new PlaygroundActionResult(opts => opts.GraphQLEndPoint = "/api/graphql"); // /api/graphql route will call RunGraphQL method
```

src/Transports.AspNetCore/Extensions/GraphQLBuilderExtensions.cs
Shane32: You're right.  I must have already decided it wasn't worth the trouble.  I'll remove the class.

src/Transports.AspNetCore/AzureFunctions/AzureGraphQLActionResult.cs
sungam3r: ok

samples/Samples.AzureFunctions/GraphQL.cs
```diff
@@ -0,0 +1,34 @@
+#pragma warning disable IDE0060 // Remove unused parameter
```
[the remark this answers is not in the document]
Shane32: Moved to csproj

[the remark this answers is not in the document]
sungam3r: ok

sungam3r:
> Technically the new AzureGraphQLActionResult is not specific to Azure Functions. 

I would keep naming as is to accent on Azure applicability.

sungam3r:
> We could eliminate the need to register the middleware in DI, similar to how the UI action results work...

50/50 here, not sure. `IAzureGraphQLMiddleware` interface looks a bit odd.

Shane32:
> > We could eliminate the need to register the middleware in DI, similar to how the UI action results work...
> 
> 50/50 here, not sure. `IAzureGraphQLMiddleware` interface looks a bit odd.

I'll do a separate PR after merging this so we can both review the alternative design.  I'm 50/50 also.

Shane32:
@sungam3r I added tests, documentation, and etc.  Ready to complete review and merge.

README.md
```diff
@@ -196,6 +196,57 @@ public class HomeController : Controller
...
+configure subscription support, as Azure Functions do not support WebSockets since it is
+a serverless environment.
+
+See the `Samples.AzureFunctions` project for a complete sample based on the
+.NET template for Azure Functions.
+
+Please note that the GraphQL schema needs to be initialized for every call through
+Azure Functions, since it is a serverless environment.  This is done automatically
+but will come at a performance cost.  If you are using a schema that is expensive
+to initialize, you may want to consider using a different hosting environment.
```
sungam3r, line 248: I suspected that the schema should be created every time so your note about overhead here
> We could eliminate the need to register the middleware in DI, similar to how the UI action results work, but then there is a bit of startup overhead for each request. Not sure if this occurs anyway for azure functions or not, or which would be a better approach.

will not affect the overall costs.

Shane32: I was thinking the same thing.  I think we will find that the alternative approach for Azure Functions is better overall.

Shane32: It's possible that if multiple calls are made in succession maybe then the schema would not need to be reinitialized.

But regardless, I looked at GraphQLHttpMiddleware and it seems that we can make some private readonly fields static anyway.  At which point, initializing one class should have little effect on speed/memory.

sungam3r: It will have any effect only if Azure Function environment does not unload entirely the whole app (dll/domain/process/whatever) from memory.

sungam3r (approved):

PR #968, opened 2022-12-09
Simplify Azure Functions support

Shane32:
I like these changes better

It's easier to use, likely has very little impact on performance, and allows multiple configurations of the middleware for each instance.  It also moves the options configuration to a better location in the code, meaning that they are defined where used, rather than in DI setup.  And it's overall less code.

Also in this PR:
- Rename to `GraphQLExecutionActionResult` and eliminate dedicated namespace, matching UI projects
- Update controller sample to demonstrate using `GraphQLExecutionActionResult`
- Update readme
- Revert Azure Functions sample to .NET 6 -- it does not work with .NET 7 on my development PC (although it builds and tests run fine)


src/Transports.AspNetCore/GraphQLActionResult.cs
```diff
@@ -0,0 +1,87 @@
+namespace GraphQL.Server.Transports.AspNetCore;
+
+/// <summary>
+/// An <see cref="IActionResult"/> that executes a GraphQL request for the specified schema.
+/// </summary>
+public class GraphQLExecutionActionResult<TSchema> : IActionResult
```
sungam3r, line 6: File name differs.

sungam3r (approved):

PR #972, opened 2022-12-13
Hide logo and readme in Solution Explorer

sungam3r:


Shane32 (approved):

sungam3r:
Hah! Fresh errors from fresh .NET SDK!

sungam3r:
https://github.com/dotnet/roslyn-analyzers/blob/main/src/NetAnalyzers/Microsoft.CodeAnalysis.NetAnalyzers.md#ca1852-seal-internal-types

sungam3r:
The question is why local and CI builds differ.

samples/Samples.Jwt/JwtHelper.cs
```diff
@@ -85,8 +85,7 @@ private static (SecurityKey SecurityKey, string SecurityAlgorithm) CreateSymmetr
     {
         // hash the password and use that to create a symmetric key for signing the JWT tokens
         var passwordBytes = System.Text.Encoding.UTF8.GetBytes(password);
-        using var sha = SHA256.Create();
-        var keyBytes = sha.ComputeHash(passwordBytes);
+        var keyBytes = SHA256.HashData(passwordBytes);
```
sungam3r, line 88: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1850

sungam3r:
изображение

Shane32:
Recently it seemed that .NET Analyzers were broken to me in .NET 7.  I had to set `<AnalysisLevel>6-recommended</AnalysisLevel>` to get VS to make the proper recommendations in some of my repos.

samples/Samples.Jwt/JwtHelper.cs
Shane32: Ok, although note that this is only valid in .NET 5+.  Works here because this is a sample that targets .NET 7.  I'm okay with the change.

tests/Transports.AspNetCore.Tests/AuthorizationHelperTests.cs
```diff
@@ -3,6 +3,8 @@
 
 namespace Tests;
 
+#pragma warning disable CS1998
```
Shane32, line 6: This should not be necessary.  Test.props has

`<WarningsNotAsErrors>$(WarningsNotAsErrors);CS1998</WarningsNotAsErrors>`

Perhaps we need to change something else, such as removing `-severity warn` from dotnet format.

Shane32: Ok with all other changes posted so far

Shane32: Or we should just switch it to `<NoWarn>$(NoWarn);CS1998</NoWarn>`.  The warning certainly does not affect tests.

sungam3r:
My bad. I was builing on on 7.0.100 SDK. With 7.0.101 errors are in place:
изображение

tests/Transports.AspNetCore.Tests/AuthorizationHelperTests.cs
sungam3r: I moved it to nowarns.

Shane32:
Maybe they fixed all the analysis stuff that I was having trouble with in 101. Let's hope so anyway.

PR #977, opened 2022-12-17
Catch additional exception types when a WebSocket connection is disconnected

Shane32:
ASP.NET Core can dispose a `WebSocket` instance internally, leading to a `ObjectDisposedException` when a client disconnects.

With this PR, exceptions are rethrown as `OperationCanceledException`s.

### DIFF
--- a/tests/Transports.AspNetCore.Tests/WebSockets/WebSocketConnectionTests.cs
+++ b/tests/Transports.AspNetCore.Tests/WebSockets/WebSocketConnectionTests.cs
@@ -493,7 +493,7 @@ public class WebSocketConnectionTests : IDisposable
     }
 
     [Fact]
-    public async Task ExecuteAsync_WebSocket_EatsWebSocketExceptions()
+    public async Task ExecuteAsync_WebSocket_WebSocketExceptions_Rethrows()
     {
         _mockConnection.CallBase = true;
         var mockReceiveStream = new Mock<IOperationMessageProcessor>(MockBehavior.Strict);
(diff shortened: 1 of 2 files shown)


tests/Transports.AspNetCore.Tests/WebSockets/WebSocketConnectionTests.cs
```diff
@@ -511,7 +511,29 @@ public async Task ExecuteAsync_WebSocket_EatsWebSocketExceptions()
                 throw new WebSocketException();
             })
             .Verifiable();
-        await _connection.ExecuteAsync(mockReceiveStream.Object);
+        await Should.ThrowAsync<OperationCanceledException>(() => _connection.ExecuteAsync(mockReceiveStream.Object));
```
sungam3r, line 514: Test name contradicts with this line. No?

sungam3r: I mean test did not throw before and it throws after.

Shane32: True. Now it rethrows as OCE

Shane32: Fixed (the name of the test)

Shane32: FYI OCEs were typically thrown before upon disconnection which were not caught. Once in a while a WebSocketException was thrown instead, which was caught. So it more of a bug fix to rethrow the WebSocketException as an OCE, not a breaking change.

sungam3r (approved):

PR #981, opened 2023-01-06
Add net6.0 target framework to Transports.AspNetCore

Oceania2018:


Shane32:
Unless there is additional code that can be optimized, adding another TFM serves no purpose and simply increases the package size.  .NET 6.0 will use the .NET 5.0 TFM which already exists.

PR #985, opened 2023-01-20
Fix rare memory 'leak' with WhenAll when it is combined with Task.Delay

Shane32:
## Bug description

When calling the code:

```cs
await Task.WhenAny(task, Task.Delay(timeout, token));
```

and when `task` completes successfully, the `Task.Delay` is still active until the timeout expires or the cancellation token is signaled.  If the timeout is infinite, it is registered to the cancellation token's source until it is signaled. Or if the token is never signaled, then it lives until the timeout expires. 

## Correct operation

The `Task.Delay` should be "disposed" when `task` completes.  This is provided in .NET 6 by the Task extension method `.WaitAsync`.

## Impact

Using code like this in a tight loop would be very bad, but fortunately we only have one such use, and these are the mitigating factors:

1. Only occurs when a websocket connection is gracefully disconnected
2. Timeout is 10 seconds with default settings, not infinite, so after 10 seconds the timer would be released.
3. Assuming the connection closes normally, the cancellation token will be signaled almost immediately.

## Resolution

Add .NET 6 TFM and create an extension method to polyfill for prior TFMs.  The provided polyfill is perfectly effective although it is not nearly as efficient as the .NET 6 method.  Note that the .NET 6 source uses internal classes for the implementation so it is not practical to copy and paste the source. 

## See also

- https://andrewlock.net/a-deep-dive-into-the-new-task-waitasync-api-in-dotnet-6/
- https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task-1.waitasync?view=net-7.0
- https://github.com/Shane32/AsyncResetEvents/pull/3


src/Transports.AspNetCore/Extensions/TaskExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+            return Task.FromCanceled<TResult>(cancellationToken);
+
+        return TimeoutAfter(task, millisecondsTimeout, cancellationToken);
+
+        static async Task<TResult> TimeoutAfter(Task<TResult> task, int millisecondsDelay, CancellationToken cancellationToken)
+        {
+            // the 'using' here ensures that the Task.Delay gets disposed if the task finishes before the delay
+            using (var timeoutCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken))
+            {
+
```
sungam3r, line 31:
```suggestion
```

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
```diff
@@ -123,9 +123,11 @@ public virtual async Task ExecuteAsync(IOperationMessageProcessor operationMessa
...
                         _ = CloseAsync();
                         // wait until the close has been sent
-                        await Task.WhenAny(
-                            _outputClosed.Task,
-                            Task.Delay(_closeTimeout, RequestAborted));
+                        try
+                        {
+                            await _outputClosed.Task.WaitAsync(_closeTimeout, RequestAborted);
+                        }
+                        catch (TimeoutException) { }
```
sungam3r, line 130: 🤔 You catch only `TimeoutException` here since itis the only new one exception appeared but one more thing has changed. Old code throws `TaskCanceledException`, new code throws `OperationCanceledException` (base). Is it OK to consumers with `(catch TaskCanceledException)`? Should they use `(catch OperationCanceledException)` all the time?

src/Transports.AspNetCore/Extensions/TaskExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+            return Task.FromException<TResult>(new TimeoutException());
+        if (cancellationToken.IsCancellationRequested)
+            return Task.FromCanceled<TResult>(cancellationToken);
+
+        return TimeoutAfter(task, millisecondsTimeout, cancellationToken);
+
+        static async Task<TResult> TimeoutAfter(Task<TResult> task, int millisecondsDelay, CancellationToken cancellationToken)
+        {
+            // the 'using' here ensures that the Task.Delay gets disposed if the task finishes before the delay
+            using (var timeoutCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken))
```
sungam3r, line 29:
```suggestion
            using var timeoutCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
```

I would eliminate one level of indentation.

src/Transports.AspNetCore/WebSockets/WebSocketConnection.cs
Shane32: > Should they use `(catch OperationCanceledException)` all the time?

Absolutely.  While it may be that a task throws a TCE, methods like `CancellationToken.ThrowIfCancellation` throw a OCE.  So in my opinion, users should never `catch (TaskCanceledException)` and always catch the OCE.

PR #998, opened 2023-03-01
Revert to old version of setup-dotnet

Shane32:


Shane32:
@sungam3r Check out the workflow; notice that with `actions/setup-dotnet@v1` the correct version of the .NET SDK (version 7.0.201) is being installed whereas with `actions/setup-dotnet@v3` the wrong version gets installed.

sungam3r (approved):

PR #1001, opened 2023-03-02
Fix CI

Shane32:
Samples.AzureFunctions requires the .NET 6 SDK.  Since the workflows build the entire solution, it is required, even though we are not packing those projects.


sungam3r:
Why didn't we have a build error?

Shane32:
The test workflow already installs the other SDKs

Shane32:
Here is the build error from the last merge:

https://github.com/graphql-dotnet/server/actions/runs/4309168545

sungam3r (approved):

PR #1009, opened 2023-03-20
Misc updates for the JWT sample application

Shane32:
- Pull `TokenValidationParameters` from the call to `AddJwtBearer` so that the reference to the static property `JwtHelper.Instance` is not necessary
- Eliminate the static property `JwtHelper.Instance`
- Reconfigure `Program.cs` to use a local variable instead of a static property
- Reconfigure `OAuthController` to pull `JwtHelper` from DI
- Add the 'iat' claim to generated tokens (required for Open ID Connect)
- Bump Microsoft.AspNetCore.Authentication.JwtBearer since the project is a .NET 7 sample

### DIFF
--- a/samples/Samples.Jwt/JwtWebSocketAuthenticationService.cs
+++ b/samples/Samples.Jwt/JwtWebSocketAuthenticationService.cs
@@ -50,8 +50,9 @@ public class JwtWebSocketAuthenticationService : IWebSocketAuthenticationService
             {
                 // pull the token from the value
                 var token = authPayload.Authorization.Substring(7);
-                // parse the token in the same manner that the .NET AddJwtBearer() method does
-                // note that JwtSecurityTokenHandler maps the 'name' and 'role' claims to the 'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name' and 'http://schemas.microsoft.com/ws/2008/06/identity/claims/role' claims
+                // parse the token in the same manner that the .NET AddJwtBearer() method does:
+                // JwtSecurityTokenHandler maps the 'name' and 'role' claims to the 'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name'
+                // and 'http://schemas.microsoft.com/ws/2008/06/identity/claims/role' claims;
                 // this mapping is not performed by Microsoft.IdentityModel.JsonWebTokens.JsonWebTokenHandler
                 var handler = new System.IdentityModel.Tokens.Jwt.JwtSecurityTokenHandler();
                 var tokenValidationParameters = _jwtBearerOptionsMonitor.Get(JwtBearerDefaults.AuthenticationScheme).TokenValidationParameters;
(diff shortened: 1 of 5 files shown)


Shane32:
As a side note, I would prefer to add another nuget package for JWT handling to this repository, specifically for the `IWebSocketAuthenticationService` implementation (which does not require `JwtHelper` to be useful).  I cannot put it in the main package because it requires another dependency.  However, this makes an assumption about the format of the websocket initialization packet payload, which may be different from user to user as there is no official standard.  It also only works for JWT, and not any other authentication mechanism.  So as of now I think we should just leave it as a sample application.  I plan to keep this sample code up to date as best I can so users may copy it if they wish into their own application.

samples/Samples.Jwt/JwtWebSocketAuthenticationService.cs
```diff
@@ -10,15 +11,29 @@ namespace GraphQL.Server.Samples.Jwt;
 /// This is necessary because WebSocket connections initiated from the browser cannot
 /// authenticate via HTTP headers.
 /// <br/><br/>
-/// This class is not used when authenticating over GET/POST.
+/// Notes:
+/// <list type="bullet">
+/// <item>This class is not used when authenticating over GET/POST.</item>
+/// <item>
+/// This class pulls the <see cref="TokenValidationParameters"/> instance from the instance of
+/// <see cref="IOptions{TOptions}">IOptions</see>&lt;<see cref="JwtBearerOptions"/>&gt; registered
```
sungam3r, line 19:
```suggestion
/// <see cref="IOptionsMonitor{TOptions}">IOptionsMonitor</see>&lt;<see cref="JwtBearerOptions"/>&gt; registered
```

samples/Samples.Jwt/JwtWebSocketAuthenticationService.cs
```diff
@@ -29,25 +44,21 @@ public Task AuthenticateAsync(IWebSocketConnection connection, string subProtoco
...
-                    // set the ClaimsPrincipal for the HttpContext; authentication will take place against this object
-                    connection.HttpContext.User = principal;
-                }
+                // parse the token in the same manner that the .NET AddJwtBearer() method does:
+                // JwtSecurityTokenHandler maps the 'name' and 'role' claims to the 'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name'
+                // and 'http://schemas.microsoft.com/ws/2008/06/identity/claims/role' claims;
+                // this mapping is not performed by Microsoft.IdentityModel.JsonWebTokens.JsonWebTokenHandler
+                var handler = new System.IdentityModel.Tokens.Jwt.JwtSecurityTokenHandler();
+                var tokenValidationParameters = _jwtBearerOptionsMonitor.Get(JwtBearerDefaults.AuthenticationScheme).TokenValidationParameters;
+                var principal = handler.ValidateToken(token, tokenValidationParameters, out var securityToken);
```
sungam3r, line 59: No `TokenValidationResult` anymore? Does `ValidateToken` throw in case of any error?

sungam3r (approved):

samples/Samples.Jwt/JwtWebSocketAuthenticationService.cs
Shane32: Correct; it throws in the case that the token is invalid for any reason (according to the parameters defined in `tokenValidationParameters`). So for example, if `TokenValidationParameters.ValidateIssuer` is `true`, then the issuer must match or an exception is thrown.

PR #1017, opened 2023-04-12
Allow option to disable handling of form data for POST requests

Shane32:
There is a CSRF security vulnerability that occurs when:
- Cookie authentication is in use
- The user's web browser is logged in to the vulnerable server
- The foreign origin executes a mutation against the vulnerable server (can be done via the fetch API) and includes credentials
- The data is sent via `multipart/form-data` so there is no CORS pre-flight check performed

I have personally verified that this CSRF attack can be exploited when cookie authentication is in use.  [This branch](https://github.com/graphql-dotnet/server/tree/d75a9e17002efddc33afea84b27ff25733c14fa5) has a sample project that will successfully execute a cross-origin mutation against the Samples.Authentication project, even though the response fails CORS validation and so the response is not readable in the browser.

Sample javascript to perform the attack:

```javascript

    async function executeGraphQLRequest(serverName, query) {
        const formData = new FormData();
        formData.append('query', query);
        const response = await fetch(`https://${serverName}/graphql`, {
            method: 'POST',
            body: formData,
            credentials: 'include', // Set the credentials option to "include"
        });
        const jsonResponse = await response.json();
        return jsonResponse;
    }
    async function runme() {
        const serverName = 'localhost:7110';
        const query = '{ helloRegisteredUser }';
        const jsonResponse = await executeGraphQLRequest(serverName, query);
        console.log(jsonResponse);
    }
```

See notes added to readme

We should set the new property to `false` for v9

### DIFF
--- a/README.md
+++ b/README.md
@@ -617,6 +617,7 @@ methods allowing for different options for each configured endpoint.
 | `HandlePost`                       | Enables handling of POST requests. | True |
 | `HandleWebSockets`                 | Enables handling of WebSockets requests. | True |
 | `ReadExtensionsFromQueryString`    | Enables reading extensions from the query string. | True |
+| `ReadFormOnPost`                   | Enables parsing of form data for POST requests (may have security implications). | True |
 | `ReadQueryStringOnPost`            | Enables parsing the query string on POST requests. | True |
 | `ReadVariablesFromQueryString`     | Enables reading variables from the query string. | True |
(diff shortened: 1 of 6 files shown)


Shane32:
See also:
- https://github.com/jaydenseric/graphql-multipart-request-spec/pull/64

sungam3r (approved):

PR #1018, opened 2023-04-14
Bump build number of GraphQL version

Shane32:
- Gets the latest bug fixes for subscription handling

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <VersionPrefix>7.3.0-preview</VersionPrefix>
+    <VersionPrefix>7.4.0-preview</VersionPrefix>
     <NextVersion>8.0.0</NextVersion>
     <LangVersion>latest</LangVersion>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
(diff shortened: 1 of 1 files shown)


Shane32:
Will release this as 7.4.0 since there is an API change -- ok @sungam3r ?

sungam3r:
Let's publish GraphQL.NET v7.4 and reference it here.

Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <VersionPrefix>7.3.0-preview</VersionPrefix>
+    <VersionPrefix>7.5.0-preview</VersionPrefix>
```
sungam3r, line 4: 7.5?

Shane32: this should always be at least one version ahead.  if we publish 7.4 then this needs to be at least 7.5.  as far as i'm concerned we could just set it to 7.99.0-preview and leave it be.  The only reason it needs to be ahead of the nuget version at all is when referencing a dependency that depends on this package, such as when referencing GraphQL Server 7.3.0 but trying to use a GraphQL.NET preview.  If the GraphQL.NET preview number was 7.2.0-preview then nuget would attempt to bump the nuget package of GraphQL.NET when restoring packages.  But if GraphQL.NET was 7.99.0-preview then there would never be a problem.

sungam3r: I would switch to MinVer to manage version automatically.

Shane32: Yeah, I'm not a fan of adding more build-time dependencies when they are not necessary.  Our system works perfectly now and it took me longer to write these posts than any number of version bumps over the last couple years.

sungam3r:
@Shane32 OK to release GraphQL.NET v7.4 and reference it here?

sungam3r:
>  Then both will be 7.4

This is the reason why I suggest to bump. I'll publish release.

sungam3r:
https://github.com/graphql-dotnet/graphql-dotnet/releases/tag/7.4.0

Also I need v7.4 in [my own project](https://github.com/sungam3r/graphql-introspection-model) and https://github.com/sungam3r/graphql-sdl-exporter to integrate new deprecation feature.

Directory.Build.props
```diff
@@ -32,7 +32,7 @@
     <RootNamespace>GraphQL.Server.$(MSBuildProjectName)</RootNamespace>
     <PackageId>GraphQL.Server.$(MSBuildProjectName)</PackageId>
 
-    <GraphQLVersion>7.3.0</GraphQLVersion>
+    <GraphQLVersion>7.3.1</GraphQLVersion>
```
Shane32, line 35:
```suggestion
    <GraphQLVersion>7.4.0</GraphQLVersion>
```

sungam3r (approved):

PR #1020, opened 2023-04-22
[WIP] Introduce IClassFixture

sungam3r:
The same intent here as in https://github.com/graphql-dotnet/graphql-dotnet/pull/3602. If OK I'll change other tests here. With shared ASP TestServer state tests should work a bit faster.

### DIFF
--- a/tests/Samples.Jwt.Tests/EndToEndTests.cs
+++ b/tests/Samples.Jwt.Tests/EndToEndTests.cs
@@ -7,25 +7,16 @@ using Samples.Tests;
 
 namespace Samples.Jwt.Tests;
 
-public class EndToEndTests : IDisposable
+public class EndToEndTests : IClassFixture<EndToEndTests.Fixture>
 {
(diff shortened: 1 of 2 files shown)


Shane32:
Sharing a class across tests is usually only desired when the amount of setup and teardown time is significant, for example when accessing a database. Typically it is better for each test to be independent. I’m not sure why this would be desired.

sungam3r:
I know, true. Speedup of tests is minimal and is rather a bonus. In fact, I like the fact that the server initialization code is taken out of the test class, in terms of design. In fact each executing test does not need to spin up new test server. I do not think this was your initial goal.

Shane32:
> I like the fact that the server initialization code is taken out of the test class, in terms of design.

This PR does not change that. Moving it from the constructor and disposed to a new class self contained within the parent class is not making the design better. It’s not like the tests contained the initialization code before and it doesn’t now. Rather it is just the same with the code shuffled around a very little bit. 

I do not think these changes should be made here.

PR #1115, opened 2024-01-22
Support multipart file uploads

Shane32:
Based on the GraphQL multipart request spec:
- https://github.com/jaydenseric/graphql-multipart-request-spec

Readme has been updated, sample added, and tests have been added for 100% coverage of new code.  API and endpoints are backwards compatible.  Did not introduce an option to enable/disable new behavior because it is backwards compatible, and additional behavior should have no negative side-effects for existing users.  Also added a new scalar called `FormFileGraphType` which can be used by schemas to ingest `IFormFile` values sent via multipart form posts.  The scalar needs to be added via the `.AddFormFileGraphType()` builder method.

Note that GraphQL.NET Server 7.x is vulnerable to security implications of form-type POST requests being enabled by default.  This is true with or without this feature, and is outlined in the readme. The next version of GraphQL.NET Server will have form-type POST requests disabled by default (as also noted in the readme).

I recommend reviewing the api diff, readme, and sample project to understand the extent of this PR.

Possible changes that could be made:
- Make middleware partial class and move code to separate file to perform form parsing
- Add type-first schema validation attribute to limit potential mime types that can be submitted
- Add enum property to options that configures the form post parsing methods available

### DIFF
--- a/GraphQL.Server.sln
+++ b/GraphQL.Server.sln
@@ -45,8 +45,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "workflows", "workflows", "{
 	ProjectSection(SolutionItems) = preProject
 		.github\workflows\build.yml = .github\workflows\build.yml
 		.github\workflows\codeql-analysis.yml = .github\workflows\codeql-analysis.yml
-		.github\workflows\label.yml = .github\workflows\label.yml
 		.github\workflows\format.yml = .github\workflows\format.yml
+		.github\workflows\label.yml = .github\workflows\label.yml
 		.github\workflows\publish.yml = .github\workflows\publish.yml
 		.github\workflows\test.yml = .github\workflows\test.yml
 		.github\workflows\wipcheck.yml = .github\workflows\wipcheck.yml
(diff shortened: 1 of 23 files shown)


gao-artur:
> Add type-first schema validation attribute to limit potential mime types that can be submitted

This looks important.

The feature looks very cool! It's not an official spec, though, right?

Shane32:
> > Add type-first schema validation attribute to limit potential mime types that can be submitted
> 
> This looks important.

Ok then I suggest we do that as a subsequent PR.  There are some questions in my mind which will be worth review.  So long as the underlying feature (this PR) is fine, let's merge it in first.

> The feature looks very cool! It's not an official spec, though, right?

The [GraphQL working group](https://github.com/graphql/graphql-wg) hasn't ratified any [GraphQL over HTTP protocol](https://github.com/graphql/graphql-over-http).  They do have a proposal, which does not encompass this, the existing WebSockets protocols, persisted queries, or automatic persisted queries.  However, as can be seen by the list in the [spec's readme](https://github.com/jaydenseric/graphql-multipart-request-spec), many existing GraphQL clients and servers support this feature.  There is an existing nuget package to extend GraphQL.NET Server to support this feature [here](https://github.com/JannikLassahn/graphql-dotnet-upload?tab=readme-ov-file), but it suffers from being a separate piece of ASP.NET Core middleware.  I think we should add this as a first-class feature of GraphQL.NET Server.

PR #1121, opened 2024-02-16
Respect the configured serializer in Ui.Playground

jp2masa:
This should allow using AOT with Ui.Playground, but adds an extra dependency to `GraphQL`.

### DIFF
--- a/src/Ui.Playground/Internal/PlaygroundPageModel.cs
+++ b/src/Ui.Playground/Internal/PlaygroundPageModel.cs
@@ -8,10 +8,12 @@ internal sealed class PlaygroundPageModel
     private string? _playgroundCSHtml;
 
     private readonly PlaygroundOptions _options;
+    private readonly IServiceProvider _services;
 
-    public PlaygroundPageModel(PlaygroundOptions options)
+    public PlaygroundPageModel(PlaygroundOptions options, IServiceProvider services)
     {
         _options = options;
+        _services = services;
     }
 
     public string Render()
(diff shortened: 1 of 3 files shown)


Shane32:
It seems that the purpose of this PR is either to (a) reduce dependencies, or (b) reduce code used in AOT scenarios.  I don't think this PR really fulfills the intended purpose.

For ASP.NET Core 2.x, Ui.Playground uses the Newtonsoft converter that ASP.NET Core directly references.  Utilizing the GraphQL serializer won't reduce dependencies and AOT isn't applicable for this version of ASP.NET Core.

Newer versions of ASP.NET Core rely on versions of .NET that include System.Text.Json.  So if the purpose is to reduce dependencies, then this doesn't help.  Newtonsoft.Json isn't designed for AOT, but ignoring this, you could say that using the same serializer as GraphQL.NET would allow System.Text.Json to be trimmed.  But this argument doesn't hold when you look at the PR: the analyzer is going to see that System.Text.Json will be used if `GetService` returns `null`, and so will need to include STJ in any case.  Futher, adding a dependency on GraphQL.NET is going to increase the size of any application that only uses the UI project.

There is another possible reason to utilize the configured serializer, and that is if we wanted to allow the user to customize the serialization process.  But the serialization format in this case is known and fixed -- for instance, it (probably) wouldn't work if the serializer was configured for pascal case.

If you want to support AOT or other serializers, I suggest that you make a protected method to handle serialization.  Then in your own implementation you can override the method and rewrite the serialization as desired.  For AOT/trimming, the analyzer will see that the base method implementation is never referenced, and will be able to trim out the default serializer.  Then if you wish to use the GraphQL.NET serializer, then it should be relatively easy to make a derived class, pull the serializer from DI, and use that within the protected method.

Another idea specifically for NativeAOT in .NET 7 and 8 would be add .NET 7 TFM that uses source generation to eliminate reflection.

src/Ui.Playground/Internal/PlaygroundPageModel.cs
```diff
@@ -60,19 +62,30 @@ public string Render()
...
+        if (_services.GetService(typeof(IGraphQLSerializer)) is IGraphQLSerializer serializer)
+        {
+            using var stream = new MemoryStream();
+            serializer.WriteAsync(stream, value).Wait();
+
+            stream.Position = 0;
+
+            using var reader = new StreamReader(stream, Encoding.UTF8);
+            return reader.ReadToEnd();
+        }
```
Shane32, line 76: FYI, GraphQL.NET provides a synchronous method as well:

```suggestion
        var serializer = _services.GetService<IGraphQLTextSerializer>();
        if (serializer != null)
        {
            return serializer.Serialize(value);
        }
```

Shane32:
See #1124 for a PR to provide NativeAOT compatibility for this project.

jp2masa:
> See https://github.com/graphql-dotnet/server/pull/1124 for a PR to provide NativeAOT compatibility for this project.

I think that would work for me. My original idea was to use the registered serializer because, in my case, it's a System.Text.Json serializer with a source generated `JsonSerializerContext`.

But the approach in #1124 looks better, so I'm closing this.

Shane32:
v7.7.1 has been published containing #1124

PR #1142, opened 2024-08-09
Status codes per GraphQL over HTTP spec

Shane32:


docs/migration/migration8.md
```diff
@@ -6,9 +6,17 @@
...
+  and for authentication errors may return a 401 or 403 status code.  These changes are purusant
+  to the [GraphQL over HTTP specification](https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md).
+  See the breaking changes section below for more information.
 
 ## Breaking changes
 
+- `GraphQLHttpMiddlewareOptions.ValidationErrorsReturnBadRequest` is now a nullable boolean where
+   `null` means "use the default behavior".  The default behavior is to return a 200 status code
+  when the response content type is `application/json` and a 400 status code otherwise.  The
+  default value for this in v7 was `true`; set this option to retain the v7 behavior.
```
gao-artur, line 19: I think the new default is the only correct behavior. Let's deprecate this configuration, maybe?

Shane32: No, users may likely have clients that count on GraphQL returning 400 even if they use application/json. The GraphQL over http spec only recommends 200 be returned, and only because (in theory) proxy services may return 400 with application/json. Users will know if their configuration meets that criteria or not, and if such a concern is warranted. And other users’ client code may always expect 200 even in the case of an error. 

I think this is rather important to keep flexible, just as it was configurable in v7.

gao-artur: The client can choose the status codes by using either `application/json` or `application/graphql-response+json`. The additional option: using `application/json` and expecting 4xx codes is not forbidden but discouraged. What can force users to prefer this behavior when they can just change the `Accept` header?

Shane32: You're assuming that they can control the clients behavior easily.  In our company, we have a large infrastructure of GraphQL services that talk to each other.  One client may be .NET code, one may be javascript, we even have python code.  Changing the format of responses would require us to review each of these different pieces of code scattered throughout our various codebases.  Any tests we've written that validate the returned status code may need to change.  Since we are not exposing GraphQL to the public, compatibility with the GraphQL over HTTP protocol is not a concern.  And if we had a public API, it would be important to maintain behavior of the endpoint throughout the supported version of that API.  I've always had our servers return 400 for validation errors, and I expect that to continue.

It is also important to note that Apollo GraphQL server behaves as if this option were `false` by default, always returning 200 for validation errors (which are not transport errors).  For compatibility with other servers alone, we may want to keep this option.

I really don't see any reason this isn't a valid and normal configuration option for the server project, just like changing the default content type returned, enabling/disabling CSRF protection, or enabling/disabling form parsing.  Keep in mind that it's just an option, and it's configured to the recommended default per the draft GraphQL over HTTP spec.  (The spec isn't even finalized yet!)

gao-artur: Understood

gao-artur (approved):

PR #1153, opened 2024-10-01
Pin graphql-voyager to 1.3.0, fix #1152

### LINKED ISSUE #1152: GraphQL.Server.Ui.Voyager recent issues with zoom to fit
The NuGet package GraphQL.Server.Ui.Voyager (the most recent 8.0.2 and older 7.7.1) started to have problems with the initial zoom to fit. The whole graph is rendered out of the screen area and it all becomes rather unusable.

I noticed this issue first on Monday but my colleague tells this started last week.

See this online example with the problem (using the package):
https://api.nexar.com/ui/voyager

and HTML copied from the above browser, with the same problem:
[nexar-html-from-package.html.zip](https://github.com/user-attachments/files/17209299/nexar-html-from-package.html.zip)

Compare with working fine original Voyager HTML, with the same GraphQL endpoint:
[nexar-voyager.html.zip](https://github.com/user-attachments/files/17206374/nexar-voyager.html.zip)
(based on the original [Voyager HTML example](https://github.com/graphql-kit/graphql-voyager/blob/main/example/cdn/index.html))

nightroman:


Shane32:
Let's pin 2.0.0; it has been in use for a year now, right?

nightroman:
> Let's pin 2.0.0; it has been in use for a year now, right?

But it does not work for me with 2.0.0.
1.3.0 is the maximum that works.

nightroman:
Take a looks at 1.3.0 (woks) and 2.0.0 (loading... infinitely)
[2-cases.zip](https://github.com/user-attachments/files/17212529/2-cases.zip)

Inok:
> Let's pin 2.0.0; it has been in use for a year now, right?

No, it hasn't been: in 2.0, `voyager.min.js` [was removed](https://github.com/graphql-kit/graphql-voyager/releases/tag/v2.0.0), so `https://cdn.jsdelivr.net/npm/graphql-voyager/dist/voyager.min.js` (without version) still points to the 1.3 version. 

However, the corresponding CSS file is still present in 2.0,  so there was CSS from 2.0 and JS from 1.3 for a year. 
And in 2.1, it seems, the CSS finally became incompatible with the JS v1.3.

Switching to the new `voyager.standalone.js` from 2.0 (supersedes `voyager.min.js`) doesn't help because of some configuration changes in 2.0.

Shane32:
Ok sounds good.  Let's add the sha hash to the tag to ensure it does not change in the future, and I'll merge it.

nightroman:
> Ok sounds good. Let's add the sha hash to the tag to ensure it does not change in the future, and I'll merge it.

Sorry, I am not sure what this is. Is it something I should do?

Shane32:
https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity

Shane32:
Probably some website like this can calculate it. 

https://www.srihash.org/

alexander-jesner-AP:
I assume the hosting code needs to be updated for the 2.0 version to work?
Like this: https://github.com/graphql-kit/graphql-voyager/blob/c644fb529f8aa1f987c505aef4527207a0c752fd/example/cdn/index.html#L31-L51

nightroman:
@alexander-jesner-AP Yes, I think so and hope this will be eventually the case, the proper fix. That's why I mentioned the same example in #1152 

@Shane32 I have engaged for the quick fix of the regression, to make things working again. I now see it's getting more involved and and it's not really my area of expertise. At the moment I am not able to implement what you are requesting. Thank you for your help.

src/Ui.Voyager/Internal/voyager.cshtml
```diff
@@ -31,8 +31,8 @@
       copy them directly into your environment, or perhaps include them in your
       favored resource bundler.
   -->
-  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/graphql-voyager/dist/voyager.css" />
-  <script src="https://cdn.jsdelivr.net/npm/graphql-voyager/dist/voyager.min.js"></script>
+  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/graphql-voyager@1.3.0/dist/voyager.css" integrity="sha384-ttbHva9FGXoTKN/NPE2LU+rPFxgnwqXZNrPGqXArLYC/NsW9Gwl/yM1Rja1eSS27" crossorigin="anonymous">
```
Shane32, line 34:
```suggestion
  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/graphql-voyager@1.3.0/dist/voyager.css" integrity="sha384-ttbHva9FGXoTKN/NPE2LU+rPFxgnwqXZNrPGqXArLYC/NsW9Gwl/yM1Rja1eSS27" crossorigin="anonymous" />
```

Shane32:
I'll issue a release with it pinned to 1.3.0

If someone wants to post a PR with it properly upgraded to 2.x, we can merge that in subsequently.

PR #1154, opened 2024-10-07
Add KeepAliveMode and SupportedWebSocketSubProtocols options

Shane32:
Adds a 'secure' keep-alive mode where:
- A ping is sent on a timer from the last received pong
- Ping messages carry a unique payload (random guid)
- Received pong messages verify the payload
- If a pong is not received with the correctly id in a timely fashion, the client is forcibly disconnected

Reason:
- With subscriptions that continuously send lots of data, there is no way to determine if socket backpressure due to slow client connections is causing resource exhaustion.  This feature will validate that buffers are drained over the specified time period, as the client would not know what id to send with a pong if it did not process the incoming messages fast enough.

Consider:
- Name of KeepAliveMode enum values
- Adding separate timer property that represents the amount of time to wait for a corresponding pong after a sent ping.  Currently this just matches the ping timer.

### DIFF
--- a/src/Transports.AspNetCore/WebSockets/GraphQLWs/PingPayload.cs
+++ b/src/Transports.AspNetCore/WebSockets/GraphQLWs/PingPayload.cs
@@ -1,0 +1,12 @@
+﻿namespace GraphQL.Server.Transports.AspNetCore.WebSockets.GraphQLWs;
+
+/// <summary>
+/// The payload of the ping message.
+/// </summary>
+public class PingPayload
+{
+    /// <summary>
+    /// The unique identifier of the ping message.
+    /// </summary>
+    public string? id { get; set; }
+}
--- a/src/Transports.AspNetCore/WebSockets/GraphQLWs/PingPayload.cs
+++ b/src/Transports.AspNetCore/WebSockets/GraphQLWs/PingPayload.cs
@@ -1,4 +1,4 @@
-﻿namespace GraphQL.Server.Transports.AspNetCore.WebSockets.GraphQLWs;
+namespace GraphQL.Server.Transports.AspNetCore.WebSockets.GraphQLWs;
 
 /// <summary>
 /// The payload of the ping message.
--- a/src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
+++ b/src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
@@ -4,6 +4,11 @@ namespace GraphQL.Server.Transports.AspNetCore.WebSockets.GraphQLWs;
 public class SubscriptionServer : BaseSubscriptionServer
 {
     private readonly IWebSocketAuthenticationService? _authenticationService;
+    private readonly IGraphQLSerializer _serializer;
+    private readonly GraphQLWebSocketOptions _options;
+    private DateTime _lastPongReceivedUtc;
+    private string? _lastPingId;
+    private readonly object _lastPingLock = new();
 
     /// <summary>
     /// The WebSocket sub-protocol used for this protocol.
@@ -67,6 +72,8 @@ public class SubscriptionServer : BaseSubscriptionServer
         UserContextBuilder = userContextBuilder ?? throw new ArgumentNullException(nameof(userContextBuilder));
         Serializer = serializer ?? throw new ArgumentNullException(nameof(serializer));
         _authenticationService = authenticationService;
+        _serializer = serializer;
+        _options = options;
     }
 
     /// <inheritdoc/>
@@ -90,7 +97,9 @@ public class SubscriptionServer : BaseSubscriptionServer
             }
             else
             {
+#pragma warning disable CS0618 // Type or member is obsolete
                 await OnConnectionInitAsync(message, true);
+#pragma warning restore CS0618 // Type or member is obsolete
             }
             return;
         }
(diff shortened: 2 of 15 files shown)


src/Transports.AspNetCore/WebSockets/GraphQLWs/PingPayload.cs
```diff
@@ -0,0 +1,12 @@
...
+
+/// <summary>
+/// The payload of the ping message.
+/// </summary>
+public class PingPayload
+{
+    /// <summary>
+    /// The unique identifier of the ping message.
+    /// </summary>
+    public string? id { get; set; }
```
Shane32, line 11: In order to serialize in camel case, we need to define the property in camel case here.  However, this property (including its name, capitalization and contents) are purely arbitrary.  So we could use pascal case -- but it would be serialized as such.

Using attributes is not an option because GraphQL supports two serialization libraries (System.Text.Json and Newtonsoft.Json), neither of which are referenced here (e.g. for .NET 4.8).

src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
```diff
@@ -113,6 +122,69 @@ public override async Task OnMessageReceivedAsync(OperationMessage message)
         }
     }
 
+    /// <inheritdoc/>
+    [Obsolete($"Please use the {nameof(OnConnectionInitAsync)} and {nameof(OnKeepAliveLoopAsync)} methods instead. This method will be removed in a future version of this library.")]
+    protected override Task OnConnectionInitAsync(OperationMessage message, bool smartKeepAlive)
```
Shane32, line 127: The code has been carefully crafted to be fully backwards compatible, just in case of the rare instance that someone derived from this class to customize behavior.

src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
```diff
@@ -113,6 +122,69 @@ public override async Task OnMessageReceivedAsync(OperationMessage message)
...
+    {
+        if (smartKeepAlive)
+            return OnConnectionInitAsync(message);
+        else
+            return base.OnConnectionInitAsync(message, smartKeepAlive);
+    }
+
+    /// <inheritdoc/>
+    protected override Task OnKeepAliveLoopAsync(TimeSpan keepAliveTimeout, KeepAliveMode keepAliveMode)
```
Shane32, line 136: The actual keep-alive code isn't actually tested. Having timer-based code within tests is notoriously unreliable, due to unpredictable and slow speed of the GitHub servers running the tests.

src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
```diff
@@ -113,6 +122,69 @@ public override async Task OnMessageReceivedAsync(OperationMessage message)
...
+            return SecureKeepAliveLoopAsync(keepAliveTimeout, keepAliveTimeout);
+        }
+        return base.OnKeepAliveLoopAsync(keepAliveTimeout, keepAliveMode);
+
+        // pingInterval is the time since the last pong was received before sending a new ping
+        // pongInterval is the time to wait for a pong after a ping was sent before forcibly closing the connection
+        async Task SecureKeepAliveLoopAsync(TimeSpan pingInterval, TimeSpan pongInterval)
+        {
+            lock (_lastPingLock)
+                _lastPongReceivedUtc = DateTime.UtcNow;
```
gao-artur, line 151: Why is there a need for a lock here? Is it because struct writing is not atomic?

src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
```diff
@@ -113,6 +122,69 @@ public override async Task OnMessageReceivedAsync(OperationMessage message)
...
+            {
+                // Wait for the next ping interval
+                TimeSpan interval;
+                var now = DateTime.UtcNow;
+                DateTime lastPongReceivedUtc;
+                lock (_lastPingLock)
+                {
+                    lastPongReceivedUtc = _lastPongReceivedUtc;
+                }
+                var nextPing = _lastPongReceivedUtc.Add(pingInterval);
```
gao-artur, line 162: Should it be the local variable `lastPongReceivedUtc` here?

src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
Shane32: That is correct; writing a struct is not atomic.  However in this case the struct contains only a single `ulong` field.  But .NET does not guarantee atomic writes on 64-bit values either (since the application may be running as a 32-bit process).

> Other types, including `double` and `long`, cannot be marked `volatile` because reads and writes to fields of those types cannot be guaranteed to be atomic.

src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
Shane32: Yes, good catch.

Shane32:
```suggestion
                var nextPing = lastPongReceivedUtc.Add(pingInterval);
```

gao-artur (approved):

src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs
Shane32: Note that resource contention is extremely low.  For every `pingInterval` (e.g. 10 seconds), there should only be a couple writes and a couple reads.

Issue #1158, opened 2024-10-23
GraphQL.Server.Ui.Playground autoComplete or Type Box doesn't disappear


CDMikeThomas:
We are using the package 'GraphQL.Server.Ui.Playground' in version 8.0.2 for having the GraphQL playground UI.
In Chromium based browsers (version latest public available, e.g. Google Chrome, Edge) the autoComplete and type boxes are not disappearing in the editor area of the playground UI. 
The issue seems to be browser specific as in Firefox (version latest public available) the issue does not exist.
Issue is equivalent to what I found here:
https://github.com/graphql/graphql-playground/issues/1429

Below are the images the bug in our software.
Here it is shown that the boxes are not disappearing:
![image](https://github.com/user-attachments/assets/07f25128-a982-43e2-ae69-7c56752245c9)
This error message keeps stacking up in the browser console:
![image (1)](https://github.com/user-attachments/assets/eae9f1bd-aaf4-42a5-a269-f8f2c71e0e11)

Shane32:
Will be released in v8.1.0

PR #1159, opened 2024-10-23
Pin playground to v1.7.28 and add polyfill for mutation events

Shane32:
Closes:
- #1158 

See:
- https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity
- https://developer.chrome.com/blog/mutation-events-deprecation
- https://github.com/mfreed7/mutation-events-polyfill#readme
- https://www.npmjs.com/package/mutation-events

### DIFF
--- a/src/Ui.Playground/Internal/playground.cshtml
+++ b/src/Ui.Playground/Internal/playground.cshtml
@@ -4,9 +4,10 @@
   <meta charset=utf-8 />
   <meta name="viewport" content="user-scalable=no, initial-scale=1.0, minimum-scale=1.0, maximum-scale=1.0, minimal-ui">
   <title>Graphcool Playground</title>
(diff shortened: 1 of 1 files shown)


gao-artur (approved):
I don't really understand what you did here but OK 🙂

Shane32:
@gao-artur I updated the PR description to more fully explain the changes.

gao-artur:
@Shane32 I think this PR broke the autocomplete in the Playground.

Shane32:
Well, it's not working properly now, but I must say, I think it was working even worse without this PR.  What do you propose?  Playground isn't supported and hasn't had a release in 5 years.  I'm not really interesting in investing time figuring out how to patch the javascript code.  My best suggestion is to mark all members as obsolete and release one final version, after which we remove the package from the solution.  I had hoped that by introducing the polyfill, it would start working properly again.  But it does not appear to have solved all the problems.

gao-artur:
I'm fine with that. We will probably stop using it soon.

PR #1180, opened 2025-01-16
Add NativeAOT sample; bump GraphQL.NET to 8.3.0

Shane32:
Adds a working sample of a GraphQL.NET Server running on .NET 8 Native AOT, plus CI tests to verify that it operates correctly after being AOT compiled.  Since the sample schema in Samples.Schemas.Chat used for the rest of the projects is type-first, not recommended for Native AOT, this sample only provides a simple 'hello world' schema.

See also:
- https://github.com/graphql-dotnet/server/discussions/1179


Directory.Build.props
```diff
@@ -32,7 +32,7 @@
     <RootNamespace>GraphQL.Server.$(MSBuildProjectName)</RootNamespace>
     <PackageId>GraphQL.Server.$(MSBuildProjectName)</PackageId>
 
-    <GraphQLVersion>[8.2.1,9.0.0)</GraphQLVersion>
+    <GraphQLVersion>[8.0.1-preview-1114]</GraphQLVersion>
```
Shane32, line 35: beta of 8.3.0; need to fix after 8.3.0 is released.

Shane32:
```suggestion
    <GraphQLVersion>[8.3.0,9.0.0)</GraphQLVersion>
```

PR #1189, opened 2025-06-12
Support [AllowAnonymous] along with [Authorize]

Shane32:
This allows bypassing parent type's authorization while defining new requirements on the field. Previously any requirements on the field were ignored.

Sample:

```gql
# .AuthorizeWithRoles("Admin")
type Mutation {
  createUser(name: String!): String # requires Admin role

  # .AllowAnonymous()
  # .Authorize()
  updateOwnProfile(email: String!): Boolean # requires any authenticated user

  # .AllowAnonymous()
  submitFeedback(message: String!): Boolean # anonymous user
}
```

Recently I found that I had assumed that `.AllowAnonymous()` would stack with `.Authorize()`, but this was not the case.  So, is this a security bug fix, or a new feature?  Not sure, but I don't think it warrants a major version bump.  All prior tests pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -404,8 +404,8 @@ Both roles and policies are supported for output graph types, fields on output g
 and query arguments.  If multiple policies are specified, all must match; if multiple roles
 are specified, any one role must match.  You may also use `.Authorize()` and/or the
 `[Authorize]` attribute to validate that the user has authenticated.  You may also use
-`.AllowAnonymous()` and/or `[AllowAnonymous]` to allow fields to be returned to
-unauthenticated users within an graph that has an authorization requirement defined.
+`.AllowAnonymous()` and/or `[AllowAnonymous]` to allow fields to bypass authorization
+requirements defined on the type that contains the field.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -404,8 +404,8 @@ Both roles and policies are supported for output graph types, fields on output g
 and query arguments.  If multiple policies are specified, all must match; if multiple roles
 are specified, any one role must match.  You may also use `.Authorize()` and/or the
 `[Authorize]` attribute to validate that the user has authenticated.  You may also use
-`.AllowAnonymous()` and/or `[AllowAnonymous]` to allow fields to be returned to
-unauthenticated users within an graph that has an authorization requirement defined.
+`.AllowAnonymous()` and/or `[AllowAnonymous]` to allow fields to bypass authorization
+requirements defined on the type that contains the field.
```
Shane32, line 408: This is not a change, just a clarification.

gao-artur (approved):

marc81-marc:
Ok it was at kids i just put it back to parent 🤯?

PR #1202, opened 2025-10-22
Fix Altair GraphQL loading

### LINKED ISSUE #1201: GraphQL.Server.Ui.Altair 8.3.1 - Doesn't work - stuck on splash screen
The Altair UI playground fails to load in the browser, it is stuck on the splash scree, and throws the following error messages in the browser console.

_Uncaught SyntaxError: Cannot use import statement outside a module_

_Uncaught ReferenceError: AltairGraphQL is not defined_

See some additional details on this issue logged by a user directly with Altair --> https://github.com/altair-graphql/altair/issues/2924

imolorhe:
Due to recent changes in Altair GraphQL v8.4.0 (updating the angular version and build system which uses native ESM module imports), the current static boilerplate no longer works properly _(Sorry about the breaking change!)_

Updating the script tags as appropriate to conform to the new setup.

Closes #1201 

Given this was a change in a recent version of Altair and the older versions still work with the older template, it might be prudent to use a fixed version of Altair to avoid breaking changes in newer versions breaking GraphQL.Server.Ui.Altair

Here's a [codesandbox](https://codesandbox.io/p/sandbox/altair-graphql-client-demo-forked-8pcx2y?file=%2Findex.html%3A34%2C13) that uses the updated boilerplate ([demo](https://8pcx2y.csb.app/))

_FYI jsdelivr allows you to set a major, minor or patch version, so it could be fixed to `cdn.jsdelivr.net/npm/altair-static@8/build/dist/` or `cdn.jsdelivr.net/npm/altair-static@8.3/build/dist/` to still get minor or patch version updates_

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -25,7 +25,7 @@ jobs:
          NUGET_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
 
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v3
+      uses: github/codeql-action/init@v4
       with:
         queries: security-and-quality
         languages: csharp
@@ -37,4 +37,4 @@ jobs:
       run: dotnet build --no-restore
 
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@v3
+      uses: github/codeql-action/analyze@v4


Shane32 (approved):

imolorhe:
@CaiusSacu there is no more runtime.js

